PR #10, opened 2021-04-23
Update docker-compose docs

whikloj:
Hey @danloveg 

Here is the `.dockerenv` I needed to get the docker-compose setup running. I passed your code along to some other devs I work with and one hit on some of the same problems I hit. I then felt guilty as I just sent them to your docs and forgot to warn them.

I also altered the `docker-compose.dev.yml` to avoid the files as directories problems.

I did get the local development setup working, but I had to run both the django server (`python3 manage.py runserver`) and the django-rq worker (`python3 manage.py rq_worker default`) outside of docker. Inside I had problems with the rq_worker losing its connection and dying (taking the whole container down with it).

This probably should not be needed but I found this way I could connect PyCharm to the one I wanted to trace and see what was happening.

### DIFF
--- a/bagitobjecttransfer/docker-compose.dev.yml
+++ b/bagitobjecttransfer/docker-compose.dev.yml
@@ -22,8 +22,8 @@ services:
     volumes:
       - ./:/app
       - /tmp:/tmp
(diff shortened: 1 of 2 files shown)


danloveg (approved):
Thanks for these changes, everything looks good to me.

PR #12, opened 2021-04-28
Use changed url name

### LINKED ISSUE #11: Transfer link in profile is broken
On the Profile page, if you have _not_ performed a transfer you get a paragraph with a link, but the link is to the current page.

> You have not sent any records to the NCTR. Start a <a href="http://localhost:8000/profile">new transfer</a> to get started.

Should be to the transfer page.

whikloj:
Resolves #11

### DIFF
--- a/bagitobjecttransfer/recordtransfer/templates/recordtransfer/profile.html
+++ b/bagitobjecttransfer/recordtransfer/templates/recordtransfer/profile.html
@@ -82,7 +82,7 @@
             {% endif %}
         {% else %}
             <span>
-                {% url 'recordtransfer:formpreparation' as transfer_url %}
+                {% url 'recordtransfer:transfer' as transfer_url %}
                 {% blocktrans %}
                     You have not sent any records to the NCTR.
                     <a href="{{ transfer_url }}">Start a new transfer</a> to get started.


whikloj:
Ohhhh actually I see you're missing the entire FormPreparation class now. So the step is missing. I guess the point was too only hit that page the first time you start a transfer?

danloveg:
You're right, there used to be a form preparation view that I had users hit before they attempted a transfer. That has been superseded by the first part of the transfer form, `forms.AcceptLegal`.

whikloj:
Oh should I have removed the view and template as part of this too?

https://github.com/danloveg/NCTR-Bagit-Record-Transfer/blob/master/bagitobjecttransfer/recordtransfer/views.py#L35-L37

https://github.com/danloveg/NCTR-Bagit-Record-Transfer/blob/master/bagitobjecttransfer/recordtransfer/templates/recordtransfer/formpreparation.html

danloveg:
Good catch, yes, there were a few remaining references to `formpreparation`, I removed them, the template, and the view in [this commit](https://github.com/danloveg/NCTR-Bagit-Record-Transfer/commit/607fea4e83002d2c3ec09773e90d1c832abcc001).

PR #29, opened 2021-05-05
Override generateHTML function

### LINKED ISSUE #19: Typing Date Makes Estimated Date Checkbox Disappear
On Step 4 of the transfer form (Record Description), if you type a full date out in the Earliest date or Latest date fields, the estimated date checkbox disappears. You can make it re-appear by clicking away from the input and clicking the input again.

The fix may be as simple as refreshing the checkbox when a date is selected.

whikloj:
Resolves #19 

It appears that when you [type in a full date](https://github.com/jquery/jquery-ui/blob/32850869d308d5e7c9bf3e3b4d483ea886d373ce/ui/widgets/datepicker.js#L725-L728) the [`_updateDatepicker`](https://github.com/jquery/jquery-ui/blob/32850869d308d5e7c9bf3e3b4d483ea886d373ce/ui/widgets/datepicker.js#L826) function is called, which calls the [`_generateHTML`](https://github.com/jquery/jquery-ui/blob/32850869d308d5e7c9bf3e3b4d483ea886d373ce/ui/widgets/datepicker.js#L1657) function to re-generate the date picker.

The `_generateHTML` just returns a string of the HTML so I override the function, call the original one, create the estimated date checkbox HTML and return both string.

As the `_generateHTML` function is called for all the other events I could remove them. 

But as it is called for all the other events, I had to also dynamically determine which input box we were currently in.

### DIFF
--- a/bagitobjecttransfer/recordtransfer/static/recordtransfer/js/widgetconfig.js
+++ b/bagitobjecttransfer/recordtransfer/static/recordtransfer/js/widgetconfig.js
@@ -2,76 +2,38 @@ $(() => {
     /***************************************************************************
      * jQuery Date Picker Setup
      **************************************************************************/
-    function addEstimatedButtonToStartDate(instance) {
-        var buttonPane = $(instance).datepicker("widget")
-        var html = [
-            '<div style="margin-bottom: 5px;">',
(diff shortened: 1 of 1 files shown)


danloveg:
Looks great, and I like how the datepicker initialization for both date pickers could be combined into one call.

PR #32, opened 2021-05-17
Make rights a manageable taxonomy

### LINKED ISSUE #20: Rights Form is Too Open
The form to fill out the rights for the transfer is too open for interpretation, and may be confusing for people who are not familiar with copyright and other types of rights.

A dropdown for the rights type may be more appropriate than an open-text field. We may want to provide an "Other" option if someone would like to enter rights in an open-text field of a type we are not familiar with or have overlooked.

whikloj:
Resolves #20 

Ok, I think I got it all but have a good look as this was an education on top of an education.

I also seem to have an extra change added to my new migration when I ran the `python manage.py makemigrations`. I didn't want to remove it, but also was not sure what it was related to.

I use PyCharm and it "cleans" up import statements so it re-ordered/removed some in `admin.py` which seem to not have been used. But I can revert any of those changes if there were made in error.

Lastly, as this was my first crack at this type of change please make any and all suggestions.


danloveg:
I'll review this after lunch, but really it looks quite good! I think that extra part of your migration was from a change I made to a model and forgot to make a migration for, so nothing to worry about there.

danloveg:
I'm going to merge this into an **issue-20** branch on this repo. I have a couple changes I want to make before I merge this into master.

I thought it was really smart to put the rights taxonomies into the database, I wish I would have thought of that first!

danloveg:
@whikloj you can see the new changes I've made in #33

PR #33, opened 2021-05-17
Address Issue #20

### LINKED ISSUE #20: Rights Form is Too Open

danloveg:
Closes #20.

I added a commit on top of @whikloj's work to modify how the rights statement is cleaned. I didn't want the other parts of the app to be aware of `other_rights_statement_type` because that isn't a part of CAAIS. I removed reference to `other_rights_statement_type` everywhere except in the `RightsForm`.

I also made it so that the note is always empty. I personally don't see the value in having a rights note when there is already an open text box (the `rights_statement_value`) for people to put in notes about the rights. The fewer fields we have users fill out, the better.


danloveg:
@whikloj check this out and let me know what you think. I didn't touch too much of your work besides removing references to `other_rights_statement_type` outside of the form.

whikloj:
Whatever you think is best, I have no experience with CAAIS so if this is needed to keep it valid then 👍

danloveg:
Cool, and thanks for your help on this!

PR #38, opened 2021-06-09
Send emails for certain account changes

### LINKED ISSUE #31: Major changes to Users should send email
If a user changes their account or has their account changed, they should be notified via email.

An email should be sent in these cases:
- A user's staff status changes
- A user's superuser status changes
- A user's password is changed
- A user has been deactivated

whikloj:
Resolves #31 

Not sure if there is a better way to intercept the password change, my theory based on the [method in Admin](https://github.com/django/django/blob/main/django/contrib/auth/admin.py#L129-L190) is if it raised an exception that would skip the message stuff and immediately raise up to the top. Otherwise we return the form or redirect response once we have (or have not) sent the email.

### DIFF
--- a/bagitobjecttransfer/recordtransfer/admin.py
+++ b/bagitobjecttransfer/recordtransfer/admin.py
@@ -4,9 +4,11 @@ import zipfile
 from io import StringIO, BytesIO
 from pathlib import Path
 
+from django.contrib.admin.utils import unquote
+
 from bagitobjecttransfer.settings.base import MEDIA_ROOT
 from django.contrib import admin, messages
-from django.contrib.auth.admin import UserAdmin
+from django.contrib.auth.admin import UserAdmin, sensitive_post_parameters_m
 from django.http import HttpResponse, HttpResponseRedirect
 from django.template.loader import render_to_string
 from django.urls import reverse
@@ -17,7 +19,7 @@ from recordtransfer.atom import flatten_meta_tree_atom_style
 from recordtransfer.bagger import update_bag
 from recordtransfer.caais import flatten_meta_tree
 from recordtransfer.forms import BagForm, InlineBagForm
-from recordtransfer.jobs import create_downloadable_bag
+from recordtransfer.jobs import create_downloadable_bag, send_user_account_updated
 from recordtransfer.models import Bag, BagGroup, User, UploadedFile, UploadSession, Job, Right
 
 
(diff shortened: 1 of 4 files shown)


whikloj:
Ohhh should I be using `gettext()` to allow for translating (I think)?

whikloj:
Additionally, if your password has been changed or your account has been deactivated and you feel that was a mistake should we include some sort of main contact email address? This could be a configured email address and not tied to any actual user.

bagitobjecttransfer/recordtransfer/admin.py
```diff
@@ -90,15 +92,68 @@ class CustomUserAdmin(UserAdmin):
...
+                change = user.is_staff
+            if message is not None:
+                message += " have been {} your account".format("added to" if change else "removed from")
+                message_list.append(message)
+        return message_list
+
+    @staticmethod
+    def _alert_user_changed(form):
+        """ Check if any of the account details updated require notification. """
+        return len([val for val in form.changed_data if val in ['is_superuser', 'is_staff']]) > 0
```
danloveg, line 156: It would make more sense to use the `any` function here

bagitobjecttransfer/recordtransfer/admin.py
```diff
@@ -90,15 +92,68 @@ class CustomUserAdmin(UserAdmin):
...
+                    context = {
+                        'subject': "Account Deactivated",
+                        'changed_item': "account",
+                        'changed_status': "deactivated"
+                    }
+                else:
+                    context = {
+                        'subject': "Account updated",
+                        'changed_item': "account",
+                        'changed_status': "updated",
```
danloveg, line 129: These should use gettext() to provide translation support

bagitobjecttransfer/recordtransfer/admin.py
```diff
@@ -90,15 +92,68 @@ class CustomUserAdmin(UserAdmin):
...
+    def user_change_password(self, request, id, form_url=''):
+        """ Send a notification email when a user's password is changed. """
+        response = super().user_change_password(request, id, form_url)
+        user = self.get_object(request, unquote(id))
+        form = self.change_password_form(user, request.POST)
+        if form.is_valid() and request.method == 'POST':
+            context = {
+                'subject': "Password updated",
+                'changed_item': "password",
+                'changed_status': "updated"
```
danloveg, line 105: These should use gettext() to provide translation support

bagitobjecttransfer/recordtransfer/admin.py
```diff
@@ -90,15 +92,68 @@ class CustomUserAdmin(UserAdmin):
...
+            if item == "is_superuser":
+                message = "Superuser privileges"
+                change = user.is_superuser
+            elif item == "is_staff":
+                message = "Staff privileges"
+                change = user.is_staff
+            if message is not None:
+                message += " have been {} your account".format("added to" if change else "removed from")
+                message_list.append(message)
+        return message_list
```
danloveg, line 151: These also need gettext(), but the way this is written is not conducive to translation. For example, what if in some language it makes grammatic sense to say "Added to your account is Staff privileges"? There are only four cases here, it should be OK to hard code them:

```python

if item == 'is_superuser' and user.is_superuser:
    msg = gettext('Superuser privileges have been added to your account')
    message_list.append(msg)
elif item == 'is_superuser' and not user.is_superuser:
    msg = gettext('Superuser privileges have been removed from your account')
    message_list.append(msg)
```

And so on

danloveg (commented):
Looking good, mostly just a few issues relating to translation. I haven't added translation support to the email templates yet (not sure why) but that'll be a future change.

danloveg:
> Additionally, if your password has been changed or your account has been deactivated and you feel that was a mistake should we include some sort of main contact email address? This could be a configured email address and not tied to any actual user.

Yes, maybe we can include a SUPPORT_EMAIL setting, and fall back to the ARCHIVIST_EMAIL if it's not set.

whikloj:
I forgot about the ARCHIVIST_EMAIL, we can use that instead if you think it is appropriate.

whikloj:
I'm realizing that you may also need a user configurable language field. So if I (as some admin) makes a change, the email comes to you (as the user who's account I altered) in your language of choice. There might be special ways to handle that, but just thought I would mention it.

danloveg:
Yes, I haven't had the chance to incorporate any sort of language switcher, but that'll definitely be something that needs to happen in the future. It's possible that a user's preferred language could be tracked, and whenever emails are sent to them, it's sent to them in the language they use most often on the site.

bagitobjecttransfer/recordtransfer/templates/recordtransfer/email/account_updated.html
```diff
@@ -15,9 +16,9 @@
         <div style="
             padding: 0px 20px;">
             <div>
-                <p>Your {{ changed_item }} for the NCTR Record Transfer Portal has been {{ changed_status }}.</p>
+                <p>{% blocktranslate %}Your {{ changed_item }} for the NCTR Record Transfer Portal has been {{ changed_status }}.{% endblocktranslate %}</p>
```
danloveg, line 19: Thanks for adding this, no changes needed here I just wanted to mention that I want to do this for all the other emails at some point, too.

bagitobjecttransfer/recordtransfer/admin.py
```diff
@@ -136,25 +137,18 @@ def save_model(self, request, obj, form, change):
...
+        if "is_superuser" in changed_data:
+            if user.is_superuser:
+                message_list.append("Superuser privileges have been added to your account.")
+            else:
+                message_list.append("Superuser privileges have been removed from your account.")
+        if "is_staff" in changed_data:
+            if user.is_staff:
+                message_list.append("Staff privileges have been added to your account.")
+            else:
+                message_list.append("Staff privileges have been removed from your account.")
```
danloveg, line 149: These should use gettext(), but I like this new structure, it's more readable than before.

danloveg (commented):
Just one small place where we still need calls to `gettext()`.

Issue #22, opened 2021-05-03
Thank You Email Should Include Accession Agreement


danloveg:
The email sent with the `transfer_success.html` template should include an accession agreement document rather than just a simple thank you with an archivist's contact info. We may want to include an accession agreement document in the media directory of the app and link out to it from the email, or if the information is relatively short, it can be included in the email itself.

danloveg:
Closed by #58

PR #58, opened 2022-01-19
Issue 22 - Add Accession Agreement to Templates

danloveg:
NCTR legal accession agreement has been added to relevant templates.


danloveg:
Closes #22

PR #63, opened 2022-03-11
Centos is dead, use RockyLinux instead

whikloj:
Also remove those unneeded requirements


danloveg:
I didn't know about Rocky Linux, right on! 👍

whikloj:
I had to google to solve this, but people I know that know more about docker than I use [alpine](https://hub.docker.com/_/alpine). It has very little installed in base, so you might need to install more but it is very compact. 

Might be something to think about. 

I'm still running this using the docker instance for the staff. I might continue to do that rather then setup a whole server.

Issue #24, opened 2021-05-03
Add Security and Upload Size Info to About


danloveg:
The about page on the website (https://example.ca/about) should have a section on security (how safe is this application?) and a section on what the maximum size a transfer can be.

The maximum size of a transfer may need to be an environment variable setting so that it is not hardcoded in the template.

danloveg:
Closed by #65

PR #65, opened 2022-03-16
Add "Maximum Transfer Size" to About page

danloveg:
This pull request does not include any information about "security" on the about page. I do not know if this should be included, but the maximum transfer size should obviously be included. If it is deemed necessary to have security info in the about page in the future, I will add it in.


danloveg:
Closes #24

Issue #67, opened 2022-03-16
Fix Permissions not Created Error During Migration


danloveg:
It is possible to have the database in a state where a model is being created, but its `Permission`s have not also been created. When this happens, an error is raised during migration when a permission pertaining to that model is accessed.

This is a simple fix, simply add a call to `django.core.management.sql.emit_post_migrate_signal` in all migrations that are accessing the `Permission` or `Group` model.

danloveg:
Closed by #70

PR #70, opened 2022-03-17
Add emit_post_migrate_signal where it's missing

danloveg:
There were two places where a call to `emit_post_migrate_signal` was needed but it wasn't being called. I've added calls in those locations.


danloveg:
Addresses #67

Issue #76, opened 2022-03-21
Allow users to save/resume a transfer


whikloj:
As the title says, allow users to save a transfer before completing it and come back to it later without having to re-enter any data.

Suggested starting points
> The transfer form is using a [SessionWizardView](https://django-formtools.readthedocs.io/en/latest/wizard.html#formtools.wizard.views.SessionWizardView), which persists the data to the database using a database session. The documentation for Django sessions is [here](https://docs.djangoproject.com/en/3.2/topics/http/sessions/#module-django.contrib.sessions).

danloveg:
Closed by #83. Some work may need to be done in the future to save the current form data when the form is saved. Open a separate issue if this comes up.

PR #83, opened 2022-04-01
Save/Resume a transfer to the database

whikloj:
This is a partial solution to #76. 

What I am missing is the best way to handle resuming a transfer and then re-saving that transfer. Currently if you save 2 transfers, you can resume 1 but you can't save it again. So I need to tell that I am in a resumed transfer and update that one instead of creating a new one.

I think I can use the `?resume_transfer=<int>` in the GET of the URL, but I also was not sure if that was a good idea. So I am stopping here to get your review and input.


whikloj:
Ok, I rebased this and added in checking for the `resume_id` when saving. So it will update the already saved transfer. Some little QOL changes that could be done but maybe could wait for the time-being?
1. **Save Form** button disappears when you already have the maximum number of saved transfers, could just be disabled.
2. The current form data is not saved, would probably need to run the `get_cleaned_data` but without form validation and then serialize it with the existing form data.
3. The deleting of a transfer has no confirmation page, probably a good idea in the long run.

danloveg (changes requested):
Everything looks great, but could you add MAX_SAVED_TRANSFER_COUNT to the table of contents at the top of the `settings.rst` file? I think this would fall under Application Features.

danloveg (changes requested):
Also, can you add validations for the new settings in the `recordtransfer/apps.py` file? I assume any number from zero to infinity could be used, and negative numbers should not be allowed.

If someone sets the value to zero, would this disable saving transfers? Can this be clarified in the `settings.rst` file?

danloveg (approved):
Great!

PR #84, opened 2022-04-06
Add user configurable email updates

### LINKED ISSUE #26: Allow Users to Turn off Emails From Profile Page
A user should be able to manage their email notifications from the Profile page. A new field will likely need to be added to the `models.User` section to allow for this. The field on the user would also need to be checked when an email is sent.

whikloj:
Resolves #26


danloveg:
Looks great, thanks!

Issue #60, opened 2022-01-27
File Contents Should be Checked Before Upload


danloveg:
The contents of every file should be verified with [ClamAV](https://www.clamav.net/) before being accepted.

Required changes:
- Link in [docker-clamav](https://hub.docker.com/r/mkodockx/docker-clamav/) to the docker-compose file so that the files can be checked when running in Docker mode
- Implement `recordtransfer.views._accept_contents` with [clamd](https://pypi.org/project/clamd/)
- Update documentation to include ClamAV setup on production server
- Send an email to administrators for every file that is found to contain malicious content

whikloj:
We have interest in getting the scanning integrated, so I'll take a crack at this.

danloveg:
Sound good. I already "kind of" implemented the logic in `recordtransfer.views._accept_contents`. I also would like to see a [docker-clamav](https://hub.docker.com/r/mkodockx/docker-clamav/) container being added to the `docker-compose.yml` file

danloveg:
> Sound good. I already "kind of" implemented the logic in `recordtransfer.views._accept_contents`. I also would like to see a [docker-clamav](https://hub.docker.com/r/mkodockx/docker-clamav/) container being added to the `docker-compose.yml` file

I didn't even look at the issue, I already wrote all this when I created the issue!

whikloj:
Sorry are we allowing clamav to be installed locally (ie. unix sockets) or remotely (ie. tcp sockets) or both?

whikloj:
Sorry I saw the link to the clamd and thought it was actual clamd. Trying that module now.

PR #85, opened 2022-04-11
Add clamav and scanning

whikloj:
Resolves: #60

The message says that an administrator would be notified, but I did not add sending an email here. As the file is not accepted into the bag, do you still want that to happen and if so what would you like it to say.

For testing this I used the EICAR test file, as in [this example](https://rtcamp.com/tutorials/mail/server/testing/antivirus/)

### DIFF
--- a/bagitobjecttransfer/recordtransfer/views.py
+++ b/bagitobjecttransfer/recordtransfer/views.py
@@ -1,6 +1,7 @@
 from typing import Union
 import logging
 
+import clamd
 from django.contrib import messages
 from django.contrib.auth import login
 from django.http import HttpResponseRedirect, JsonResponse
@@ -17,6 +18,7 @@ from recordtransfer import settings
 from recordtransfer.models import UploadedFile, UploadSession, User, BagGroup, Right, \
     SourceRole, SourceType, Submission
 from recordtransfer.jobs import bag_user_metadata_and_files, send_user_activation_email
+from recordtransfer.settings import CLAMD_HOST, CLAMD_PORT
 from recordtransfer.utils import get_human_readable_file_count, get_human_readable_size
 from recordtransfer.forms import SignUpForm, UserProfileForm
 from recordtransfer.tokens import account_activation_token
@@ -354,9 +356,8 @@ def uploadfiles(request):
             content_check = _accept_contents(_file)
             if not content_check['accepted']:
                 _file.close()
-                raise NotImplementedError(
-                    'file malware check with clamav has not been fully implemented'
-                )
+                issues.append({'file': _file.name, **content_check})
+                continue
 
             new_file = UploadedFile(session=session, file_upload=_file, name=_file.name)
             new_file.save()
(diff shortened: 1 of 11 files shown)


bagitobjecttransfer/recordtransfer/views.py
```diff
@@ -609,22 +610,30 @@ def _accept_contents(file_upload):
...
+            'error': 'Unable to scan file for malware',
             'verboseError': gettext(
-                'The file "{0}" was identified to contain malware! This issue '
-                'will be sent to the administrator'
-            ),
-            'clamav': {
-                'reason': reason
-                'status': status
-            }
+                'Unable to scan the file "{0}" due to exception {1}'.format(file_upload, exc)
```
danloveg, line 636: I don't think we should expose the exception contents to the user. If this occurs in other places we should address those

whikloj: Fair enough, so I'm logging this error already. Do you want to start having a way to send a message to administrators? Also, do we accept the files if Clamav stops working?

danloveg: You should have accession to the `UploadSession` for the transfer. You can call `remove_session_uploads()` on that session to remove all the files that may have been uploaded to the server

whikloj:
Soooo last thing is that currently this is emailing everytime a user hits the Transfer link while ClamAV is inaccessible. 

So _good_ the administrator will not fail to notice the error, but _bad_ we could be filling their inbox. 

Not sure what the best way to handle this is. Some internal flag that tracks the time the first email was sent and only allows an update email after a configurable amount of time.

ie. 
```python

ERROR_EMAIL_PERIOD = 10 * 60 # Ten minutes in seconds
LAST_EMAIL_TIME = None # datetime or None

clamd_socket = clamd.ClamdNetworkSocket(CLAMAV_HOST, CLAMAV_PORT)
try:
    clamd_socket.ping()
except clamd.ClamdError as exc:
    LOGGER.error("Unable to ping ClamAV", exc_info=exc)
    if LAST_EMAIL_TIME = None or LAST_EMAIL_TIME + datetime.timedelta(seconds=ERROR_EMAIL_PERIOD) <= datetime.datetime.now():
        LAST_EMAIL_TIME = datetime.datetime.now()
        send_admin_clamav_error(self.request.user, exc)
```

bagitobjecttransfer/recordtransfer/views.py
```diff
@@ -240,6 +245,17 @@ def get_context_data(self, form, **kwargs):
...
         return context
 
+    def get(self, request, *args, **kwargs):
+        if self.steps.current == 'acceptlegal' and CLAMAV_ENABLED:
+            clamd_socket = clamd.ClamdNetworkSocket(CLAMAV_HOST, CLAMAV_PORT)
+            try:
+                clamd_socket.ping()
+            except clamd.ClamdError as exc:
+                LOGGER.error("Unable to ping ClamAV", exc_info=exc)
+                send_admin_clamav_error(self.request.user, exc)
```
danloveg, line 255: In response to your comment on spamming admins with emails - how do you feel about simply removing this feature? I assume we will have a system in our apps when deployed to production that sends an email to _system_ administrators when an ERROR message is found in the logs.

This approach simplifies the app, and handles the case where maybe the site admins are not the actual system admins. In our case, we're both, but that might not be the case for others.

danloveg (commented):
This all looks great to me. The only issue I see is that you've linked in a system error template, but I don't see it included in the changes you've committed.

I really like the extraction of `_get_admin_recipient_list` into a function.

bagitobjecttransfer/recordtransfer/views.py
whikloj: We could also add a flag and just display a warning message on the admin pages. I always worry about not getting these notifications fast enough. But I'll remove the emailing for now, we can revisit notifications later.

danloveg: Sure, that sounds good to me

PR #91, opened 2022-05-09
Add link back to site profile page

### LINKED ISSUE #89: Unsubscribe from Emails Link does not Work
When an email is sent to a user, there is text saying "If you do not want to receive these emails anymore, click this link." But the link does not work.

Instead of having this text, we should have some text saying "if you do want to receive these emails anymore, log in to your profile to set email preferences." And we can give the users a link to log in to their profile where they can change the setting there (rather than create a new email unsubscribe view).

whikloj:
Resolves #89


danloveg (approved):
Nice and simple, thanks for handling this.

PR #93, opened 2022-05-17
Make captcha invisible for transfer form

### LINKED ISSUE #92: Add ability to disable reCaptcha or switch to invisible reCaptcha
Natalie has asked that we not have a captcha for form submission if the user is already logged in. This seems to make sense to me, but as an alternative I was trying to implement the invisible recaptcha setup. But by default it hijacks the form submission and then requires it for both the "Previous Step" and "Submit" buttons.

I tried to override both the javascript and the widget templates but they don't seem to get picked up. I tried altering the `FORM_RENDERER` setting based on some [suggestions here](https://github.com/torchbox/django-recaptcha/issues/211), but it still is not picked up.

I have considered switching to some other project (like [django-hCaptcha](https://pypi.org/project/django-hCaptcha/) to see if their invisible recaptcha might work, but before I do that I wanted to check to see if a) you have any ideas or b) you might have a reason for using `django-recaptcha`?

whikloj:
Resolves #92 

**Note**: I don't think this is quite ready as the `400.html`, `404.html`, `403.html` and `500.html` were not moved, they probably need to be.

**Note 2**: This only affects the captcha at the end of the transfer form, the sign up captcha still appears as before.


bagitobjecttransfer/recordtransfer/templates/captcha/includes/js_v2_invisible.html
```diff
@@ -0,0 +1,19 @@
+{# The provided implementation caters for only one reCAPTCHA on a page. Override this template and its logic as needed. #}
+<script src="https://{{ recaptcha_domain }}/recaptcha/api.js{% if api_params %}?{{ api_params }}{% endif %}"></script>
+<script type="text/javascript">
+    // Submit function to be called, after reCAPTCHA was successful.
+    var onSubmit_{{ widget_uuid }} = function(token) {
```
danloveg, line 5: I love how you can do stuff like this in Django.

danloveg (approved):
Looks good. We can address the error pages in a separate issue.

PR #95, opened 2022-06-17
Save current form data when saving transfer

whikloj:
This adds on to the Save Transfer ability to also save data from the current form.


danloveg (approved):
Looks great!

PR #98, opened 2022-06-21
Make required fields have a visual difference

### LINKED ISSUE #97: Required form fields be visibly distinct
Currently the only way to know that a form field is required is to submit the form, adding some visual identifier would be good.

It has been suggested to me that some text like
> Mandatory fields are marked with a *

and the asterisk appended to the field label could be bold/larger/red (or all of the those) to make it more visual

I have this working on our branch but wanted to offer it back.

PR incoming.

whikloj:
Resolves #97


danloveg (approved):
This is a great change to have. I like how you've implemented the CSS for the fields; if there's a desire to change how required fields look in the future, there is only one place that needs to be updated.

PR #99, opened 2022-06-21
Source type and source role are required fields

whikloj:
Cleaning up after #98 and realized that these two fields are required by validation but not marked as such (so not getting the red asterisk). 

I couldn't figure out a way to handle the "Other" text fields which are also required (short of making custom templates), but I think the current validation is sufficient.


danloveg:
Good catch! I agree that I don't think we need to worry too much about the "Other" fields, if there's a need later on, we can handle that then.

PR #100, opened 2022-07-04
Initial Bag removal cleanup

whikloj:
I wasn't sure how you wanted to handle my work, but I have this PR which
* Removed the Bag, SourceRole, SourceType and RightsType models from the recordtransfer app
* Added the requested content into the Submission object
* Switched to using the SourceRole, SourceType and RightsType from the CAAIS app in the code
* Makes an initial pass at `bag_user_metadata_and_files` (which is broken, but I can work on that separately)

If you pull this into your issue-78 branch (depending how much different it is from the one up here) the app should run and populate the correct SourceRole and SourceType and you can create a transfer. The actual submission just dies in the django-rq job


whikloj:
Following on this thread, what would be the correct way to instantiate a `Metadata`? It appears like you have to build it up using the parts from the form, or would it make sense to have a method in `Metadata` that takes the `OrderedDict` and tries to put the parts in the correct fields/subfields to generate a whole object?

danloveg:
Thanks for making these changes, I'll review when I'm back from vacation on July 11th. Merging in to the issue-78 branch makes sense to me, I don't mind if things break in that branch while we make this change.

As for instantiating a `Metadata`, there'll definitely be a decent amount of code needed to construct it from the form. I'd hesitate to put this in the `caais` application; I don't want any of the CAAIS models to be coupled with the `recordtransfer` app. I think a "simple" function (simple in that it takes the form as input, and outputs a `Metadata` object) in the recordtransfer app would be preferable to adding a staticmethod to the Metadata model.

PR #105, opened 2022-09-21
Generate model instances from form_data

whikloj:
First pass at integrating some of the new Models into the submission handling. 

Sorry this languished a bit, I was re-using the existing form handling code. So I added a test so I could understand what it was doing.

More to come

PR #108, opened 2022-11-23
Use models to store metadata

whikloj:
This is the contents of branch issue-78 and my PR against that branch (#105), along with some work to merge the migrations. 

I'm closing #105 in favour of this one.


whikloj:
So now I am seeing the issues you are describing. The python migrations have become intermingled and so some are not run. I might need to manually re-order them. 

Might want to start a process of squashing them more regularly when merges happen to the main branch.

whikloj:
Hmmm I guess I'm not sure if we are considering the migrations as actual migrations? Is the expectation that someone that was using this tool before this change could just start using this version? 

But let me know what you'd like to see, I can work to cleaning up and reorganizing the migrations if you want.

Issue #82, opened 2022-03-31
Allow CAAIS Metadata Download without Sending Files


danloveg:
It may be useful for users to be able to generate CAAIS-compliant metadata without sending files. This would be used for when an entity wants to send files _without_ using the transfer application, but still wants to generate the proper metadata for a transfer of files.

danloveg:
Resolved by #138

PR #138, opened 2023-07-28
Addresses Issue 82 (Allow CAAIS Metadata Download without Sending Files)

danloveg:
Add a `FILE_UPLOAD_ENABLED` setting, and allow submissions to be sent without uploading files.


danloveg:
Resolves #82

PR #189, opened 2024-07-30
fix: added executable file permission to entrypoint.sh to be able to run on MacOS

yenaing-oo:
- branch initially contained a separate compose file for development that is compatible with MacOS ARM
- upon further testing of dev and prod containers with both Docker and Podman, it appears to run fine, aside from a few warnings of platform mismatch for some services
- it appears that the only issue before was entrypoint.sh not running, causing the app service to exit


danloveg:
Nice and easy to review! I realize most of the work was testing it out in Podman+Docker on MacOS, so thanks for doing that.

Issue #190, opened 2024-09-18
Testing Needed for move-containerfile-to-root Branch


danloveg:
The following branch has a number of changes that need to be tested in both Windows and ARM-based MacOS: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/tree/move-containerfile-to-root

I wrote this branch on my laptop running Fedora 40, which uses [SELinux](https://www.redhat.com/en/topics/linux/what-is-selinux). Because of this, I had to add `:z` labels to the volumes in the compose files. I don't know if these volume labels will break the configuration in MacOS or Windows. If they *do* break the app such that it can't be run in Windows or MacOS, the labels should be changed back to `:ro`. Supporting SELinux is not one of the requirements of this application, so we don't need to really worry about breaking it so that it doesn't work with SELinux systems.

There is also a new multi-step build used in the `Containerfile`. Verify that the yuglify binary gets built, and that it gets copied to the Python image and is picked up by the `YUGLIFY_BINARY` setting.

Also, the new behaviour needed to run the app needs to be added to the README and to the docs.

yenaing-oo:
The app ran fine in development since static assets were being served directly from their original location without being minified. In production, however, Django couldn't seem to location `manage.py`:

![image](https://github.com/user-attachments/assets/298a36ef-87a8-4521-87c9-70316ed3b2ef)

I looked through the container's files and discovered that the **contents** of `./bagitobjecttransfer` were copied into `/app` rather than the directory iteself:
![image](https://github.com/user-attachments/assets/1c5d710f-39a5-48b7-b334-93a1989e78d6)

I think the Docker's `COPY` instruction copies the contents of a directory instead of the directory. Here's a similar [question ](https://stackoverflow.com/questions/26504846/copy-directory-to-another-directory-using-add-command)on Stackoverflow, but for the `ADD` instruction.

I changed the part in `Containerfile` copying sources files to copy files and directory separately, and this seemed to fix issue of `manage.py` not being found.

The next issue I ran into is that Django keeps trying to serve static assets from  `app/static` rather than `app/bagitobjecttransfer/static`. (I don't know how to recreate the error logs for this, I got another error instead when going back to this point).

I tried changing every possible setting in `base.py` to make it look for static files in the latter, but nothing seemed to work. The only thing that does work is changing `STATIC_ROOT` to the `app/static`.

danloveg:
See pull request - #197

[...]

PR #192, opened 2024-09-23
Move containerfile to root

yenaing-oo:
Regarding https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/190

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -21,7 +21,8 @@ ENV PYTHONUNBUFFERED=1
 WORKDIR /app/
 
 # Copy files to container
-COPY pyproject.toml README.md ./docker/entrypoint.sh ./bagitobjecttransfer/ /app/
+COPY pyproject.toml README.md ./docker/entrypoint.sh /app/
+COPY ./bagitobjecttransfer /app/bagitobjecttransfer
 
 # Copy built binary from builder
 COPY --from=builder /build/node_modules/yuglify/dist/* node_modules/yuglify/dist/
--- a/bagitobjecttransfer/bagitobjecttransfer/settings/docker_prod.py
+++ b/bagitobjecttransfer/bagitobjecttransfer/settings/docker_prod.py
@@ -5,6 +5,10 @@ from decouple import config
 from .base import *
 import re
 
+# Overwrite base.py settings for media and static roots
+MEDIA_ROOT = "/app/media"
+STATIC_ROOT = "/app/static"
+
 DEBUG = False
 SITE_ID = config('SITE_ID', default=1, cast=int)
 


yenaing-oo:
The app ran fine in development since static assets were being served directly from their original location without being minified. In production, however, Django couldn't seem to location `manage.py`:

image

I looked through the container's files and discovered that the **contents** of `./bagitobjecttransfer` were copied into `/app` rather than the directory iteself:
image

I think the Docker's `COPY` instruction copies the contents of a directory instead of the directory. Here's a similar [question ](https://stackoverflow.com/questions/26504846/copy-directory-to-another-directory-using-add-command)on Stackoverflow, but for the `ADD` instruction.

I changed the part in `Containerfile` copying sources files to copy files and directory separately, and this seemed to fix issue of `manage.py` not being found.

The next issue I ran into is that Django keeps trying to serve static assets from  `app/static` rather than `app/bagitobjecttransfer/static`. (I don't know how to recreate the error logs for this, I got another error instead when going back to this point).

I tried changing every possible setting in `base.py` to make it look for static files in the latter, but nothing seemed to work. The only thing that does work is changing `STATIC_ROOT` to the `app/static`.

bagitobjecttransfer/bagitobjecttransfer/settings/base.py
```diff
@@ -151,7 +151,7 @@
 MEDIA_URL = '/media/'
 STATIC_URL = '/static/'
 MEDIA_ROOT = os.path.join(BASE_DIR, 'media')
-STATIC_ROOT = os.path.join(BASE_DIR, 'static/')
+STATIC_ROOT = os.path.join(BASE_DIR, '..', 'static')
```
danloveg, line 154: Can you undo this change, and make the following change to `docker_prod.py`:

```python
# Overwrite base.py settings for media and static roots

MEDIA_ROOT = "/app/media"
STATIC_ROOT = "/app/static"
```

danloveg (changes requested):
You're absolutely right that the static files need to be at `/app/static`, that was a good catch. I even [built that behaviour into the prod compose file](https://github.com/yenaing-oo/Secure-Record-Transfer/blob/move-containerfile-to-root/compose.prod.yml#L91), I must have just forgotten that I had made that decision.

We only need that change in the production configuration, hence why I've asked you to move that change to the base settings file to `docker_prod.py`.

bagitobjecttransfer/bagitobjecttransfer/settings/base.py
yenaing-oo: If we're keeping the path of the static directories fixed, would it make sense to use environment variables for this? We could perhaps use the same environment variable within `compose.prod.yml` and `ngix.conf`.

yenaing-oo:
> 

Ah I think I found the other culprit for this behaviour in [ngix.conf](https://github.com/yenaing-oo/Secure-Record-Transfer/blob/1fc4438dd4d6a3f72ee3aa0fc8131f95bf82ff1c/docker/nginx/nginx.conf#L18-L20). It looks like this is telling the Ngix web server to serve static assets from `app/static`, whereas the line you mentioned in `compose.prod.yml` is defining a shared volume between Django and Ngix.

bagitobjecttransfer/bagitobjecttransfer/settings/base.py
yenaing-oo: > Can you undo this change, and make the following change to `docker_prod.py`:
> 
> ```python
> # Overwrite base.py settings for media and static roots
> 
> MEDIA_ROOT = "/app/media"
> STATIC_ROOT = "/app/static"
> ```

Resolved by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/192/commits/1fc4438dd4d6a3f72ee3aa0fc8131f95bf82ff1c and

danloveg (approved):

Issue #184, opened 2024-07-22
Log handler is missing for clamav


danloveg:
A console log handler should be added to the settings for `clamav`.

yenaing-oo:
Looking at this project's settings, I see that there are two settings files for production: `production.py` and `docker_prod.py`. I was under the impression that in actual production, the production container would be run on a machine to service requests. 

- Is Docker only useful for emulating a production environment?
- How would the different services for this application be set up for actual production, if not using Docker Compose?

danloveg:
> Looking at this project's settings, I see that there are two settings files for production: `production.py` and `docker_prod.py`. I was under the impression that in actual production, the production container would be run on a machine to service requests.
> 
>     * Is Docker only useful for emulating a production environment?
> 
>     * How would the different services for this application be set up for actual production, if not using Docker Compose?

The fact that there's a `production.py` file is a hold-over from the way we were previously running this app (not in containers). I will open a separate issue to remove that file. There may be some references to it elsewhere so it might not be as simple as just removing the file.

danloveg:
Closed by #194.

PR #194, opened 2024-09-25
add log handlers for clamav

yenaing-oo:
Regarding [issue 184](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/184).

Is this everything needed for this issue?


danloveg (approved):
Looks good, yes this all that's needed.

Issue #183, opened 2024-07-22
Timezone should be configurable


danloveg:
The `TIME_ZONE` in the base settings (`base.py`) should be able to be set using an environment variable.

PR #196, opened 2024-09-25
Make timezone configurable

yenaing-oo:
Closes [issue 183](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/183)

I'm not sure how TIMEZONE, and other environment variables such as SECRET_KEY will be set since I could not find them in the `example.*.env` files. Perhaps these will be added later?


danloveg (approved):

danloveg:
> I'm not sure how TIMEZONE, and other environment variables such as SECRET_KEY will be set since I could not find them in the `example.*.env` files. Perhaps these will be added later?

They can be set in the `.env` file, but they're currently not documented. We can document them later.

PR #202, opened 2024-10-02
Add pre-commit hook to add and update linting configuration

yenaing-oo:
### Notes

- The git configuration command to use this repo's git hooks can possibly be automated with Poetry by defining it as an entry script.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+curl --fail -s -S -o "$REPO_ROOT/ruff.toml" \
+    https://raw.githubusercontent.com/NationalCentreTruthReconciliation/Python-Development-Guide/refs/heads/main/ruff.toml
--- a/.gitignore
+++ b/.gitignore
@@ -100,3 +100,4 @@ bagitobjecttransfer/static/
 env/
 package-lock.json
 node_modules
+ruff.toml
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -1,0 +1,5 @@
+## For Developers
+Before beginning development, please configure git to use the git hooks (`.git/hooks`) provided in the repository. This will ensure that the linting configuration is updated before each commit.
+```bash
+git config core.hooksPath .githooks
+```


danloveg (changes requested):
You're on the right track here, but I don't know if it needs to be this complicated. Your pre-commit script could be as simple as this:

```shell
#!/usr/bin/env bash

REPO_ROOT="$(git rev-parse --show-toplevel)"
curl --fail -s -S -o "$REPO_ROOT/ruff.toml" \
    https://raw.githubusercontent.com/NationalCentreTruthReconciliation/Python-DevelopmentGuide/refs/heads/main/ruff.toml
```

And your install script could be as simple as this:

```shell
#!/usr/bin/env bash

if [[ ! -d .git ]]; then
    echo "Could not find .git directory - call this script from the repository root"
    exit 1
fi

if [[ -f .git/hooks/pre-commit ]]; then
    echo "Pre-commit hook already set up at .git/hooks/pre-commit"
    exit 1
fi

if [[ ! -d .git/hooks ]]; then
    mkdir .git/hooks
fi

cat << EOF > .git/hooks/pre-commit
#!/usr/bin/env bash

REPO_ROOT="$(git rev-parse --show-toplevel)"
curl --fail -s -S -o "$REPO_ROOT/ruff.toml" \
    https://raw.githubusercontent.com/NationalCentreTruthReconciliation/Python-Development-Guide/refs/heads/main/ruff.toml
EOF

echo "pre-commit hook written to .git/hooks/pre-commit"
```

yenaing-oo:
This is definitely much cleaner!

I tested it out, and there are is an change and additional steps required to make it work:
- Change `cat << EOF > .git/hooks/pre-commit` to `cat << 'EOF' > .git/hooks/pre-commit` to interpret the `REPO_ROOT` variable when the hook is called, rather than when writing the content of the hook file.
-  After installing:
    - set the permission on the installation shell script to be executable
    - set the permission of the generated `.git/hooks/pre-commit` hook to be executable
    
I think we can do away with the installation script and simply include the `pre-commit` hook with executable permissions and track that. Because git uses the hooks in `.git/hooks` by default, I found a way to configure git to use our custom git hooks instead, by calling:
```shell
git config core.hooksPath <custom_githooks_dir_name>
```

I've made these changes in my PR.

danloveg (changes requested):
Looks great, much simpler!

Can you add a DEVELOPERS.md file that contains the setup needed to get this to work? Likely just calling that git commit. Make sure to include `--local` so it only affects the local git repository.

yenaing-oo:
I believe the default behaviour of `git config` is local. 

https://git-scm.com/docs/git-config#_description

> When writing, the new value is written to the repository local configuration file by default, and options --system, --global, --worktree, --file <filename> can be used to tell the command to write to that location (you can say --local but that is the default).

PR #203, opened 2024-10-03
Auto-build container image and publish to ghcr io

danloveg:
Automatically build the image specified by Containerfile whenever a release is made, or when the workflow is run manually.

I added in pull_request to be able to test this workflow out.


danloveg:
Realized now that this depends on #197. Will need to wait for that to be merged in before testing again.

Issue #198, opened 2024-10-01
Add Ability for Users to Reset Their Own Passwords


danloveg:
It is currently impossible for a logged-in user to reset their own password. This functionality should be added so that it's accessible from the Profile page.

danloveg:
Adding some more detail:

- The `UserProfileForm` should get three new fields:
  - current_password
  - new_password_1
  - new_password_2 (Re-enter the password to confirm)
- `UserProfileForm` should be moved out of `admin_forms.py`, it doesn't make sense for it to be there
- The form should pass validation if none of those fields are filled out. Only if any of them have content should validation run against the fields.
- The form fields should be rendered in a new "password reset" section in the `<form>` in `profile.html`.
- After submit + validation, the user's password should be changed (see: [Changing passwords](https://docs.djangoproject.com/en/3.2/topics/auth/default/#changing-passwords))

PR #208, opened 2024-10-07
Ability for user to reset their password

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/198

Note:
- Current behaviour after resetting password is to keep the user logged in by replacing their session, as opposed to the default behaviour of logging them out of all sessions. Is this the desired behaviour?

### DIFF
--- a/bagitobjecttransfer/recordtransfer/forms/user_forms.py
+++ b/bagitobjecttransfer/recordtransfer/forms/user_forms.py
@@ -12,19 +12,25 @@ class UserProfileForm(forms.ModelForm):
     """Form for editing user profile."""
 
     gets_notification_emails = forms.BooleanField(
-        required=False, label=_("Receive notification emails?")
+        widget=forms.CheckboxInput(attrs={"id": "id_gets_notification_emails"}),
+        label=_("Receive notification emails?"),
+        required=False,
     )
     current_password = forms.CharField(
-        widget=forms.PasswordInput, label=_("Current Password"), required=False
+        widget=forms.PasswordInput(attrs={"id": "current_password"}),
+        label=_("Current Password"),
+        required=False,
     )
     new_password = forms.CharField(
-        widget=forms.PasswordInput,
+        widget=forms.PasswordInput(attrs={"id": "new_password"}),
         label=_("New Password"),
         required=False,
         validators=[password_validation.validate_password],
     )
     confirm_new_password = forms.CharField(
-        widget=forms.PasswordInput, label=_("Confirm New Password"), required=False
+        widget=forms.PasswordInput(attrs={"id": "confirm_new_password"}),
+        label=_("Confirm New Password"),
+        required=False,
     )
 
     class Meta:
@@ -40,47 +46,51 @@ class UserProfileForm(forms.ModelForm):
         new_password = cleaned_data.get("new_password")
         confirm_new_password = cleaned_data.get("confirm_new_password")
 
-        if not current_password:
-            raise ValidationError(
-                {
-                    "current_password": [
-                        _("Required"),
-                    ]
-                }
-            )
-        if not new_password:
-            raise ValidationError(
-                {
-                    "new_password": [
-                        _("Required"),
-                    ]
-                }
-            )
-        if not confirm_new_password:
-            raise ValidationError(
-                {
-                    "confirm_new_password": [
-                        _("Required"),
-                    ]
-                }
-            )
+        if current_password or new_password or confirm_new_password:
+            if not current_password:
+                raise ValidationError(
+                    {
+                        "current_password": [
+                            _("Required"),
+                        ]
+                    }
+                )
+            if not new_password:
+                raise ValidationError(
+                    {
+                        "new_password": [
+                            _("Required"),
+                        ]
+                    }
+                )
+            if not confirm_new_password:
+                raise ValidationError(
+                    {
+                        "confirm_new_password": [
+                            _("Required"),
+                        ]
+                    }
+                )
 
-        if not self.instance.check_password(current_password):
-            raise ValidationError(
-                {
-                    "current_password": [
-                        _("Password is incorrect."),
-                    ]
-                }
-            )
-        if new_password != confirm_new_password:
-            raise ValidationError(
-                {
-                    "confirm_new_password": [
-                        _("The new passwords do not match."),
-                    ]
-                }
-            )
+            if not self.instance.check_password(current_password):
+                raise ValidationError(
+                    {
+                        "current_password": [
+                            _("Password is incorrect."),
+                        ]
+                    }
+                )
+            if new_password != confirm_new_password:
+                raise ValidationError(
+                    {
+                        "confirm_new_password": [
+                            _("The new passwords do not match."),
+                        ]
+                    }
+                )
+
+        if not self.has_changed():
+            raise ValidationError(_("No fields have been changed."))
 
         return cleaned_data
 
--- a/bagitobjecttransfer/recordtransfer/forms/user_forms.py
+++ b/bagitobjecttransfer/recordtransfer/forms/user_forms.py
@@ -3,8 +3,14 @@ from typing import Any
 from django import forms
 from django.contrib.auth import password_validation
 from django.core.exceptions import ValidationError
-from django.utils.translation import ugettext_lazy as _
+from django.utils.translation import gettext_lazy as _
 
+from recordtransfer.constants import (
+    ID_CONFIRM_NEW_PASSWORD,
+    ID_CURRENT_PASSWORD,
+    ID_GETS_NOTIFICATION_EMAILS,
+    ID_NEW_PASSWORD,
+)
 from recordtransfer.models import User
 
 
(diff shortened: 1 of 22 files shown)


bagitobjecttransfer/recordtransfer/views.py
```diff
@@ -15,100 +18,129 @@
...
+        )
         return context
 
     def form_valid(self, form):
-        messages.success(self.request, 'Preferences updated')
-        return super().form_valid(form)
+        response = super().form_valid(form)
+        update_session_auth_hash(self.request, form.instance)
+        messages.success(self.request, "Preferences updated")
+        return response
```
danloveg, line 94: Three things here:

1. If the password was changed, "Preferences updated" should read "Password updated successfully"
2. `update_session_auth_hash` should only be called if the password was changed
3. Can you send an email like in `recordtransfer.admin.CustomUserAdmin.user_change_password` when the user's password is changed? The same email function can be used.

bagitobjecttransfer/recordtransfer/templates/recordtransfer/profile.html
```diff
@@ -1,145 +1,177 @@
...
+            </div>
+            <div class="flex-item">
+                <label>{% trans "Username" %}</label>
+                <input disabled value="{{ user.username }}">
+            </div>
+            <div class="flex-item">
+                <label for="{{ form.gets_notification_emails.id_for_label }}">{% trans "Receive notification emails?" %}</label>
+                {{ form.gets_notification_emails }}
+            </div>
+            <br>
```
danloveg, line 28: Instead of a `<br>` can you add a heading that says "Password Reset"? The same medium-text class as "Past Transfers" can be used.

bagitobjecttransfer/recordtransfer/templates/includes/messages.html
```diff
@@ -0,0 +1,9 @@
+{% if messages %}
+    <div class="messages">
+        {% for message in messages %}
+            <div class="message {{ message.tags }}">
+                {{ message }}
+            </div>
+        {% endfor %}
+    </div>
+{% endif %}
```
danloveg, line 9: Is this file being used? If not, it can be deleted.

bagitobjecttransfer/recordtransfer/forms/user_forms.py
```diff
@@ -0,0 +1,60 @@
...
+    )
+    confirm_new_password = forms.CharField(
+        widget=forms.PasswordInput, label=_("Confirm New Password"), required=False
+    )
+
+    class Meta:
+        """Meta class for UserProfileForm."""
+
+        model = User
+        fields = ("gets_notification_emails",)
```
danloveg, line 33: Add current_password, new_password, and confirm_new_password to `fields`

danloveg (changes requested):
I got distracted by the colourful styles on the messages so I fixed those in a commit.

Other than that, you've got the functionality down exactly right here, you just need to change a few little things for the user experience (send an email, update success message, etc.).

bagitobjecttransfer/recordtransfer/views.py
yenaing-oo: 1) If the user also changed their email notification preference, what message should I display?

danloveg: I would say the password message can override the email notification preference. So if they both change, just show the password changed in the email

bagitobjecttransfer/recordtransfer/forms/user_forms.py
yenaing-oo: From reading about [Meta's field attribute](https://docs.djangoproject.com/en/5.1/topics/forms/modelforms/#selecting-the-fields-to-use), it seems that this is used to select fields **from the specified Model** that we wish to expose to the user so that their values can be modified.

Since `current_password`, `new_password`, and `confirm_new_password` are not part of the User model (perhaps the current User password is but I think Django offers functions like `User.check_password` and `User.set_password` to deal with verification and changing of password), I think the way I'm defining fields for these outside of Meta is appropriate.

danloveg: Good catch, yes, that makes sense. I forgot that that's what Meta was for, it's been a little while since i made a form.

danloveg:
The new logic in the `UserProfileForm` should be tested in a new `test_forms.py` test module inside `recordtransfer.test`. Bonus points if you want to test the new view too but that's less important.

bagitobjecttransfer/recordtransfer/templates/includes/messages.html
yenaing-oo: Yes, this file is being used to display the messages

bagitobjecttransfer/recordtransfer/templates/recordtransfer/profile.html
yenaing-oo: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/208/commits/6842f1408d83a768a44705f2aec5c5a7e3d03798

bagitobjecttransfer/recordtransfer/views.py
yenaing-oo: I added that functionality here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/208/commits/abcb88874eeb72b269501482fce394db2c9c236f

However, I'm not sure how `.delay()` is being used, since I couldn't find any reference to it. Looking at other places where it's used, I see the same thing, such as in `views.CreateAccount`:

image

yenaing-oo: Also, is there an actual email server set up to send emails for testing purposes?

danloveg: Yes, in the development application you can go to localhost:8025 to access any emails - any email sent is intercepted by MailHog so you can view them there

danloveg: > However, I'm not sure how `.delay()` is being used, since I couldn't find any reference to it. Looking at other places where it's used, I see the same thing, such as in `views.CreateAccount`:

See docs here for how `delay()` works. https://github.com/rq/django-rq?tab=readme-ov-file#job-decorator

yenaing-oo:
I have an issue where a JS file (`profile.js`) I created and included previously in `static/recordtransfer/js/base` keeps being served, even though that file has been deleted locally.

image
image

It also appears in the `app/static/...` directory 
image

I tried hard refreshing the browser, deleting the container and rebuilding, but weirdly neither method got rid of that old file.
The correct JS file to use on the Profile page is also being served: `/static/recordtransfer/js/profile/profile.js`, which is good.

You might not be able to recreate this issue.

danloveg:
> I have an issue where a JS file (`profile.js`) I created and included previously in `static/recordtransfer/js/base` keeps being served, even though that file has been deleted locally.
> 
> I tried hard refreshing the browser, deleting the container and rebuilding, but weirdly neither method got rid of that old file. The correct JS file to use on the Profile page is also being served: `/static/recordtransfer/js/profile/profile.js`, which is good.

Which compose file are you using? You can try running this in the app container to clear any old files:

```shell
python manage.py collectstatic --clear
```

yenaing-oo:
> > I have an issue where a JS file (`profile.js`) I created and included previously in `static/recordtransfer/js/base` keeps being served, even though that file has been deleted locally.
> > I tried hard refreshing the browser, deleting the container and rebuilding, but weirdly neither method got rid of that old file. The correct JS file to use on the Profile page is also being served: `/static/recordtransfer/js/profile/profile.js`, which is good.
> 
> Which compose file are you using? You can try running this in the app container to clear any old files:
> 
> ```shell
> python manage.py collectstatic --clear
> ```
> 
> By default, old static files are not removed when collecting static files

I was using the dev compose file. The command you suggested fixed it! Is there any reason we want to retain static files that may no longer be in use? If not, perhaps we can override this default behaviour.

danloveg:
> I was using the dev compose file. The command you suggested fixed it! Is there any reason for retaining static files that may no longer be in use? If not, perhaps we can override this default behaviour, since having old static files can cause conflicts in stylesheets and JavaScript.

Yes, let's make that the default if conflicts are possible without it.

yenaing-oo:
> > I was using the dev compose file. The command you suggested fixed it! Is there any reason for retaining static files that may no longer be in use? If not, perhaps we can override this default behaviour, since having old static files can cause conflicts in stylesheets and JavaScript.
> 
> Yes, let's make that the default if conflicts are possible without it.

I actually couldn't find a setting I can set in Django to clear existing static files before collecting, or to remove any unlinked static files.

yenaing-oo:
> > > > I was using the dev compose file. The command you suggested fixed it! Is there any reason for retaining static files that may no longer be in use? If not, perhaps we can override this default behaviour, since having old static files can cause conflicts in stylesheets and JavaScript.
> > > 
> > > 
> > > Yes, let's make that the default if conflicts are possible without it.
> > 
> > 
> > I actually couldn't find a setting I can set in Django to clear existing static files before collecting, or to remove any unlinked static files.
> 
> It's in the entrypoint script

Ah I see. I can just add the flag `--clear` to the `collectstatic` command there.

bagitobjecttransfer/recordtransfer/views.py
```diff
@@ -15,100 +18,159 @@
...
+
     def form_valid(self, form):
-        messages.success(self.request, 'Preferences updated')
-        return super().form_valid(form)
+        response = super().form_valid(form)
+        success_message = "Preferences updated"
+
+        if form.cleaned_data.get("new_password"):
+            update_session_auth_hash(self.request, form.instance)
+            success_message = "Password updated"
```
danloveg, line 114: "Password updated" should be surrounded by `gettext()`.

bagitobjecttransfer/recordtransfer/views.py
```diff
@@ -15,100 +18,159 @@
...
+            }
+            send_user_account_updated.delay(self.get_object(), context)
+
+        messages.success(self.request, success_message)
+        return response
+
+    def form_invalid(self, form):
+        messages.error(
+            self.request,
+            "There was an error updating your preferences. Please check the form and try again.",
```
danloveg, line 129: This message should also be surrounded by `gettext()`.

danloveg (changes requested):
Looks great. The tests are succeeding on my machine, and the user experience with changing a password feels right.

Just noticed two small issues where `gettext()` is missing. We don't have any translations yet, but I want to be sure that all translatable messages do have `gettext()` in case we add translations in the future.

bagitobjecttransfer/recordtransfer/views.py
yenaing-oo: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/208/commits/fd426e65267fdd2f018dc0cc8f56073b7b0271ec

bagitobjecttransfer/recordtransfer/views.py
yenaing-oo: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/208/commits/fd426e65267fdd2f018dc0cc8f56073b7b0271ec

danloveg (approved):

Issue #201, opened 2024-10-02
Missing "stylesheet" Template Tag Load in submission_detail.html


danloveg:
The submission detail report fails to be rendered due to this error:

`Invalid block tag on line 6: 'stylesheet'. Did you forget to register or load this tag?`

This can be addressed by loading that template tag into the template with `{% load pipeline %}`.

PR #209, opened 2024-10-07
Fix missing stylesheet for submission detail page

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/201

### Notes:

I've opted to create a similar template to `base_css.html`, called `submission_detail_css.html`, which I load into `submission_detail.html`.

I think this would be better than to include that stylesheet in `base_css.html`, since only one template will be using `recordtransfer_submission_detail_styles`.


danloveg (approved):
This makes a lot more sense than simply including `{% load pipeline %}`.

Good catch on including the title and meta in the head of submission_detail too.

PR #214, opened 2024-10-18
Address Broken Debugging Configuration

danloveg:
I updated the required launch.json to debug the app, and moved it to DEVELOPERS.md. There is now also a link from the Debugging setup part of the docs to the DEVELOPERS.md file.

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -70,7 +70,7 @@ Ensure that you've installed the `dev` dependencies locally (preferably in a vir
 pip install .[dev]
 ```
 
-If using VSCode for development, use these settings so that tests are discovered correctly:
+Use these settings in your [VSCode settings.json](https://code.visualstudio.com/docs/getstarted/settings#_settings-json-file) so that tests are discovered correctly in the Testing menu:
 
 ```json
 {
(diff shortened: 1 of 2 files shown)


yenaing-oo (dismissed):
Everything looks good!

danloveg:
Had to fix a merge conflict. I think since those settings in setting.json would be referred to in the Python Style Guide that there's no need to include them again in the DEVELOPERS.md file

DEVELOPERS.md
```diff
@@ -18,16 +74,6 @@ If using VSCode for development, use these settings so that tests are discovered
```
yenaing-oo, line 74: Would it be a good idea to advise the developer to include this in the local `settings.json` file _for this project_?

yenaing-oo (changes requested):
Just one comment

DEVELOPERS.md
danloveg: Good point, I agree with you. Also, since I said at the top of the file we're assuming you're using VSCode, I removed "If using VSCode"

Resolved by 182d9f2

danloveg: I skimmed your message too quickly, I also clarified that those settings are for the *local* settings in 47bffbb

yenaing-oo: Yes, that makes sense.

yenaing-oo (approved):

PR #223, opened 2024-10-22
Upgrade Django to 4.2 LTS

danloveg:
This upgrades Django to version 4.2, the current LTS release.

[django-upgrade](https://github.com/adamchainz/django-upgrade) was used to upgrade all Python files to 4.2 compatible code. This command was used to do the upgrade:

```powershell
git ls-files -- "*.py" | ForEach-Object { django-upgrade --target-version 4.2 $_ }
```

Related dependencies were also upgraded, importantly `django-recaptcha` and `gunicorn`.

The ignored `version:` element in the compose files has been removed, [since it is now obsolete](https://docs.docker.com/reference/compose-file/version-and-name/#version-top-level-element-obsolete).

Firefox was complaining that the page was in [Quirks mode](https://developer.mozilla.org/en-US/docs/Web/HTML/Quirks_Mode_and_Standards_Mode) due to a missing `<!DOCTYPE html>`, so that has been added to the base template as well.


yenaing-oo (approved):
Everything looks good! The pages that require CSRF tokens work fine now in production, and the styling of the footer is fixed.

Everything else functions correctly too,

danloveg:
> Everything looks good! The pages that require CSRF tokens work fine now in production, and the styling of the footer is fixed.
> 
> Everything else functions correctly too.

Awesome, thanks for reviewing.

PR #224, opened 2024-10-23
Enable file upload tests regardless of settings

danloveg:
Enables disabled tests. I've also added documentation for running the tests outside of VSCode,


yenaing-oo (approved):
I tested out the external command to run tests, and it worked fine.

Everything else looks good too.

Issue #222, opened 2024-10-22
Group Handling Overhaul


danloveg:
The way the group form is set up right now is brittle and is prone to breaking.

https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/ec155f4c86937167e2bd379b0001e6c6b4fc0365/bagitobjecttransfer/recordtransfer/forms/transfer_forms.py#L659

I will suggest that the whole group selection flow is changed, so that:

- It is not possible to add a new group from the transfer form. This will greatly simplify the logic and JavaScript for this form.
- The SubmissionGroup model should have a UUID field, the way Submission does.
- The users's groups should be rendered on the profile page, below Past Transfers.
- A new view for group/uuid/ should be added that shows a list of all submissions for a group, the same way it's being rendered from the profile page.
- A new view for group/new/ should be added that lets a user add a new group from their profile.

I've made these samples in the dev console with how I'm imagining these changes looking:

# The profile page (URL: /profile/)

The profile page will need a new table like this:

![profile_html](https://github.com/user-attachments/assets/f2d07679-3e59-4bcf-a709-c158ab141239)

The links in the **View group** column will take you to ...

# The group view page (URL: /group/uuid/ - New View)

There should be a new group page that will display all past transfers related to that group. It will be dual-function like the current profile page is, meaning, it will be used to display a related set of models, and can be used to change the underlying object.

I've omitted the name field but that should be able to be changed in this form as well.

![group_view_html](https://github.com/user-attachments/assets/f4154ff1-3927-4ecc-8e51-2a3142d41cdc)

# The group add page (URL: /group/new/ - New View)

There should be a new page to add a new group. The link on the profile page **Click to add a new group** should take you to this page.

![group_add_html](https://github.com/user-attachments/assets/d19fef4c-b328-41f8-aaa9-bec2e0eb7d7f)

danloveg:
Since a user may still want to add a new group while filling out a form, we can render the same form as /group/new/ inside a modal.

The JavaScript for rendering modals is already included, see for example this template:

https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/ec155f4c86937167e2bd379b0001e6c6b4fc0365/bagitobjecttransfer/recordtransfer/templates/recordtransfer/transferform_sourceinfo.html#L18

Which gets rendered as a modal, here:

[...]

PR #226, opened 2024-10-23
Group Handling Overhaul

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/222


bagitobjecttransfer/recordtransfer/forms/transfer_forms.py
```diff
@@ -659,57 +662,35 @@ def clean(self):
...
-            self.add_error('new_group_name', 'Group name cannot be empty')
-        if group_name == 'Add New Group' and not cleaned_data['group_description']:
-            self.add_error('group_description', 'Group description cannot be empty')
+        group_id = cleaned_data.get('group_id')
+
+        users_groups = SubmissionGroup.objects.filter(created_by=self.user)
+        valid_group_ids = users_groups.values_list('uuid', flat=True)
+        if group_id and group_id not in valid_group_ids:
+            self.add_error('group_id', 'Invalid group selected.')
         return cleaned_data
```
danloveg, line 680: I suggest simplifying this logic to:

```suggestion
        if group_id and not SubmissionGroup.objects.filter(created_by=self.user, uuid=group_id).exists():
            self.add_error('group_id', 'Invalid group selected.')
        return cleaned_data
```

bagitobjecttransfer/recordtransfer/static/recordtransfer/js/base/forms.js
```diff
@@ -413,12 +459,46 @@ $(() => {
...
     })
+
+    /***************************************************************************
+     * Group Description Display
+    **************************************************************************/
+    // Add a change event listener to the group selection field
+    const selectField = $('#' + ID_SUBMISSION_GROUP_SELECTION);
+    selectField.on('change', updateGroupDescription);
+    
+    initializeGroupTransferForm();
```
danloveg, line 470: I'm still seeing an error in the console, maybe try only running this code when that variable exists, like this:

```suggestion
    if (typeof ID_SUBMISSION_GROUP_SELECTION !== 'undefined') {
        const selectField = $('#' + ID_SUBMISSION_GROUP_SELECTION);
        selectField.on('change', updateGroupDescription);
        initializeGroupTransferForm();
    }
```

bagitobjecttransfer/recordtransfer/static/recordtransfer/js/base/forms.js
```diff
@@ -413,12 +459,46 @@ $(() => {
...
+            url: $(this).attr('action'),
+            type: $(this).attr('method'),
+            data: $(this).serialize(),
+            success: async function (response) {
+                try {
+                    handleNewGroupAdded(response.group);
+                    $('#add-new-group-dialog').dialog("close");
+                }
+                catch (error) {
+                    alert("Failed to update group options.");
```
danloveg, line 488: This error is popping up on the `/submission_group/uuid/` page when I try to make an edit to an existing group - I think this is because the ID of the form is the same in the modal and the edit form.

bagitobjecttransfer/recordtransfer/templates/includes/profile_form.html
```diff
@@ -0,0 +1,60 @@
+{% load i18n %}
+{% include "includes/messages.html" %}
+<script>
+    // Pass constants to JavaScript
+    const ID_GETS_NOTIFICATION_EMAILS = "{{ ID_GETS_NOTIFICATION_EMAILS }}";
+    const ID_CURRENT_PASSWORD = "{{ ID_CURRENT_PASSWORD }}";
+    const ID_NEW_PASSWORD = "{{ ID_NEW_PASSWORD }}";
+    const ID_CONFIRM_NEW_PASSWORD = "{{ ID_CONFIRM_NEW_PASSWORD }}";
+</script>
```
danloveg, line 9: This will place your JS in the head of the document which would be preferred to having a random <script> in the middle of the document.

```suggestion
{% block javascript %}
    {{ block.super }}
    <script type="text/javascript">
        // Pass constants to JavaScript
        const ID_GETS_NOTIFICATION_EMAILS = "{{ ID_GETS_NOTIFICATION_EMAILS }}";
        const ID_CURRENT_PASSWORD = "{{ ID_CURRENT_PASSWORD }}";
        const ID_NEW_PASSWORD = "{{ ID_NEW_PASSWORD }}";
        const ID_CONFIRM_NEW_PASSWORD = "{{ ID_CONFIRM_NEW_PASSWORD }}";
    </script>
{% endblock %}
```

bagitobjecttransfer/recordtransfer/templates/includes/submission_group_form.html
```diff
@@ -0,0 +1,26 @@
+{% load i18n %}
+<script>
+    // Pass IDs of form fields elements to JavaScript
+    const ID_SUBMISSION_GROUP_DESCRIPTION = "{{ ID_SUBMISSION_GROUP_DESCRIPTION }}";
+    const ID_SUBMISSION_GROUP_NAME = "{{ ID_SUBMISSION_GROUP_NAME }}";
+</script>
```
danloveg, line 6: Same suggestion as above:

```suggestion
{% block javascript %}
    {{ block.super }}
    <script>
        // Pass IDs of form fields elements to JavaScript
        const ID_SUBMISSION_GROUP_DESCRIPTION = "{{ ID_SUBMISSION_GROUP_DESCRIPTION }}";
        const ID_SUBMISSION_GROUP_NAME = "{{ ID_SUBMISSION_GROUP_NAME }}";
    </script>
{% endblock %}
```

bagitobjecttransfer/recordtransfer/templates/recordtransfer/transferform_group.html
```diff
@@ -1,45 +1,45 @@
...
 {% load i18n %}
-
 {% block buttonarrayclass %}flex-item{% endblock %}
 {% block formfields %}
+    <script>
+        const ID_DISPLAY_GROUP_DESCRIPTION = "{{ ID_DISPLAY_GROUP_DESCRIPTION }}";
+        const ID_SUBMISSION_GROUP_SELECTION = "{{ ID_SUBMISSION_GROUP_SELECTION }}";
+        const fetchUsersGroupDescriptionsUrl = "{% url 'recordtransfer:get_users_group_descriptions' user_id=request.user.id %}";
+        const DEFAULT_GROUP_ID = {% if DEFAULT_GROUP_ID is not None %}"{{ DEFAULT_GROUP_ID }}"{% else %}null{% endif %};
+    </script>
```
danloveg, line 11: I suggest moving this to `{% block javascript %}` as well. Make sure it's not in `{% block formfields %}`

Also: rename `'recordtransfer:get_users_group_descriptions'` to the new name of that URL (see urls.py).

bagitobjecttransfer/recordtransfer/urls.py
```diff
@@ -58,6 +62,21 @@
...
+    path(
+        "submission_group/<uuid:uuid>/",
+        login_required(views.SubmissionGroupDetailView.as_view()),
+        name="submissiongroupdetail",
+    ),
+    path(
+        "user/<int:user_id>/group-descriptions/",
+        login_required(views.get_users_group_descriptions),
+        name="get_users_group_descriptions",
+    ),
```
danloveg, line 79: Suggestion to make the URL a bit more uniform with the other URLs:

```suggestion
    path(
        "user/<int:user_id>/submission_groups/",
        login_required(views.get_user_submission_groups),
        name="user_submission_groups",
    ),
```

bagitobjecttransfer/recordtransfer/views.py
```diff
@@ -947,3 +952,135 @@ def get(self, request, *args, **kwargs):
...
+        error_message = next(iter(form.errors.values()))[0]
+        if "transfer" in referer:
+            return JsonResponse({"message": error_message, "status": "error"}, status=400)
+        messages.error(
+            self.request,
+            self.error_message,
+        )
+        return super().form_invalid(form)
+
+def get_users_group_descriptions(request: HttpRequest, user_id: int) -> JsonResponse:
```
danloveg, line 1074:
```suggestion
def get_user_submission_groups(request: HttpRequest, user_id: int) -> JsonResponse:
```

bagitobjecttransfer/recordtransfer/views.py
```diff
@@ -536,32 +558,14 @@ def get_submission_group(self, cleaned_form_data: dict) -> Optional[SubmissionGr
...
-                    group = SubmissionGroup.objects.get(
-                        name=group_name, created_by=self.request.user
-                    )
-                    LOGGER.info('Associating Submission with "%s" SubmissionGroup', group.name)
-
-                except SubmissionGroup.DoesNotExist as exc:
-                    LOGGER.error('Could not find "%s" SubmissionGroup', group.name, exc_info=exc)
-                    group = None
+            except SubmissionGroup.DoesNotExist as exc:
+                LOGGER.error('Could not find SubmissionGroup with ID %d', group_id, exc_info=exc)
```
danloveg, line 568: Small tweak to error message (since it's a UUID, not an ID, which someone might confuse for a primary key).

Also, this will cause an error because `%d` is for numerical data, and `group_id` is a string now.

```suggestion
                LOGGER.error('Could not find SubmissionGroup with UUID %s', group_id, exc_info=exc)
```

danloveg (changes requested):
Functionality is looking great, except for when I try to change an existing group from the group change form. I think the AJAX request is intercepting that group's form submit and it's causing an error.

bagitobjecttransfer/recordtransfer/templates/includes/profile_form.html
yenaing-oo: Could you check if this works for you?

I'm getting back this rendering error:
```
Error during template rendering
In template /app/bagitobjecttransfer/recordtransfer/templates/includes/profile_form.html, error at line 4

'BlockNode' object has no attribute 'context'. Did you use {{ block.super }} in a base template?
1	{% load i18n %}
2	{% include "includes/messages.html" %}
3	{% block javascript %}
4	    {{ block.super }}

```

I think it might be because the `javascript` block is defined in `base.html`, which `profile.html` is extending. `profile_form.html` is included in `profile.html`, but maybe `profile_form.html` doesn't have access to `base.html`'s context unless it is directly extending it?

bagitobjecttransfer/recordtransfer/static/recordtransfer/js/base/forms.js
yenaing-oo: This works! I've applied your suggestion here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/226/commits/eff87b112dc8234775c6f7ae623db8fed1808561. However, I do think that it would be a good idea to:

- split the relevant JS for each form step into separate files and load them where needed (and)
- remove `forms.js` from `static/recordtransfer/js/base`

This way, only the relevant JS is loaded for each page, and we don't have to avoid running portions of JS if constants are missing, since we expect them to be there.

bagitobjecttransfer/recordtransfer/forms/transfer_forms.py
yenaing-oo: Much cleaner, I like it.

Resolved by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/226/commits/2078c7cb0ee06ff3cd8373e6d6a6ae63c587c235

bagitobjecttransfer/recordtransfer/static/recordtransfer/js/base/forms.js
yenaing-oo: The error was actually coming from the validation in `SubmissionGroupForm`. It was raising a validation error for having a group name that already exists, which happens when I try to modify an existing group by only changing the description.

I've fixed it to exclude comparing names with itself:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/226/commits/2078c7cb0ee06ff3cd8373e6d6a6ae63c587c235

bagitobjecttransfer/recordtransfer/views.py
yenaing-oo: Good catch.

Resolved by:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/226/commits/80ff0ff2cb064cd3099c525980ff554d3c2a2627

bagitobjecttransfer/recordtransfer/urls.py
yenaing-oo: I am using this path to send back specifically submission group **descriptions**. Would it be better to send back entire groups instead?

bagitobjecttransfer/recordtransfer/views.py
yenaing-oo: Same concern as above.

bagitobjecttransfer/recordtransfer/static/recordtransfer/js/base/forms.js
danloveg: Feel free to make that change, that makes sense to me

bagitobjecttransfer/recordtransfer/templates/includes/profile_form.html
danloveg: If it's an issue with extending a block that doesn't exist in the extended form, maybe it would work to add the block in profile.html without changes.

```
# (in profile.html)
{% block javascript %}
    {{ block.super }}
{% endblock %}
```

bagitobjecttransfer/recordtransfer/static/recordtransfer/js/base/forms.js
yenaing-oo: I think it would be better to do this in a separate issue. I'll open one up for this.

bagitobjecttransfer/recordtransfer/templates/includes/profile_form.html
yenaing-oo: Resolved by:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/226/commits/bf1d9d5a012c5a2c091b4604b858c9350c06a4b3

bagitobjecttransfer/recordtransfer/templates/includes/submission_group_form.html
yenaing-oo: Resolved by:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/226/commits/bf1d9d5a012c5a2c091b4604b858c9350c06a4b3

bagitobjecttransfer/recordtransfer/templates/recordtransfer/transferform_group.html
yenaing-oo: Resolved by:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/226/commits/bf1d9d5a012c5a2c091b4604b858c9350c06a4b3
and 
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/226/commits/b60e49d6b068284632a665faf533ec11b1fd0a0b

bagitobjecttransfer/recordtransfer/urls.py
yenaing-oo: Resolved by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/226/commits/b60e49d6b068284632a665faf533ec11b1fd0a0b

bagitobjecttransfer/recordtransfer/views.py
yenaing-oo: Resolved by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/226/commits/b60e49d6b068284632a665faf533ec11b1fd0a0b

bagitobjecttransfer/recordtransfer/static/recordtransfer/js/base/forms.js
```diff
@@ -243,17 +244,88 @@ function removeOther(id) {
...
+    updateGroupDescription();
+}
+
+/**
+ * Handles the addition of a new group to the selection field.
+ *
+ * @param {Object} group - The group object containing details of the new group.
+ * @param {string} group.name - The name of the new group.
+ * @param {string} group.uuid - The unique identifier of the new group.
+ * @param {string} group.description - The description of the new group.
```
danloveg, line 344: This is a confusing block, shouldn't `@param` correspond to actual arguments in the function? It's probably preferable to note that these are required attributes of the group object, not that they are each a separate `@param`

yenaing-oo:
> Functionality is looking great, except for when I try to change an existing group from the group change form. I think the AJAX request is intercepting that group's form submit and it's causing an error.

I see how this is happening now. Strangely, this issue was not present for me when testing initially. I've made a temporary fix for now, but I will be replacing this with a cleaner solution when working on issue https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/235/.

Resolved by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/226/commits/6bad910f0e4da387ad9f10b24308753ad1a8b4e3

danloveg (changes requested):
I'm still getting an error when updating a group from the group change page.

If I change the description and click "Save" I get an alert saying `Failed to update group options.`. I logged the error to the console and saw this:

Screenshot 2024-11-08 132827

Are you certain that `forms.js` needs to be loaded on the group change page? That AJAX query is still intercepting Django's default form submit behaviour.

yenaing-oo:
> I'm still getting an error when updating a group from the group change page.
> 
> If I change the description and click "Save" I get an alert saying `Failed to update group options.`. I logged the error to the console and saw this:
> 
> Screenshot 2024-11-08 132827
> 

I believe that issue should be fixed now. Perhaps your browser is using the old JS? Did you call `collectstatic` before hard refreshing the browser page?

> Are you certain that `forms.js` needs to be loaded on the group change page? That AJAX query is still intercepting Django's default form submit behaviour.

No, `forms.js` does not need to be loaded on the group change page, or on any page that does not involve the transfer form. But since `forms.js` has been included in `recordtransfer/static/js/base`, I've been working with having it being loaded on every page. This is why I think it's a good idea to remove `forms.js` from `base`, as mentioned in https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/235.

bagitobjecttransfer/recordtransfer/static/recordtransfer/js/base/forms.js
yenaing-oo: I agree. I've fixed that here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/226/commits/c2c1bca3ca2a6be9ef485fd3eec5a10a6d8b95d5

danloveg:
> > I'm still getting an error when updating a group from the group change page.
> > If I change the description and click "Save" I get an alert saying `Failed to update group options.`. I logged the error to the console and saw this:
> > Screenshot 2024-11-08 132827
> 
> That's strange, it's working fine on my end. Perhaps your browser is using the old JS? Did you call `collectstatic` before hard refreshing the browser page?

I tried that and am still getting the issue. Can you try updating the group description in Firefox?

> > Are you certain that `forms.js` needs to be loaded on the group change page? That AJAX query is still intercepting Django's default form submit behaviour.
> 
> No, `forms.js` does not need to be loaded on the group change page, or on any page that does not involve the transfer form. But since `forms.js` was already included in `recordtransfer/static/js/base` before I began working on this issue, I've been working with it being loaded on every page. This is why I think it's a good idea to remove `forms.js` from `base`, as mentioned in #235.

I understand now, OK that makes sense. I still think this one issue needs to be resolved before we merge this in, still trying to figure out if it's just me or not.

danloveg:
Wait, now it's working. There were issues with my repo pulling the latest code, but now I've got your most recent changes and it's working well.

danloveg (changes requested):
All the changes look good!

I have one more ask, because we need to make sure this integrates nicely with the Django 4.2 changes on the master branch.

Can you do these steps:

- Merge `master` into this branch (and fix any conflicts if there are any)
- *Locally*, install the dependencies in a virtual environment if they aren't already installed.
- Run `pip install django-upgrade` in your local environment.
- Run `git ls-files -z -- '*.py' | xargs -0 django-upgrade --target-version 4.2` from the root of the repo in your terminal.
- Commit any changes made by that tool. It upgrades Python files in-place. It *might* need to make changes but I don't expect there will be many since I already ran that tool against the code on the master branch.
- Test it again and make sure it works.

I remember django-upgrade formatting code differently to how Ruff does, so if it formats the code, make sure you don't commit those changes - you may need to re-format the formatting it does.

yenaing-oo:
> * git ls-files -z -- '*.py' | xargs -0 django-upgrade --target-version 4.2

Only one thing was modified by this command. I made sure I formatted it with Ruff before committing.

Resolved by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/226/commits/51071b75a9d47dcec22c61e98c848070b55c21cc

danloveg (approved):
Thumbs up, this looks good to go

PR #227, opened 2024-10-25
Enable Metadata Instances to be Created via Admin

### LINKED ISSUE #225: CAAIS Metadata Objects Can't be Created Manually
CAAIS Metadata objects can't be created manually from the admin site due to this error:

`save() prohibited to prevent data loss due to unsaved related object`

A `save()` call should be added before this line: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/ec155f4c86937167e2bd379b0001e6c6b4fc0365/bagitobjecttransfer/caais/forms.py#L139 to save the Metadata model before saving the identifier linked to the metadata object.

danloveg:
New Metadata objects can now be created manually from the admin backend at this page:

/admin/caais/metadata/add/

The issue was in accessing the related set `identifiers` on the Metadata object before the object was being saved.

Closes #225


bagitobjecttransfer/caais/forms.py
```diff
@@ -132,10 +132,13 @@ def save(self, commit=True):
         ''' Save the accession identifier input as an Identifier with the type
         name "Accession Identifier"
         '''
-        metadata = super().save(commit=commit)
+        metadata: Metadata = super().save(commit=commit)
 
         accession_id = self.cleaned_data['accession_identifier']
-        if not metadata.accession_identifier:
+
+        if accession_id and not metadata.accession_identifier:
```
yenaing-oo, line 139: I think this block will be easier to understand if it's refactored like this:
```
if accession_id:
    if accession_id != metadata.accession_identifier:
        if not metadata.pk:
            metadata.save()  # Ensure the metadata instance is saved to get a primary key

        if not metadata.accession_identifier:
            # Create a new Identifier if it doesn't exist
            Identifier.objects.create(
                metadata=metadata,
                identifier_type='Accession Identifier',
                identifier_value=accession_id,
            )
        else:
            # Update the existing accession identifier
            metadata.update_accession_id(accession_id)
```

bagitobjecttransfer/caais/models.py
```diff
@@ -218,7 +218,8 @@ class Meta:
 
     @property
     def accession_identifier(self) -> str:
-        if self.identifiers:
+        # We check self.pk here in case the model is not saved yet
+        if self.pk and self.identifiers.count() > 0:
             accession_id = self.identifiers.accession_identifier()
```
yenaing-oo, line 223: I understand that a reverse relationship is being used to get all the `Identifier` models are related to this instance of `Metadata`. However, I could not find the method `accession_identifier()` defined in the `Identifier` model class, so I'm not sure how `self.identifiers.accession_identifier()` is being called.

yenaing-oo (changes requested):

bagitobjecttransfer/caais/models.py
danloveg: Great question! If `metadata` is an instance of `caais.models.Metadata`, then `metadata.identifiers` is an instance of `caais.managers.IdentifierManager`. The identifiers property is actually a manager rather than a queryset of Identifier objects. The manager for the model is set here:

https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/86ecc3cdb5bf4524bfb65733ac5583484b76993e/bagitobjecttransfer/caais/models.py#L466

And you can see that `accession_identifier()` is indeed a method defined on that manager:

https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/86ecc3cdb5bf4524bfb65733ac5583484b76993e/bagitobjecttransfer/caais/managers.py#L55

bagitobjecttransfer/caais/forms.py
danloveg: How does this look? [d15c317](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/227/commits/d15c3176ba4d211a8dfbc8d6de34e734968aae1f)

I generally prefer to avoid nested if statements, let me know if the new version is more readable.

danloveg:
I also addressed #231 here.

bagitobjecttransfer/caais/models.py
yenaing-oo: I see, thanks for clarifying!

bagitobjecttransfer/caais/forms.py
yenaing-oo: I agree, a triple nested if statement is not ideal. Yes, the new version is more readable.

I have a different concern - when the method `save` is called with `commit=False`, this part of the method:
```
if not metadata.pk:
            metadata.save()
```
commits to the save regardless of whether `commit` is `True` or `False`.

I made some suggested changes on some lines, could you check check whether the changes would reflect the desired behaviour?

yenaing-oo (changes requested):
Some changes to address the mismatch in behaviour in save() when commit is `False`

bagitobjecttransfer/caais/models.py
yenaing-oo: I see, thanks for clarifying!

bagitobjecttransfer/caais/forms.py
yenaing-oo: I accidentally deleted my comment for this, so I've written it here again.

I agree that a triple-nested if statement is not ideal. Yes, the new version is more readable.

I have a different concern, which I've written below.

bagitobjecttransfer/caais/forms.py
```diff
@@ -128,22 +128,33 @@ class Meta:
...
+
+        accession_id = self.cleaned_data["accession_identifier"]
+
+        # No changes required - return early
+        if not accession_id or accession_id == metadata.accession_identifier:
+            return metadata
+
+        # The metadata needs to be saved before updating the accession identifier if it hasn't
+        # already been saved
+        if not metadata.pk:
```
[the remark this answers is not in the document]
danloveg: You're absolutely right, there is. I addressed this in [4256002](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/227/commits/4256002fc4f17b94b7fab1c43228e5194a69d285)

bagitobjecttransfer/caais/forms.py
```diff
@@ -128,22 +128,33 @@ class Meta:
...
 
-        accession_id = self.cleaned_data['accession_identifier']
         if not metadata.accession_identifier:
             new_id = Identifier(
                 metadata=metadata,
-                identifier_type='Accession Identifier',
+                identifier_type="Accession Identifier",
                 identifier_value=accession_id,
             )
             new_id.save()
```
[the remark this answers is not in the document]
danloveg: See: [4256002](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/227/commits/4256002fc4f17b94b7fab1c43228e5194a69d285)

bagitobjecttransfer/caais/forms.py
```diff
@@ -128,22 +128,33 @@ class Meta:
...
 
-        elif accession_id != metadata.accession_identifier:
+        else:
             metadata.update_accession_id(accession_id)
```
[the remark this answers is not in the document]
danloveg: See: [4256002](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/227/commits/4256002fc4f17b94b7fab1c43228e5194a69d285)

yenaing-oo (approved):

PR #230, opened 2024-10-25
Allow ACCEPTED_FILE_FORMATS to be Specified in Config

danloveg:
I didn't end up going with a YAML format, but defined a new data structure that can be defined as a string, and parsed as a list of file extensions grouped by file type.

The accepted file types can now be specified like this:

ACCEPTED_FILE_FORMATS="Image:png,jpg|Spreadsheet:csv,xlsx,ods|PDF:pdf"

There is some pretty robust error checking and edge case handling when this string gets parsed.

The parser is in a new file called configuration.py, where any future custom config parsers can go.

To test that the data structure gets parsed correctly, you can go to the /about/ page, where a visual representation of the ACCEPTED_FILE_FORMATS variable is shown.


bagitobjecttransfer/recordtransfer/configuration.py
```diff
@@ -0,0 +1,95 @@
...
+                if not ext_match:
+                    raise ValueError(
+                        f"the extension '{raw_extension}' in the '{name}' file group does not "
+                        "look like a file extension - it must only contain letters and numbers, "
+                        "and may start with a period"
+                    )
+
+                extension = ext_match.group("extension").lower()
+
+                if extension in all_extensions and extension not in accepted_types[name]:
```
yenaing-oo, line 86: Is it the desired behaviour to accept duplicate file extensions within the same file group?
For e.g. "Image : jpg, .jpg"

As `accepted_types[name]` is a set, this won't be an issue, but I was just wondering if this is something we want to let the user know about by throwing an error.

yenaing-oo (commented):
Just had one comment.

yenaing-oo (dismissed):

bagitobjecttransfer/recordtransfer/configuration.py
danloveg: That's an edge case that wasn't tested, thanks for identifying that.

This edge case is automatically handled because we're adding to a `set`, and I don't see a need to show a warning. See the test I added here: [699ea35](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/230/commits/699ea3540dac6e2bd1d3c313dec0cd2ad67e7f9f)

yenaing-oo: This makes sense. Thanks for adding a test for that.

yenaing-oo (approved):

Issue #245, opened 2024-11-27
Minified static assets (JS and CSS) are being generated for development


yenaing-oo:
Minified CSS and JS should not be generated when running in development. This causes issues such as the same JS running 3 times:

![image](https://github.com/user-attachments/assets/f3d4881f-bc17-4922-bd65-dac05bbd04bc)

yenaing-oo:
@danloveg 

I found that even though `DEBUG` is set to `True` in `docker_dev.py`, `DEBUG` itself never actually got passed  as context to templates:

![image](https://github.com/user-attachments/assets/0fe3843e-25dd-4bde-803f-a1e06824207c)

If you add the above without any other changes and open any page of the app on your browser, you will see that the `else` condition is fired and the value of `DEBUG` is empty.

It turns out that the `debug` context processor only passes the `debug` variable as context if the request made came from an IP that is listed in `settings.INTERNAL_IPS` (a setting that has not been set yet), in addition to `settings.DEBUG` being `True`:
![image](https://github.com/user-attachments/assets/d294dc94-6844-49c7-9348-9e24b23e08d4)
![image](https://github.com/user-attachments/assets/8502ab23-174c-4d81-9bc4-2ba99300de82)

I tried adding this setting to `docker_dev.py`:
```
INTERNAL_IPS = [
    '127.0.0.1',
    'localhost',
]
```

But this didn't work. Upon studying the value of `self.request.META["REMOTE_ADDR"]` a given request (say the About page), I discovered that the IP address is actually not `localhost` ('127.0.0.1'), but `192.168.65.1` instead.

It appears that Docker is intercepting any requests made by my local host and relaying it to the `app` service using its own default IP address `192.168.65.1`. 

I believe this IP address can be configured in Docker's settings here:
![image](https://github.com/user-attachments/assets/64000ec0-2bd0-4831-a421-0bf7ce433b4c)
, although the actual IP address on the request is slightly different (note the .1 at the end instead of .0).

Modifying this setting:
```
INTERNAL_IPS = [
     '192.168.65.1'
]
```

and replacing any `{% if DEBUG %}` with `{% if debug %}` fixes this issue.

danloveg:
That makes sense. My question is, it seems like django-pipeline will send un-minified assets when DEBUG=True automatically - what happens when you try to use this without checking whether debug is true in the template?

```
{% load pipeline %}
{% javascript 'recordtransfer_base_js' %}
```

[...]

PR #246, opened 2024-11-27
Fix minified assets being generated in development

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/245

### DIFF
--- a/bagitobjecttransfer/bagitobjecttransfer/settings/docker_dev.py
+++ b/bagitobjecttransfer/bagitobjecttransfer/settings/docker_dev.py
@@ -1,7 +1,7 @@
 # pylint: disable=wildcard-import
 # pylint: disable=unused-wildcard-import
-from pathlib import Path
 from decouple import config
+
 from .base import *
 
 DEBUG = True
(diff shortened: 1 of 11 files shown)


yenaing-oo:
Strangely, after these changes, the static assets are no longer being minified in production:
image
It looks like the pipeline is being used, since the asset names have the numbers at the back.

Development works perfectly fine though.

danloveg:
I wonder if explicitly setting `PIPELINE_ENABLED: USE_PIPELINE` would make a difference in the dictionary that configures django-pipeline:

https://django-pipeline.readthedocs.io/en/latest/configuration.html

yenaing-oo:
> I wonder if explicitly setting `PIPELINE_ENABLED: USE_PIPELINE` would make a difference in the dictionary that configures django-pipeline:
> 
> https://django-pipeline.readthedocs.io/en/latest/configuration.html

Unfortunately not. Downgrading to Django ^3.0.0 did not do the trick either.

danloveg:
Alright I think I'm going to call it here - we're spending way too much time wrestling with this django-pipeline tool. Let's work to migrate to some other JavaScript/TypeScript tool that's more suitable and customizable for doing the bundling/minification.

I'll open a new issue for migrating away from django-pipeline.

Issue #247, opened 2024-11-28
Migrate to Webpack (or similar tool)


danloveg:
Debugging django-pipeline is taking too much of our time and focus away from writing code. I would like us to migrate to a static asset bundler like [Webpack](https://webpack.js.org/).

If possible, I would prefer to use [Deno](https://github.com/denoland/deno) over NodeJS because it has Typescript support out of the box (it doesn't need to run a transpiler to run .ts files directly). All of the Webpack documentation online focuses on NodeJS though so if it's too much trouble to use Deno then we can just use node.

I see changes needing to be made in a few places:

- In the builder stage of the `Containerfile`, there should be a call to webpack to bundle the static assets to prepare them
- If using NodeJS, all the current dependencies should be uninstalled, and the new package bundler should be installed
- The special `{% include %}` template tags for including JavaScript should be removed
- The `{% load pipeline %}` tags should be removed from the templates too
- The way JavaScript and CSS are loaded in the `<head>` of the document will likely need to change to reflect the location of the new bundled assets
- The pipeline setup in `base.py` can be completely removed

PR #248, opened 2024-11-29
Migrate to webpack

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/247


yenaing-oo:
Note for self:
- Serve unbundled assets in development
- Hot reloading upon static asset change/ config change

compose.dev.yml
```diff
@@ -65,7 +65,7 @@ services:
       context: .
       dockerfile: Containerfile
     container_name: recordtransfer_django_app
-    command: python manage.py runserver 0.0.0.0:8000
+    command: /bin/sh -c "npm run build && python manage.py runserver 0.0.0.0:8000"
```
danloveg, line 68: Just taking a quick look at your changes so far, `npm run build` should be run from `entrypoint.sh` instead of here.

Containerfile
```diff
@@ -20,16 +7,23 @@ ENV PYTHONUNBUFFERED=1
...
 # Copy files to container
 COPY pyproject.toml README.md ./docker/entrypoint.sh /app/
 COPY ./bagitobjecttransfer /app/bagitobjecttransfer
 
-# Install Python dependencies
+# # Install Python dependencies
 RUN pip install .
 
+# Install Node.js
+RUN apt-get update && apt-get install -y curl && \
```
danloveg, line 18: Another comment which we can discuss - NodeJS was installed in the previous image, `FROM docker.io/node:20 AS builder`, so it shouldn't be required to install it in the Python image.

compose.dev.yml
danloveg: Or actually, I think I see why this is here. If you're developing the JavaScript you'd need to be able to re-run the webpack command

yenaing-oo: I think it should be fine if I move this to `entrypoint.sh` - I tried it and there were no issues. I'm not sure what you mean by being able to run the webpack command.

danloveg: The webpack command I'm referring to is what's being run during `npm run build` - i.e., `webpack --config webpack.config.js`

yenaing-oo:
Note to self: Check if it is standard practice to bundle external libraries such as jQuery loaded from a CDN into our own JS.

Containerfile
yenaing-oo: Note: try swapping to this image that includes both Python and Node
https://hub.docker.com/r/nikolaik/python-nodejs
python3.10-nodejs22-slim

compose.dev.yml
yenaing-oo: Ah I see what you mean. Moving `npm run build` to the entrypoint would not be an issue since npm scripts can be run from anywhere within a node project.

Containerfile
yenaing-oo: Swapped to the above mentioned image.

Resolved by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/248/commits/5347e12e5a114de6abddc1df1870926f8fc52de2

yenaing-oo:
@danloveg 

I tried completely removing`django-pipeline` and any settings that use it, but after doing so, requests for static assets were always responded with a 404 error.

If you have the chance, could you help me figure out what's causing this issue? The very last commit I made (https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/248/commits/fe4ed8056af90ffbdd9f35797c5db975e131eab6) is my attempt at removing `django-pipeline`. Everything before that works fine.

danloveg:
> @danloveg
> 
> I tried completely removing`django-pipeline` and any settings that use it, but after doing so, requests for static assets were always responded with a 404 error. This only happens in development and not the production container.
> 
> If you have the chance, could you help me figure out what's causing this issue? The very last commit I made ([fe4ed80](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/commit/fe4ed8056af90ffbdd9f35797c5db975e131eab6)) is my attempt at removing `django-pipeline`. Everything before that works fine.

I'll take a look tomorrow. I think it's because you removed `STATICFILES_FINDERS` and `STORAGES` but I can double check whether adding those back fixes the problem.

yenaing-oo:
> > @danloveg
> > I tried completely removing`django-pipeline` and any settings that use it, but after doing so, requests for static assets were always responded with a 404 error. This only happens in development and not the production container.
> > If you have the chance, could you help me figure out what's causing this issue? The very last commit I made ([fe4ed80](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/commit/fe4ed8056af90ffbdd9f35797c5db975e131eab6)) is my attempt at removing `django-pipeline`. Everything before that works fine.
> 
> I'll take a look tomorrow. I think it's because you removed `STATICFILES_FINDERS` and `STORAGES` but I can double check whether adding those back fixes the problem.

Ok, thank you. I don't think that's the issue because I tried keeping both of them and it did not make a difference. Also, according to[ Django 4.2's docs](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-STATICFILES_FINDERS), both settings have a default value. But feel free to double-check.

danloveg:
> > > @danloveg
> > > I tried completely removing`django-pipeline` and any settings that use it, but after doing so, requests for static assets were always responded with a 404 error. This only happens in development and not the production container.
> > > If you have the chance, could you help me figure out what's causing this issue? The very last commit I made ([fe4ed80](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/commit/fe4ed8056af90ffbdd9f35797c5db975e131eab6)) is my attempt at removing `django-pipeline`. Everything before that works fine.
> > 
> > 
> > I'll take a look tomorrow. I think it's because you removed `STATICFILES_FINDERS` and `STORAGES` but I can double check whether adding those back fixes the problem.
> 
> Ok, thank you. I don't think that's the issue because I tried keeping both of them and it did not make a difference. Also, according to[ Django 4.2's docs](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-STATICFILES_FINDERS), both settings have a default value. But feel free to double-check.

I fixed this issue on this branch in d1ea5e421d3aa0f6f5977851c8bb9c175600c6e7. This issue was happening because the bundled files weren't collected during collectstatic. So even though they were present in the static/ folder, Django didn't know about them, so it was saying they didn't exist.

What I did to fix this was build the Webpack assets to a `dist/` folder, and told Django that static files should be copied from there (using `STATICFILES_DIRS`). This also means that webpack has to run *before* collectstatic, because now collectstatic will only copy those bundled files after they've been created.

While this is working, I could possibly see this as being an issue if we want to do hot re-loading of static assets. Webpack can do hot-reloading, of course, but without running collectstatic after a hot-reload, the new files won't be picked up. Maybe there's a simple solution to that problem.

yenaing-oo:
> I fixed this issue on this branch in [d1ea5e4](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/commit/d1ea5e421d3aa0f6f5977851c8bb9c175600c6e7). This issue was happening because the bundled files weren't collected during collectstatic. So even though they were present in the static/ folder, Django didn't know about them, so it was saying they didn't exist.
> 
> What I did to fix this was build the Webpack assets to a `dist/` folder, and told Django that static files should be copied from there (using `STATICFILES_DIRS`). This also means that webpack has to run _before_ collectstatic, because now collectstatic will only copy those bundled files after they've been created.
> 
> While this is working, I could possibly see this as being an issue if we want to do hot re-loading of static assets. Webpack can do hot-reloading, of course, but without running collectstatic after a hot-reload, the new files won't be picked up. Maybe there's a simple solution to that problem.

https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/248/commits/d1ea5e421d3aa0f6f5977851c8bb9c175600c6e7: Thank you so much for looking into this! I had no idea Django needed all static files in `STATIC_ROOT` had to have been picked up by `collectstatic` for them to be recognized and served. Strangely, my method of adding the assets generated by webpack into `STATIC_ROOT` seemed to work before the pipeline was removed. 

https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/248/commits/bd779db822cb02a6f53355df92c7aac910c726ee: Thanks for doing this, I was planning on it myself after we agreed that CSS should also be bundled in development.

https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/248/commits/b4f62170d25e8224952aa238009f9bae03d38372: I understand how the installation of Python dependencies is made more efficient from usage of Poetry and its `poetry.lock` file. However, isn't the installation of node dependencies already efficient to begin with, since `package.lock.json` was already being copied over? Also, I'm not sure how the new change makes sure that `npm install` runs if there is a change in `webpack.config.js`, because I think only dependencies are tracked within `package.lock.json`.

danloveg:
> [d1ea5e4](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/commit/d1ea5e421d3aa0f6f5977851c8bb9c175600c6e7): Thank you so much for looking into this! I had no idea Django needed all static files in `STATIC_ROOT` had to have been picked up by `collectstatic` for them to be recognized and served. Strangely, my method of adding the assets generated by webpack into `STATIC_ROOT` seemed to work before the pipeline was removed.

I don't pretend to know what django-pipeline was doing, it's possible it was telling Django about those static resources somehow while it was still activated. Good news is we don't have to worry about django-pipeline anymore!

> [bd779db](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/commit/bd779db822cb02a6f53355df92c7aac910c726ee): Thanks for doing this, I was planning on it myself after we agreed that CSS should also be bundled in development, which meant the static asset file names used in templates was going to be the same, regardless or whether in development or production.

No problem 👍

> [b4f6217](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/commit/b4f62170d25e8224952aa238009f9bae03d38372): I understand how the installation of Python dependencies is made more efficient from usage of Poetry and its `poetry.lock` file. However, isn't the installation of node dependencies already efficient to begin with, since `package.lock.json` was already being copied over? Also, I'm not sure how the new change makes sure that `npm install` runs if there is a change in `webpack.config.js`, because I think only dependencies are tracked within `package.lock.json`.

Great question. The way Docker (or Podman) builds containers is based on layers. Each operation, like `RUN` or `COPY` makes a new *layer* in the image. If something changes in that layer, it invalidates the cache for that layer *and all future layers.*

If you look at the image in Docker, you'll see this:

layers

If Layer 20 changes, then all layers from 20 onwards have to be re-built. The package.json and webpack.config.js likely aren't going to change much, so we can cache the packages installed in layer 21 for the next time the image is built.

Similarly, the pyproject.toml file is not going to change often, so unless it (or poetry.lock, README.md) changes, then layer 23 will be cached and we don't have to re-install the packages.

Previously, this was causing all the packages to be re-installed:

```
COPY pyproject.toml README.md ./docker/entrypoint.sh /app/
COPY ./bagitobjecttransfer /app/bagitobjecttransfer

RUN pip install .
RUN npm install
```

If you consider the layers here, any time *anything* in `./bagitobjecttransfer` changes, the cache from that point gets invalidated, and all the dependencies have to be re-installed again, which does not actually need to happen.

yenaing-oo:
> Great question. The way Docker (or Podman) builds containers is based on layers. Each operation, like `RUN` or `COPY` makes a new _layer_ in the image. If something changes in that layer, it invalidates the cache for that layer _and all future layers._
> 
> If you look at the image in Docker, you'll see this:
> 
> If Layer 20 changes, then all layers from 20 onwards have to be re-built. The package.json and webpack.config.js likely aren't going to change much, so we can cache the packages installed in layer 21 for the next time the image is built.
> 
> Similarly, the pyproject.toml file is not going to change often, so unless it (or poetry.lock, README.md) changes, then layer 23 will be cached and we don't have to re-install the packages.
> 
> Previously, this was causing all the packages to be re-installed:
> 
> ```
> COPY pyproject.toml README.md ./docker/entrypoint.sh /app/
> COPY ./bagitobjecttransfer /app/bagitobjecttransfer
> 
> RUN pip install .
> RUN npm install
> ```
> 
> If you consider the layers here, any time _anything_ in `./bagitobjecttransfer` changes, the cache from that point gets invalidated, and all the dependencies have to be re-installed again, which does not actually need to happen.

This makes a lot of sense, thanks for the detailed explanation.

yenaing-oo:
Since we still have to rely on `collectstatic` for static assets to be recognized and served, would it be redundant to use webpack to process images as well?

I know that webpack compresses images and has the option to convert to WebP - a more optimized file format for browser. Although, right now, we don't have that many images for this app.

danloveg:
> Since we still have to rely on `collectstatic` for static assets to be recognized and served, would it be redundant to use webpack to process images as well?
> 
> I know that webpack compresses images and has the option to convert to WebP - a more optimized file format for browsers. Although, right now, we don't have that many images for this app.

I wouldn't say `collectstatic` is redundant even if we use webpack for images, since it would still need to collect those processed images.

How much work would it be to have it process images?

danloveg:
> > > Since we still have to rely on `collectstatic` for static assets to be recognized and served, would it be redundant to use webpack to process images as well?
> > > I know that webpack compresses images and has the option to convert to WebP - a more optimized file format for browsers. Although, right now, we don't have that many images for this app.
> > 
> > 
> > I wouldn't say `collectstatic` is redundant even if we use webpack for images, since it would still need to collect those processed images.
> > How much work would it be to have it process images?
> 
> Sorry, that wasn't very clear. I meant that using webpack for images may be redundant since we can still reference them within templates the way it's currently being done.
> 
> No, I don't think it's much work, looking through the docs. Thinking about it, I think it would be good to use webpack for images as well, for consistency.

Sounds good, go ahead and see about using webpack for images inthat case. If it doesn't seem to be adding much value then we can skip it but I agree, for consistency it would make sense to use webpack for all our static assets.

yenaing-oo:
Image format conversion to webp appears to work fine on production, but not does not seem to function in development. It is recommended that I only generate webp in production, but I need to be able to convert images to webp in development as well since images are being referenced by their filename in templates like this:

``

where the extension is assumed to always to be `.webp`

danloveg:
When you say it doesn't function, do you mean webp files aren't being created?

yenaing-oo:
> When you say it doesn't function, do you mean webp files aren't being created?

Yes, the original images are not being converted into webp.

danloveg:
Do you think this is worth debugging further, or should we skip the image stuff and look at it later?

yenaing-oo:
I'll try one last thing - changing the WEBPACK_MODE to production for the dev container, just to make sure it's because of the mode.

yenaing-oo:
Yeap, it's definitely the mode that's affecting it.

This is what happens when change `WEBPACK_MODE` to development in the `compose.dev.yml`:
image

The images do get converted into webp.

danloveg:
Alright, let's worry about images later, then. There's some undocumented behaviour here that which makes me not trust this plugin. The only way I could see this working right now is to use a different path in development vs. production, but that would require bringing back the MODE context processor which is not something I think we should do

yenaing-oo:
> Alright, let's worry about images later, then. There's some undocumented behaviour here which makes me not trust this plugin. The only way I could see this working right now is to use a different path in development vs. production, but that would require bringing back the MODE context processor which is not something I think we should do

In that case, is there anything else needed in this pull request before a full review? I think hot-reload would be better to work on in a separate issue.

danloveg:
> > Alright, let's worry about images later, then. There's some undocumented behaviour here which makes me not trust this plugin. The only way I could see this working right now is to use a different path in development vs. production, but that would require bringing back the MODE context processor which is not something I think we should do
> 
> In that case, is there anything else needed in this pull request before a full review? I think hot-reload would be better to work on in a separate issue. I can revert the image conversion to webp commits.

We're good for a review now. There are two things we can work on in separate issues:

- Include jQuery/Dropzone via webpack
- Minify images via webpack

yenaing-oo:
> > > Alright, let's worry about images later, then. There's some undocumented behaviour here which makes me not trust this plugin. The only way I could see this working right now is to use a different path in development vs. production, but that would require bringing back the MODE context processor which is not something I think we should do
> > 
> > 
> > In that case, is there anything else needed in this pull request before a full review? I think hot-reload would be better to work on in a separate issue. I can revert the image conversion to webp commits.
> 
> We're good for a review now. There are two things we can work on in separate issues:
> 
> * Include jQuery/Dropzone via webpack
> * Minify images via webpack

Sounds good. I'll go ahead and create those issues.

bagitobjecttransfer/recordtransfer/settings.py
```diff
@@ -96,3 +96,5 @@
     cast=AcceptedFileTypes(),
     default="Archive:zip|Audio:mp3,wav,flac|Document:docx,odt,pdf,txt,html|Image:jpg,jpeg,png,gif|Spreadsheet:xlsx,csv|Video:mkv,mp4",
 )
+
+MODE = config("MODE", default="DEFAULT_MODE")
```
danloveg, line 100: MODE can be removed - this is one place we missed it

bagitobjecttransfer/recordtransfer/context_processors.py
```diff
@@ -1,5 +1,5 @@
 from recordtransfer.settings import SIGN_UP_ENABLED, MAX_SINGLE_UPLOAD_SIZE, \
-    MAX_TOTAL_UPLOAD_COUNT, MAX_TOTAL_UPLOAD_SIZE, FILE_UPLOAD_ENABLED
+    MAX_TOTAL_UPLOAD_COUNT, MAX_TOTAL_UPLOAD_SIZE, FILE_UPLOAD_ENABLED, MODE
```
danloveg, line 2: MODE can be removed here

danloveg (changes requested):

bagitobjecttransfer/recordtransfer/settings.py
yenaing-oo: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/248/commits/ee359aacf3ca6f2c42350fbfcbe9de5eb549d3c0

bagitobjecttransfer/recordtransfer/context_processors.py
yenaing-oo: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/248/commits/ee359aacf3ca6f2c42350fbfcbe9de5eb549d3c0

yenaing-oo: Resolved by [Remove redundant MODE setting](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/248/commits/ee359aacf3ca6f2c42350fbfcbe9de5eb549d3c0)

danloveg (approved):

PR #254, opened 2024-12-06
Install Dropzone and Jquery as Node dependencies rather than loading from a CDN

yenaing-oo:


danloveg:
I am looking to completely remove jQuery in another issue (#256), so feel free to mark this as ready to review once the Dropzone change has been confirmed to work.

yenaing-oo:
Sounds good. Just one thing, is the white bar supposed to be a loading bar to show the status of an upload? Just wanted to check if this is intentional. I think it was the same before this change.

image

Also, I noticed that the text "Drop files here or click to upload" disappears after a file is removed. That was not the behaviour before, I'm working on fixing this.
image

yenaing-oo:
The dropzone message's visibility is now updated whenever a file is added or removed.

Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/254/commits/1a27d7d879fdcd9278de4d7e076fc912c032957d

package.json
```diff
@@ -8,6 +8,7 @@
   "dependencies": {
     "css-loader": "^7.1.2",
     "css-minimizer-webpack-plugin": "^7.0.0",
+    "dropzone": "^6.0.0-beta.2",
```
danloveg, line 11: Only one change - can you change the version of this package to [5.9.3](https://www.npmjs.com/package/dropzone/v/5.9.3)? I'm wary about using a beta version.

danloveg (changes requested):

package.json
yenaing-oo: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/254/commits/8e8335135f28f348ed7f94d0a632d960746d12d9

danloveg (approved):

Issue #252, opened 2024-12-06
Minify images using webpack


yenaing-oo:
It appears that generating images in  the `.webp` format (as a form of minification) from the orignal images using  Webpack's [ImageMinimizedWebpackPlugin](https://webpack.js.org/plugins/image-minimizer-webpack-plugin/#generate-webp-images-from-copied-assets) does not work when in development mode. This is an issue as the images have to be referenced by their exact filename and extension in Django templates.

An alternative would be to try loading images dynamically from JS ([see example](https://webpack.js.org/plugins/image-minimizer-webpack-plugin/#query-parameters-only-squoosh-and-sharp-currently)] and modify the DOM to add these images, which webpack will optimize/minify during its build of static assets.

If that does not work, try other Webpack plugins, and as a last resort, try using a django library instead.

danloveg:
This should be complete as #264 is merged in, right? @yenaing-oo

yenaing-oo:
> This should be complete as #264 is merged in, right? @yenaing-oo

Yes that's correct. We are using `sharp` in favour of `webpack` plugins, so we can close this issue.

PR #258, opened 2024-12-09
Load images using JS for webpack compression

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/252


yenaing-oo:
@danloveg 

How do you feel about this method of compressing images? Images would have to be loaded through JS and added as the `src` of pre-existing empty `` tags. The image file formats would not change, but there would be options to change the quality and level of compression for each image file format.

I think this method of compressing images using webpack is more suited for React JSX, which combine HTML-like syntax with JS. Here, I would have to make a separate JS file.

The only option for generating more optimized versions of images using webpack without the use of importing in JS/CSS that I can find is the previous method we explored ([Generating webp images from copied assets](https://webpack.js.org/plugins/image-minimizer-webpack-plugin/#generate-webp-images-from-copied-assets)).
But unfortunately, this only works in production mode. There is a way to force that plugin to work even in development by doing this:

```
optimization: {
        minimize: true,         # forces minification regardless of mode
        minimizer: [
            '...', // This keeps the default JavaScript minifier
            new CssMinimizerPlugin(), // Add this line to minify CSS
            new ImageMinimizerPlugin({
                      ...
```

But this would also cause the JS and CSS to also be minified in development.


A possible idea is to run a separate script using a npm package like `sharp` to compress the desired images ourselves.

danloveg:
> I made another [pull request](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/264) using a separate branch to demonstrate compression of images using `sharp` and a npm script.

I like the simplicity of the other method compared to using Webpack

danloveg:
Closing this pull request - #264 is now merged in to master

PR #260, opened 2024-12-09
Set up webpack --watch

danloveg:
Adds an `npm run watch` command which can be ran in the development container to re-build assets when they change.

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -62,6 +62,36 @@ Use this `launch.json` configuration in VSCode to debug the application:
 }
 ```
 
+## Javascript Development
+
+If you're editing any of the `.js` files in this repo, add these settings to your [local VSCode settings.json](https://code.visualstudio.com/docs/getstarted/settings#_settings-json-file) for this project to set your environment up.
+
+```json
+{
+    "[javascript]": {
+        "editor.formatOnSave": true,
+        "editor.defaultFormatter": "vscode.typescript-language-features",
+        "editor.rulers": [
+            99
+        ]
+    }
+}
+```
+
+### Re-build JS as Changes are Made
+
+After you run the dev container (`compose.dev.yml`), you can automatically re-build the JS files any time they change with this command:
+
+```shell
+# Using Docker:
+docker compose -f compose.dev.yml exec app npm run watch
+
+# Using Podman:
+podman-compose -f compose.dev.yml exec app npm run watch
+```
+
+This will re-build the bundled JS files any time you save a change to a `.js` file which is useful while you're writing Javascript. This command does not exit until you press CTRL-C, so make sure to run it in a separate terminal.
+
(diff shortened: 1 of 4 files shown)


yenaing-oo:
This is really handy - I don't have to call `collectstatic` whenever I modify a static asset anymore. I was hoping that it would also re-build assets whenever a new static asset was added to one of the directories in the entry points, but it looks like its only the original static files that are being tracked.

One thing I don't get is how the app manages to serve updated static assets without needing to call `collectstatic` after re-building. The bundled assets files are located and updated in the `/app/dist/` directory, but the corresponding bundled assets in `/app/bagitobjecttransfer/static` (from which Django serves static assets from) remain the same. However, refreshing my browser somehow fetches me the updated bundled asset.

danloveg:
> One thing I don't get is how the app manages to serve updated static assets without needing to call `collectstatic` after re-building. The bundled assets files are located and updated in the `/app/dist/` directory, but the corresponding bundled assets in `/app/bagitobjecttransfer/static` (from which Django serves static assets from) remain the same. However, refreshing my browser somehow fetches me the updated bundled asset.

This stumped me too, so I did a little bit more digging. It looks like the combination of `django.contrib.staticfiles.finders.FileSystemFinder` with `STATICFILES_DIRS` tells Django where to pick up static files. Since static files don't actually need to be collected with the development server, I believe it just accesses them directly from where they are (in this case `/app/dist/`).

If this is true, `collectstatic` probably does not need to be run in development ever.

yenaing-oo:
I see, this would make sense. 

However, I recall you having issues with an outdated JS file in the last issue I worked on (https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/243#discussion_r1860902600) due to not having called `collectstatic` in development, at which point you decided to add `collectstatic` to `entrypoint.sh`.

But I also recall making a change to a CSS file to modify the colour of the text in the footer and seeing the update take effect on my browser. This was in development and before webpack was being used.

In any case, the current configuration for rebuilding assets with `watch` is working fine, so I am not too concerned.

yenaing-oo (approved):

danloveg:
> However, I recall you having issues with an outdated JS file in the last issue I worked on ([#243 (comment)](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/243#discussion_r1860902600)) due to not having called `collectstatic` in development, at which point you decided to add `collectstatic` to `entrypoint.sh`.
> 
> But I also recall making a change to a CSS file to modify the colour of the text in the footer and seeing the update take effect on my browser. This was in development and before webpack was being used.

I believe this was because of the three-file problem, and my browser was using outdated code before running `collectstatic` with `django-pipeline`.

I'll open a new issue to investigate whether `collectstatic` is needed in development, for now I'll leave it in.

Thanks for reviewing.

PR #261, opened 2024-12-09
Make Entrypoint Script Executable

danloveg:
This change is needed when the image is built with Podman on WSL.


yenaing-oo (approved):
Being on a Mac, I'm not able to test this myself but this makes sense.

PR #262, opened 2024-12-09
Change re-directed page to transfersent

danloveg:
Sends the user to `recordtransfer:transfersent` instead of `recordtransfer:userprofile`.


yenaing-oo (approved):
Works as expected!

Issue #236, opened 2024-11-08
"Other source type" and "Other source role" missing asterisks


yenaing-oo:
In the Source Information Step (Step 3) of the transfer form, the form fields "Other source type" and "Other source role" are missing asterisks.

danloveg:
I'm currently working on #259 which is related, I can address this issue on the branch associated with that issue

yenaing-oo:
Addressed by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/263

PR #263, opened 2024-12-09
Simplify source information section

danloveg:
Adds a radio button to the Source Information part of the transfer form to conditionally show the Source Information fields.


danloveg:
@yenaing-oo some extra context:

- I added a new function `setupSelectOtherToggle2` which is related to removing jQuery. So right now I'm aware there are two ways to set up a "select other" toggle. I'll remove the old one in a future pull request.
- I was getting some errors in the dropzone form so I had to fix those to make the transfer form able to submit
- These changes also address #236

yenaing-oo (changes requested):

yenaing-oo:
image
- I think it would be better if "Name" was empty, Source Type and Role were set to "Please select one" by default if the user is submitting on behalf of someone else, since we are assuming the source is not the user.

danloveg:
> I think it would be better if Name of source was empty, Source Type and Role were set to "Please select one" by default if the user is submitting on behalf of someone else, since we are assuming the source is not the user.
> 
> I know that these values are initialized the way they are right now to account for the user selecting "No", so perhaps some JS can be used to clear these fields if pressing "Yes" for the first time?

That makes sense, I'll make that change.

bagitobjecttransfer/recordtransfer/static/recordtransfer/js/transferform/forms.js
```diff
@@ -530,6 +567,71 @@ $(() => {
...
+        const sourceRoleSelect = document.getElementById(sourceInfoContext["id_source_role"]);
+        const enterManualInfoInputId = sourceInfoContext["id_enter_manual_source_info"];
+        const enterManualInfoInputElement = document.getElementById(enterManualInfoInputId);
+
+        if (enterManualInfoInputElement) {
+            enterManualInfoInputElement.addEventListener("change", function (event) {
+                const selected = this.querySelector("input[type=radio]:checked").value;
+
+                if (selected === "yes") {
+                    document.querySelectorAll(".initially-hidden").forEach((el) => {
```
yenaing-oo, line 614: I see why you have two different classes related to hiding elements now: `initially-hidden` is used to mark an element as hideable, and `hidden-item` actually hides it.

bagitobjecttransfer/recordtransfer/static/recordtransfer/js/transferform/forms.js
```diff
@@ -530,6 +567,71 @@ $(() => {
...
+                        el.classList.remove("hidden-item");
+                    });
+
+                    // Dispatch events to update "other" fields if they were just shown by accident
+                    sourceTypeSelect.dispatchEvent(new Event("change"));
```
yenaing-oo, line 619: I also get why you need to trigger an event to update the hidden status of the other source role and other source type fields. I was initially confused by why there were two separate event listeners - one for when the user changes whether they are entering the source info manually (Yes/No) and one for whenever the select value of the dropdown for other source/type changes

yenaing-oo: I wonder which is better practice:

1) (current) Defining a function inline to call upon event trigger and then triggering event
```javascript
enterManualInfoInputElement.addEventListener("change", function (event) {
// contents
}

enterManualInfoInputElement.dispatchEvent(new Event("change")); # manually triggering event 
```
2) Defining function outside, and referencing it as the the function to call when defining an event listener:
```javascript
def updateHiddenFormFields() {}
enterManualInfoInputElement.addEventListener("change", updateHiddenFormFields)

updateHiddenFormField() // Manually calling function for the first time upon load
```

danloveg:
> I think it would be better if Name of source was empty, Source Type and Role were set to "Please select one" by default if the user is submitting on behalf of someone else, since we are assuming the source is not the user.

This is now resolved in 1d1f6c85bd5669ea6de1792da3bd46006e35aace.

bagitobjecttransfer/recordtransfer/static/recordtransfer/js/transferform/forms.js
danloveg: Unfortunately I'm using `this` inside the function, which gives me a direct reference to the element where the event originated from.

https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/f0e560c8143537e8a24a07e15051c73d125f9f3d/bagitobjecttransfer/recordtransfer/static/recordtransfer/js/transferform/forms.js#L611

If I call the function outside of an event listener I lose the ability to use `this`.

yenaing-oo (commented):
Everything appears to be working. The tests for `SourceInfoForm` cover all relevant cases and pass as expected. This should be good to go, aside from my one concern about `package-lock.json`.

bagitobjecttransfer/recordtransfer/static/recordtransfer/js/transferform/forms.js
yenaing-oo: I see, this makes sense in that case.

package-lock.json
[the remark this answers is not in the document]
danloveg: Yes, I ran `npm install` locally and the package-lock.json got updated. I think the last time we installed a package the package-lock.json forgot to be committed.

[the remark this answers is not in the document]
yenaing-oo: I see, that makes sense.

danloveg:
I think you need to approve this one again - GitHub isn't letting me merge this yet until you explicitly approve

yenaing-oo (approved):

yenaing-oo:
> I think you need to approve this one again - GitHub isn't letting me merge this yet until you explicitly approve

I've approved it again now 👍

PR #264, opened 2024-12-10
Compress images using sharp

yenaing-oo:
An alternative method to compressing images, that does not rely on webpack.

### DIFF
--- a/scripts/compress_images.js
+++ b/scripts/compress_images.js
@@ -1,0 +1,39 @@
+const sharp = require('sharp');
+const fs = require('fs');
+const path = require('path');
+
+const parentDir = path.dirname(__dirname);
+const inputDir = path.join(parentDir, 'bagitobjecttransfer/recordtransfer/static/recordtransfer/img');
+const outputDir = path.join(parentDir, 'dist');
+
+if (!fs.existsSync(outputDir)) {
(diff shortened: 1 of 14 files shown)


yenaing-oo:
Here is a comparison of the original image file sizes vs the compressed ones:

Original
image

Compressed
image

scripts/compress_images.js
danloveg, line 1: Nice, you wrote this in Javascript!

I do have a comment, can you add a console.log at the start and end so there's more feedback in the logs about when this script starts and ends?

danloveg (changes requested):
I like this, it's simple and works in development and production.

I also like if we accidentally forget to link to the `.webp` file that the original file will still load fine.

scripts/compress_images.js
yenaing-oo: Absolutely.

Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/264/commits/6086edc73c6cd7fb157737e36fddb86a3c82459e

danloveg (dismissed):
This all looks great.

This opens the door to migrate this code to a custom webpack plugin. For the time being, though, there's no need to do anything more complex than this.

danloveg (approved):

Issue #210, opened 2024-10-11
Migrate recordtransfer.settings to bagitobjecttransfer.settings.base


danloveg:
There is no reason for there to be a separate settings module outside of the `DJANGO_SETTINGS_MODULE`. All of the settings in recordtransfer.settings should be moved over to the base settings module.

This will likely affect many different files, so I am expecting the pull request to be quite large.

danloveg:
This issue is currently blocking #215.

PR #265, opened 2024-12-10
Migrate recordtransfer settings to base settings

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/210


yenaing-oo:
I removed ALLOW_BAG_CHANGES in this commit https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/265/commits/638e73d1b90f9d00accb44b0814c4f59ef7b9d6f because it was unused. Do we want to keep that for future use? If so, I can revert that change.

There were also plenty of default CAAIS form values that were unused, but I did move those to the base settings. Let me know if we want to keep those as well.

bagitobjecttransfer/recordtransfer/apps.py
```diff
@@ -1,150 +1,155 @@
...
 import re
 
+from bagit import CHECKSUM_ALGOS
 from django.apps import AppConfig
 from django.core.exceptions import ImproperlyConfigured
 
-from bagit import CHECKSUM_ALGOS
-
-from recordtransfer import settings
+from bagitobjecttransfer import settings
```
danloveg, line 10: This should be: `from django.conf import settings`

bagitobjecttransfer/bagitobjecttransfer/settings/__init__.py
danloveg, line 1: I don't think this file is necessary if you use `from django.conf import settings`

bagitobjecttransfer/bagitobjecttransfer/settings/base.py
```diff
@@ -1,149 +1,146 @@
 import os
 
-from decouple import config
+from decouple import Csv, config
+from recordtransfer.configuration import AcceptedFileTypes
```
danloveg, line 4: The configuration file should be moved to `bagitobjecttransfer/`. This file should not depend on `recordtransfer`

bagitobjecttransfer/recordtransfer/context_processors.py
```diff
@@ -1,15 +1,25 @@
...
-    MAX_TOTAL_UPLOAD_COUNT, MAX_TOTAL_UPLOAD_SIZE, FILE_UPLOAD_ENABLED
+from bagitobjecttransfer.settings.base import (
+    FILE_UPLOAD_ENABLED,
+    MAX_SINGLE_UPLOAD_SIZE,
+    MAX_TOTAL_UPLOAD_SIZE,
+    SIGN_UP_ENABLED,
+)
+from bagitobjecttransfer.settings.base import (
+    MAX_TOTAL_UPLOAD_COUNT,
+)
```
danloveg, line 9: Importing the settings like this makes it easier to test:

```suggestion
from django.conf import settings
```

bagitobjecttransfer/recordtransfer/context_processors.py
```diff
@@ -1,15 +1,25 @@
...
+
 
 def signup_status(request):
-    return {'SIGN_UP_ENABLED': SIGN_UP_ENABLED}
+    return {"SIGN_UP_ENABLED": SIGN_UP_ENABLED}
```
danloveg, line 13: This should be:

```suggestion
    return {"SIGN_UP_ENABLED": settings.SIGN_UP_ENABLED}
```

And the same for all the other settings in this file.

bagitobjecttransfer/recordtransfer/emails.py
```diff
@@ -109,7 +109,7 @@ def send_thank_you_for_your_transfer(form_data: dict, submission: Submission):
             subject="Thank You For Your Transfer",
             template_name="recordtransfer/email/transfer_success.html",
             context={
-                "archivist_email": settings.ARCHIVIST_EMAIL,
+                "archivist_email": bagitobjecttransfer.settings.base.ARCHIVIST_EMAIL,
```
danloveg, line 112: If you import settings from django.conf, you can just do this:

```suggestion
                "archivist_email": settings.ARCHIVIST_EMAIL,
```

danloveg (changes requested):
I like where you're going with this change, but I do have a change that will affect most of the files you've changed.

Instead of `from bagitobjecttransfer import settings`, I need you to switch to `from django.conf import settings`.

I recommend you take a quick look at this documentation here: https://docs.djangoproject.com/en/4.2/topics/settings/#using-settings-in-python-code

bagitobjecttransfer/bagitobjecttransfer/settings/__init__.py
yenaing-oo: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/265/commits/f2e977ea9048655dcf0d30610f5cad5ee2a3eb6b

bagitobjecttransfer/recordtransfer/context_processors.py
yenaing-oo: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/265/commits/c510965457dec23d9fcfbb7b45ea77955c52f5ba

bagitobjecttransfer/recordtransfer/context_processors.py
yenaing-oo: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/265/commits/c510965457dec23d9fcfbb7b45ea77955c52f5ba

yenaing-oo:
Thanks for the feedback, I didn't know DJANGO_SETTINGS_MODULE was an actual built-in Django setting, from which settings can be accessed by `django.conf`. I made the mentioned changes, in addition to these two:
- Standardize usage of `django.conf` everywhere: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/265/commits/81bd11c28ed28c4f4beb434cb890e419245c21a3
- Discovered a potential mistake in setting module naming in `asgi.py`: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/265/commits/81bd11c28ed28c4f4beb434cb890e419245c21a3

yenaing-oo:
Last batch of commits was a rebase on `master`

danloveg (changes requested):
Great to get that code migrated to the Django settings!

I did notice that a bunch of the tests started failing due to lines like this:

```python
@patch("recordtransfer.settings.MAX_TOTAL_UPLOAD_SIZE", 3)
```

Could you fix those failing tests?

I have a feeling that `bagitobjecttransfer.settings.test` will also need these settings defined:

- `BAG_STORAGE_FOLDER`
- `UPLOAD_STORAGE_FOLDER`

danloveg:
> I removed ALLOW_BAG_CHANGES in this commit [638e73d](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/commit/638e73d1b90f9d00accb44b0814c4f59ef7b9d6f) because it was unused. Do we want to keep that for future use? If so, I can revert that change.

I also forgot to reply to this - no we don't want to keep that, that's a setting for a feature that no longer exists. Thanks for taking it out.

yenaing-oo:
> Great to get that code migrated to the Django settings!
> 
> I did notice that a bunch of the tests started failing due to lines like this:
> 
> ```python
> @patch("recordtransfer.settings.MAX_TOTAL_UPLOAD_SIZE", 3)
> ```
> 
> Could you fix those failing tests?
> 
> I have a feeling that `bagitobjecttransfer.settings.test` will also need these settings defined:
> 
> * `BAG_STORAGE_FOLDER`
> * `UPLOAD_STORAGE_FOLDER`

Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/265/commits/038104a76773bdabd830f33df2090cc2b2c13585

yenaing-oo:
I noticed that the Sphinx documentation for `Application Settings` would also have to be updated. I'm less familiar with how this works and what changes would be appropriate. For instance, in `docs.settings.index.rst`:

```
Application Settings
====================

.. automodule:: recordtransfer.settings
   :members:

```

What should we change `.. automodule:: recordtransfer.settings` to?

There are also references to `recordtransfer.settings` in `bagitobjecttransfer/recordtransfer/caais.py` that I'm not sure how to handle.

danloveg:
> I noticed that the Sphinx documentation for `Application Settings` would also have to be updated. I'm less familiar with how this works and what changes would be appropriate.

I updated the Sphinx documentation. `automodule` tries to extract as much information about public functions from docstrings, but since there were no public functions, the only thing being pulled from `recordtransfer.settings` was the module docstring.

danloveg:
> There are also references to `recordtransfer.settings` in `bagitobjecttransfer/recordtransfer/caais.py` that I'm not sure how to handle. I did this ([679b100](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/commit/679b100781ad57bac57db75f8c49cba68f3f483b)) for now.

That was the right change to make, I have no issues with that.

danloveg (approved):
Tests are working, everything looks good. Anything else before I merge this in?

yenaing-oo:
> Tests are working, everything looks good. Anything else before I merge this in?

No, everything looks good to me!

Issue #268, opened 2024-12-11
Update DEVELOPERS.md to use poetry to install dependencies instead of pip


yenaing-oo:
When the change to use `poetry` for dependency management was made, the related documentation for developers was not updated. Instructions to install dependencies using poetry should should be included, rather than `pip` in `DEVELOPERS.md`.

PR #269, opened 2024-12-11
Update DEVELOPERS.md documentation to use poetry

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/268

- Included instructions on setting up virtual environment and installing poetry before installing relevant dependencies with defined extras

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -139,7 +139,10 @@ DJANGO_SETTINGS_MODULE=bagitobjecttransfer.settings.test python manage.py test
 
 The documentation for this repository is built with [Sphinx](https://sphinx-doc.org).
 
-Before building the documentation, ensure you have the `dev` dependencies installed.
+Before building the documentation, ensure you have the `docs` dependencies installed:
+```shell
+poetry install --extras "docs"
+```
 
 To build the docs, run:
 
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -94,11 +94,18 @@ This will re-build the bundled JS files any time you save a change to a `.js` fi
 
 ## Testing Setup
 
-Ensure that you've installed the `dev` dependencies locally (preferably in a virtual environment):
+[Poetry](https://python-poetry.org/) is used to manage dependencies for this project. Install Poetry on your system:
 
 ```shell
-pip install .[dev]
+pip install poetry
 ```
+
+Install the `dev` dependencies with Poetry:
+
+```shell
+poetry install --extras "dev"
+```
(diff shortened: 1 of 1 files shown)


yenaing-oo:
@danloveg 

Could you test that these instructions work as expected on a fresh work environment? I have multiple Python versions installed on my machine, so I just wanted to double-check that there aren't any issues.

DEVELOPERS.md
```diff
@@ -92,12 +92,31 @@ podman-compose -f compose.dev.yml exec app npm run watch
...
+.\env\Scripts\Activate.ps1
+pip install poetry
+```
+
+On MacOS/Linux:
+```shell
+python3 -m venv env
+source env/bin/activate
+pip install poetry
+```
```
danloveg, line 110: In the [Poetry installation documentation](https://python-poetry.org/docs/#installation), there's a big warning that poetry needs to be installed in its own environment, separate from the dependencies it manages. The reasoning is that `poetry install` might install an incompatible version of a package that it itself depends on.

The recommended method to install Poetry is with `pipx`.

I would recommend changing this documentation to:

---

# Virtual Environment and Poetry Setup

Python dependencies are managed with [Poetry](https://python-poetry.org/). To install Poetry and create a virtual environment with the minimally necessary packages, follow these instructions:

First, install `pipx` via the [official pipx installation instructions](https://pipx.pypa.io/stable/installation/).

After installing `pipx`, install [Poetry](https://python-poetry.org/):

```shell
pipx install poetry
```

Next, create an activate a virtual environment from the root of the repository:

```shell
# If on Windows:
python -m venv env
.\env\Scripts\Activate.ps1

# If on MacOS or Linux
python3 -m venv env
source env/bin/activate
```

Finally, install the dependencies with `poetry`:

```shell
poetry install
```

danloveg (changes requested):
I recommend installing poetry via `pipx` as a safer method of installing Poetry

DEVELOPERS.md
yenaing-oo: > In the [Poetry installation documentation](https://python-poetry.org/docs/#installation), there's a big warning that poetry needs to be installed in its own environment, separate from the dependencies it manages. The reasoning is that poetry install might install an incompatible version of a package that it itself depends on.

Ah I misunderstood it for "Install everything in a dedicated virtual environment". This makes a lot of sense.

Resolved by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/269/commits/efcda9b6e8aeaca65c9c51fd3bfdd16774362412

danloveg (approved):

PR #274, opened 2024-12-12
Fix Read the Docs Build Failing

danloveg:
Uses poetry instead of pip to install dependencies, since `package-mode = false` in `pyproject.toml` can't be handled with pip.

The working build can be seen here:

https://secure-record-transfer.readthedocs.io/en/273-fix-read-the-docs-build-failing/

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -13,7 +13,7 @@ build:
       - poetry config virtualenvs.create false
 
     post_install:
-      - source "$READTHEDOCS_VIRTUALENV_PATH/bin/activate" && poetry install --extras docs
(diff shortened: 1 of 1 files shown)


danloveg:
@yenaing-oo I know you don't have a way to test this but I'd appreciate you taking a super quick look to see if anything looks horribly wrong.

yenaing-oo:
> @yenaing-oo I know you don't have a way to test this but I'd appreciate you taking a super quick look to see if anything looks horribly wrong.

This looks good to me. I think that as long as the build shows no errors and the docs match the latest changes in the repository, we can trust that this works with no issues.

yenaing-oo (approved):

PR #277, opened 2024-12-13
Fix Header Styling

danloveg:
- Fixes the dead space under the header in the main application
- Fixes the header colours when in dark mode in the admin

### DIFF
--- a/bagitobjecttransfer/recordtransfer/templates/admin/base_site.html
+++ b/bagitobjecttransfer/recordtransfer/templates/admin/base_site.html
@@ -1,45 +1,97 @@
 {% extends "admin/base_site.html" %}
 {% load static %}
+
 {% block extrahead %}
+    {{ block.super }}
     <link rel="icon" href="{% static 'recordtransfer/img/favicon.ico'  %}" />
+{% endblock %}
+
+{% block extrastyle %}
+    {{ block.super }}
     <style>
-        #header {
-            background: #eee;
-            box-shadow: 0 4px 3px -3px #bbb;
+        html[data-theme="light"], :root {
+            --primary: #b13d32;
+            --secondary: #f0f0f0;
+            --breadcrumbs-bg: #b13d32;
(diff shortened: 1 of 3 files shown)


bagitobjecttransfer/recordtransfer/templates/admin/base_site.html
```diff
@@ -1,45 +1,97 @@
...
+            --header-color: #cecece;
+            --header-link-color: var(--accent);
+            --header-bg: #1a1a1a;
+            --default-button-bg: #3c83a3;
+            --button-bg: #325261;
+            --breadcrumbs-bg: #292929;
+
+            /* Override the section having color: var(--header-link-color) */
+            .section {
+                color: var(--primary-fg);
```
yenaing-oo, line 47: - Could you please explain where --primary-fg is being accessed from? I don't see it being defined anywhere. 
- As I understand, this part:
```
 --primary: #6b0900;
            --secondary: #5c5c5c;
            --accent:#ebebeb;
            
...
```
is defining css variables. But I don't see them being used anywhere. Are these configuring the way Django's base site looks? I can't seem to find documentation for these variables.

yenaing-oo (commented):
The styling for the navbar spacing makes sense to me, but I'm don't understand how the the different themes are being applied on the admin site, as I couldn't find documentation for this on Django.

bagitobjecttransfer/recordtransfer/templates/admin/base_site.html
yenaing-oo: Ah never mind, I found it: 
https://docs.djangoproject.com/en/5.1/ref/contrib/admin/
https://github.com/django/django/blob/main/django/contrib/admin/static/admin/css/base.css

And the CSS variables are being applied on the parent `html` element, which the elements that Django renders have access to.

yenaing-oo (approved):

PR #278, opened 2024-12-13
Add GitHub Workflow to Build and Push Image to ghcr.io

danloveg:
The purpose of this pull request is to build a new container image automatically whenever a release is made.

I previously tried getting this to work but it was complaining about the `Containerfile` not existing. That was before we moved the `Containerfile` to the root of the repository.


danloveg:
Built packages get created here:

https://github.com/NationalCentreTruthReconciliation?tab=packages&repo_name=Secure-Record-Transfer

yenaing-oo (commented):
Would it be a good idea to:
- add a check to see if a Containerfile exists right after checking out the repository?
- limit this workflow to act on the `master` branch only? I believe this workflow can be manually triggered on any branch (perhaps by accident)

danloveg:
> Would it be a good idea to:
> 
>  - add a check to see if a Containerfile exists right after checking out the repository?
>  - limit this workflow to act on the `master` branch only? I believe that right now, this workflow can be manually triggered on any branch (perhaps by accident)

I agree with both of these things. I addressed this feedback in [34397e1](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/278/commits/34397e1f629b9ccc4664449b3dcb0e4532afd306)

yenaing-oo:
I wonder if it's possible to test this workflow (without actually pushing the container image) on a branch without the Containerfile to see if it does exit early. Perhaps this is unnecessary if you trust the file checking works.

danloveg:
> I wonder if it's possible to test this workflow (without actually pushing the container image) on a branch without the Containerfile to see if it does exit early. Perhaps this is unnecessary if you trust the file checking works.

I will try that this afternoon. I'll have to temporarily add back the manual trigger to try it out

danloveg:
@yenaing-oo there's no way to test manually triggering this job at the moment. From [This page of the GitHub docs](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow) says:

> To trigger the workflow_dispatch event, your workflow must be in the default branch.

So we would need to merge it in before testing it out. Let's merge this in, and if it fails to run next time it runs I'll open up an issue to fix it.

yenaing-oo:
> @yenaing-oo there's no way to test manually triggering this job at the moment. From [This page of the GitHub docs](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow) says:
> 
> > To trigger the workflow_dispatch event, your workflow must be in the default branch.
> 
> So we would need to merge it in before testing it out. Let's merge this in, and if it fails to run next time it runs I'll open up an issue to fix it.

I see, sounds good!

yenaing-oo (approved):

Issue #253, opened 2024-12-06
Submission are not Paginated on Profile Page


danloveg:
Submissions are not being paginated on the profile page. These three tables should be paginated:

- In-progress Submissions
- Submission Groups
- Submissions

Pagination is pretty easy with Django: https://docs.djangoproject.com/en/4.2/topics/pagination/#using-paginator-in-a-view-function

PR #280, opened 2024-12-13
Paginate submission lists

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/253


yenaing-oo:
TODO:
- Pass name of query parameter to use for pagination as context
- Use query parameter name in template as a `with` option to pass to `pagination` include to control `href` value
- Make sure that `href` reincludes existing query parameters

yenaing-oo:
`paginate_by` is currently set to 2 per page for ease of testing. This can easily be changed to the desired value after testing.

bagitobjecttransfer/recordtransfer/templates/includes/pagination.html
```diff
@@ -0,0 +1,17 @@
+{% load i18n %}
+{% load query_params %}
+<div class="pagination">
+    {% if page_obj.has_previous %}
+        <a href="?{% update_query_params pagination_name page_obj.previous_page_number %}"
+           class="btn-arrow btn-arrow-left">{% trans "Previous" %}</a>
+    {% else %}
+        <a class="btn-arrow btn-arrow-left disabled" aria-disabled="true">{% trans "Previous" %}</a>
+    {% endif %}
+    <span>{% trans "Page" %} {{ page_obj.number }} {% trans "of" %} {{ page_obj.paginator.num_pages }}</span>
```
danloveg, line 10: A translated version of this page may not have the page number in the same form "page 1 of 3". In another language, it might be more natural to write "1 of 3 pages", for example.

In these cases when you have to use variables inside a translatable block, you can use `blocktrans` with `with`.

```suggestion
    <span>
        {% blocktrans with curr_page=page_obj.number total_pages=page_obj.paginator.num_pages %}
        Page {{ curr_page }} of {{ total_pages }}
        {% endblocktrans %}
    </span>
```

bagitobjecttransfer/recordtransfer/views.py
```diff
@@ -126,7 +130,7 @@ class UserProfile(UpdateView):
     """
 
     template_name = "recordtransfer/profile.html"
-    paginate_by = 10
+    paginate_by = 2
```
danloveg, line 133: Go ahead and put this back up to 10, looks like it's working well.

danloveg (changes requested):
The pagination functions just like I expected it to, with three different GET parameters for each of the tables. I especially like how the `pagination` was abstracted out into a reusable template.

I just had one small comment about improving the "translate-ability" of the pagination template.

bagitobjecttransfer/recordtransfer/templates/includes/pagination.html
yenaing-oo: That's an interesting case I didn't consider, thanks for pointing it out!

Resolved by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/280/commits/35e11ad778671b75c2f894b1639882caea2d0def

bagitobjecttransfer/recordtransfer/views.py
yenaing-oo: Resolved by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/280/commits/968f85ca9b89d347b94ae80188356f8875c33af6

danloveg (approved):

Issue #251, opened 2024-12-06
Enable name field on profile form


danloveg:
A user should be able to update their name on the profile page. The field is currently greyed out - the field should be enabled so that the user can change their name.

yenaing-oo:
@danloveg 

When a user first signs up, they had to enter their first and last name:
![image](https://github.com/user-attachments/assets/1bdb21e5-e451-4d14-bce3-9c35a6437576)

Currently, the "Name" that is being displayed on the Profile page is a user's full name, constructed from their first and last name:

```html
    <div class="flex-item margin-top-25px">
        <label>{% trans "Name" %}</label>
        <input disabled value="{{ user.get_full_name }}">
    </div>
```


```python
class User(AbstractUser):
    """The main User object used to authenticate users"""

    gets_submission_email_updates = models.BooleanField(default=False)
    confirmed_email = models.BooleanField(default=False)
    gets_notification_emails = models.BooleanField(default=True)

    def get_full_name(self):
        return self.first_name + " " + self.last_name
```

Do we want to add fields in the profile page to modify the first and last name separately? This makes the most sense to me, but I just wanted to check with you.

danloveg:
> Do we want to add fields in the profile page to modify the first and last name separately? This makes the most sense to me, but I just wanted to check with you.

Good point. Yes, that field should be split into two, one for the first name and one for the last name. Both should be required, since they're both required on the sign up form.

PR #281, opened 2024-12-16
Allow user to update their first and last name

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/251

Additional changes:
- Added a `reset` method to `UserProfileForm` to reset the form to original values after an invalid form submission
- Added tests for first and last name form fields
- Removed unnecessary form data for test cases, such as `get_email_notification` field when testing password changes or password fields when testing email notification changes.
- Fixed issue of messages not being dismissed after close button is clicked

### DIFF
--- a/bagitobjecttransfer/recordtransfer/forms/user_forms.py
+++ b/bagitobjecttransfer/recordtransfer/forms/user_forms.py
@@ -57,7 +57,7 @@ class UserProfileForm(forms.ModelForm):
         """Meta class for UserProfileForm."""
 
         model = User
-        fields = ("gets_notification_emails",)
+        fields = ("gets_notification_emails", "first_name", "last_name")
 
     def __init__(self, *args, **kwargs):
         user:User = kwargs.pop("user", None)
--- a/bagitobjecttransfer/recordtransfer/forms/user_forms.py
+++ b/bagitobjecttransfer/recordtransfer/forms/user_forms.py
@@ -8,15 +8,27 @@ from django.utils.translation import gettext_lazy as _
 from recordtransfer.constants import (
     ID_CONFIRM_NEW_PASSWORD,
     ID_CURRENT_PASSWORD,
+    ID_FIRST_NAME,
     ID_GETS_NOTIFICATION_EMAILS,
+    ID_LAST_NAME,
     ID_NEW_PASSWORD,
 )
 from recordtransfer.models import User
 
+import re
 
 class UserProfileForm(forms.ModelForm):
     """Form for editing user profile."""
-
+    first_name = forms.CharField(
+        widget=forms.TextInput(attrs={"id": ID_FIRST_NAME}),
+        label=_("First Name"),
+        required=True,
+    )
+    last_name = forms.CharField(
+        widget=forms.TextInput(attrs={"id": ID_LAST_NAME}),
+        label=_("Last Name"),
+        required=True,
+    )
     gets_notification_emails = forms.BooleanField(
         widget=forms.CheckboxInput(attrs={"id": ID_GETS_NOTIFICATION_EMAILS}),
         label=_("Receive notification emails?"),
@@ -39,6 +51,8 @@ class UserProfileForm(forms.ModelForm):
         required=False,
     )
 
+    NAME_PATTERN = re.compile(r"^(?!.*[.\s']{2})[A-Za-z\s.'-]+$")
+
     class Meta:
         """Meta class for UserProfileForm."""
 
@@ -46,9 +60,11 @@ class UserProfileForm(forms.ModelForm):
         fields = ("gets_notification_emails",)
 
     def __init__(self, *args, **kwargs):
-        user = kwargs.pop("user", None)
+        user:User = kwargs.pop("user", None)
         super().__init__(*args, **kwargs)
         if user:
+            self.fields["first_name"].initial = user.first_name
+            self.fields["last_name"].initial = user.last_name
             self.fields["gets_notification_emails"].initial = user.gets_notification_emails
 
     def clean(self) -> "dict[str, Any]":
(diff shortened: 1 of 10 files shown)


bagitobjecttransfer/recordtransfer/views.py
```diff
@@ -169,11 +171,15 @@ def get_context_data(self, **kwargs) -> dict[str, Any]:
         context["groups_page_obj"] = groups_paginator.get_page(groups_page_number)
 
         context.update(
-            {
+            {   
+                # Form field element IDs
+                "ID_FIRST_NAME":  ID_FIRST_NAME,
```
yenaing-oo, line 176: @danloveg 
I applied the same method of passing constants as context in this pull request. I know you proposed a cleaner way of passing context from template to JS by converting the python dictionary object into a JSON:

```
{{ js_context|json_script:"py_context_source_info" }}
```
This method does allow us to avoid defining each individual constant we want pass to the JS in the template. But I wonder if this is the best approach since we would have to dereference values using raw strings in the JS after parsing the JSON:

```javascript
// Context passed from template to JS
const sourceInfoContextElement = document.getElementById("py_context_source_info");
sourceInfoContext["id_source_type"] // 
```

yenaing-oo: > We still end up having to use VARIABLE_NAME. If you change VARIABLE_NAME in the view, you have to update the Javascript to reflect the new name in both of those cases. The fact that one is in a string doesn't make a difference regarding maintainability.

I see, both a change in the variable name in the view (for the method of passing individual constants) and a change in the string key in the view would require a change in the JS in both cases.

In that case, I'll update the way I'm currently passing context to use your suggested method.

danloveg: Great, I will review this pull request after you make that change

yenaing-oo: I've updated the way I'm passing context variables to be consistent with the new method in https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/281/commits/3de7ffd5e7c2d9d8791403762de1f02a8e4a08a5

bagitobjecttransfer/recordtransfer/forms/user_forms.py
```diff
@@ -39,17 +53,19 @@ class UserProfileForm(forms.ModelForm):
         required=False,
     )
 
+    NAME_PATTERN = re.compile(r"^(?!.*[.\s']{2})[A-Za-z\s.'-]+$")
+    NAME_VALIDATION_MESSAGE = _(
+        "Name must contain only letters and may include these symbols: ' - . \n"
+    )
```
danloveg, line 59: This validation is too strict - a person may have accented characters in their name (e.g., é) and this would fail validation

bagitobjecttransfer/recordtransfer/forms/user_forms.py
```diff
@@ -8,15 +8,29 @@
...
+    first_name = forms.CharField(
+        widget=forms.TextInput(attrs={"id": ID_FIRST_NAME}),
+        label=_("First Name"),
+        required=True,
+    )
+    last_name = forms.CharField(
+        widget=forms.TextInput(attrs={"id": ID_LAST_NAME}),
+        label=_("Last Name"),
+        required=True,
+    )
```
danloveg, line 33: Django has a built-in field type for regex validation:

```suggestion
    first_name = forms.RegexField(
        regex=UserProfileForm.NAME_PATTERN,
        error_messages={
            "required": _("First name is required."),
            "invalid": UserProfileForm.NAME_VALIDATION_MESSAGE,
        },
        widget=forms.TextInput(attrs={"id": ID_FIRST_NAME}),
        label=_("First Name"),
        required=True,
    )
    last_name = forms.RegexField(
        regex=UserProfileForm.NAME_PATTERN,
        error_messages={
            "required": _("Last name is required."),
            "invalid": UserProfileForm.NAME_VALIDATION_MESSAGE,
        },
        widget=forms.TextInput(attrs={"id": ID_LAST_NAME}),
        label=_("Last Name"),
        required=True,
    )
```

bagitobjecttransfer/recordtransfer/forms/user_forms.py
```diff
@@ -60,7 +76,15 @@ def clean(self) -> "dict[str, Any]":
...
-        gets_notification_emails = cleaned_data.get("gets_notification_emails")
+
+        first_name = cleaned_data.get("first_name")
+        last_name = cleaned_data.get("last_name")
+
+        if first_name and not self.NAME_PATTERN.match(first_name):
+            raise ValidationError({"first_name": [self.NAME_VALIDATION_MESSAGE]})
+
+        if last_name and not self.NAME_PATTERN.match(last_name):
+            raise ValidationError({"last_name": [self.NAME_VALIDATION_MESSAGE]})
```
danloveg, line 87: This is unnecessary with `RegexField`

```suggestion

        first_name = cleaned_data.get("first_name")
        last_name = cleaned_data.get("last_name")

        if first_name and not self.NAME_PATTERN.match(first_name):
            raise ValidationError({"first_name": [self.NAME_VALIDATION_MESSAGE]})

        if last_name and not self.NAME_PATTERN.match(last_name):
            raise ValidationError({"last_name": [self.NAME_VALIDATION_MESSAGE]})
```

danloveg (changes requested):

bagitobjecttransfer/recordtransfer/forms/user_forms.py
yenaing-oo: I've made the validation less strict by allowing these characters as well:
- `\u00C0-\u00D6`:  À to Ö.
- `\u00D8-\u00F6`:  Ø to ö
- `\u00F8-\u01FF`: ø to ÿ
For reference on Unicode characters: https://en.wikipedia.org/wiki/Latin-1_Supplement

Resolved by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/281/commits/70fde46c390930b972d9f5174fa95b75d94f5ee7

bagitobjecttransfer/recordtransfer/forms/user_forms.py
yenaing-oo: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/281/commits/70fde46c390930b972d9f5174fa95b75d94f5ee7

bagitobjecttransfer/recordtransfer/forms/user_forms.py
yenaing-oo: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/281/commits/70fde46c390930b972d9f5174fa95b75d94f5ee7

danloveg (approved):

Issue #282, opened 2024-12-16
Add ESLint Configuration for Javascript Linting/Formatting


danloveg:
Similar to how we're using `ruff.toml` for Python linting/formatting, we should add a linter for Javascript.

[ESLint](https://eslint.org/) should be used for this purpose. There is a VSCode plugin for this tool [here](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint).

PR #287, opened 2024-12-19
Add ESLint configuration

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/282

Notes:
- In addition to the [recommended rules](https://eslint.org/docs/latest/rules) (rules with a green tick beside them) defined by ESLint, I have also included some additional rules which can be found in the bottom section of `eslint.config.mjs`.
- ESLint deprecated all their formatting rules last year, but they recommend using the [ESLint Stylistic Plugin](https://eslint.style/), which I'm using
- I'm also using `eslint-plugin-import` in place of the [`sort-imports` rule](https://eslint.org/docs/latest/rules/sort-imports) since it's not able to fix all import sorting problems that it highlights
- Another plugin that's used is `eslint-plugin-jsdoc` for enforcing docstrings for JS functions
- I also updated the `DEVELOPERS.md` docs to include the instructions to use ESLint on VSCode

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -30,6 +30,12 @@ Additionally, install the [ESLint VSCode extension](https://marketplace.visualst
 }
 ```
 
+Finally, install Node dependencies for development:
+
+```shell
+npm install --include=dev
+```
+
 ## Debugging the Application
 
 The ports 8009 and 8010 are exposed to allow you to debug the web application, and the asynchronous job queue, respectively. This functionality is only available when `DEBUG = True`. These ports can be attached to with `debugpy`
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -14,6 +14,19 @@ Before beginning development, please configure git to use the git hooks (`.git/h
 git config core.hooksPath .githooks
 ```
 
+Additionaly, add this configuration to your [local VSCode settings.json](https://code.visualstudio.com/docs/getstarted/settings#_settings-json-file) for JavaScript linting/formatting:
+
+```json
+{
+    "editor.codeActionsOnSave": {
+        "source.fixAll.eslint": "explicit",
+    },
+    "eslint.validate": [
+        "javascript"
+    ]
+}
+```
+
 ## Debugging the Application
 
 The ports 8009 and 8010 are exposed to allow you to debug the web application, and the asynchronous job queue, respectively. This functionality is only available when `DEBUG = True`. These ports can be attached to with `debugpy`
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -14,7 +14,21 @@ Before beginning development, please configure git to use the git hooks (`.git/h
 git config core.hooksPath .githooks
 ```
 
-Additionally, install the [ESLint VSCode extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) for JavaScript linting.
+Additionally, install the [ESLint VSCode extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) and add this configuration to your [local VSCode settings.json](https://code.visualstudio.com/docs/getstarted/settings#_settings-json-file) for JavaScript linting:
+
+```json
+{
+    "[javascript]": {
+        "editor.formatOnSave": false,
+        "editor.codeActionsOnSave": {
+            "source.fixAll.eslint": "explicit"
+        },
+        "editor.rulers": [
+            99
+        ]
+    }
+}
+```
 
 ## Debugging the Application
 
@@ -68,21 +82,6 @@ Use this `launch.json` configuration in VSCode to debug the application:
 
 A GitHub Actions workflow is set up to run Django tests on every pull request to the master branch. All tests must pass before a merge is allowed. The workflow configuration can be found in `.github/workflows/django-tests.yml`.
 
-## Javascript Development
-
-If you're editing any of the `.js` files in this repo, add these settings to your [local VSCode settings.json](https://code.visualstudio.com/docs/getstarted/settings#_settings-json-file) for this project to set your environment up.
-
-```json
-{
-    "[javascript]": {
-        "editor.formatOnSave": true,
-        "editor.defaultFormatter": "vscode.typescript-language-features",
-        "editor.rulers": [
-            99
-        ]
-    }
-}
-```
 
 ### Re-build JS as Changes are Made
 
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -8,24 +8,13 @@ Follow [the NCTR's Python Style Guide](https://github.com/NationalCentreTruthRec
 
 ## Linting + Formatting
 
-Before beginning development, please configure git to use the git hooks (`.git/hooks`) provided in the repository. This will ensure that the linting configuration is updated before each commit.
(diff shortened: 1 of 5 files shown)


yenaing-oo:
Sorry for adding on more changes after requesting a review - this PR is ready for a review now.

DEVELOPERS.md
```diff
@@ -8,12 +8,28 @@ Follow [the NCTR's Python Style Guide](https://github.com/NationalCentreTruthRec
...
+        "editor.codeActionsOnSave": {
+            "source.fixAll.eslint": "explicit"
+        },
+        "editor.rulers": [
+            99
+        ]
+    }
+}
+```
+
```
danloveg, line 32: Can you add a note to run `npm install --include=dev` here?

danloveg (changes requested):
This is working great for ES6 .js files. It complains a lot about the `webpack.config.js` file, though. Is there a way to tell ESLint which files are CommonJS or does it work in only one mode at a time?

For example, it complains about these globals not existing in the webpack config:

- `module`
- `__dirname`
- `process`

These are Node-specific globals that aren't available in the browser.

If it's not possible to have ESLint work in both modes then I'm fine with how it's configured now.

yenaing-oo:
> Is there a way to tell ESLint which files are CommonJS or does it work in only one mode at a time?

This is indeed possible! I can override the language options and linting rules for specific files, which I've done for the webpack config file here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/287/commits/82329cf79b1a6cd4eaa3d5fdb846b1f140aa3576

DEVELOPERS.md
yenaing-oo: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/287/commits/184b98bcafdc2623cced22cdb22c701837979ce9

yenaing-oo:
I've also added a `lint` script to conveniently view linting errors in the terminal: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/287/commits/965c8ff68565fd34f6d23e313b6b2ea01135a685

package.json
```diff
@@ -4,17 +4,26 @@
   "description": "The secure record transfer application has been designed for the purpose of transferring files and standardized metadata over the internet to an institution.",
   "scripts": {
     "build": "webpack --config webpack.config.js && node scripts/compress_images.js",
-    "watch": "webpack --watch --config webpack.config.js"
+    "watch": "webpack --watch --config webpack.config.js",
+    "lint": "eslint '**/*.{js,cjs,mjs}'"
```
danloveg, line 8: Running `npm run lint` on Windows fails with this error:

```
> bagitobjecttransfer@2024.10.22 lint
> eslint '**/*.{js,cjs,mjs}'


Oops! Something went wrong! :(

ESLint: 9.17.0

No files matching the pattern "'**/*.{js,cjs,mjs}'" were found.
Please check for typing mistakes in the pattern.
```

I also realized that if you're targeting everything matching that pattern, you're going to lint these files which is not the desired behaviour:

- `node_modules/`
- `env/` (JS files in Python virtual environment)
- `dist/` (Built static resources)
- `bagitobjecttransfer/static/` (Collected static resources)

This suggestion works in tandem with another change I'm requesting in this review:

```suggestion
    "lint": "eslint"
```

eslint.config.mjs
```diff
@@ -0,0 +1,44 @@
+import pluginJs from "@eslint/js";
+import stylisticJs from "@stylistic/eslint-plugin-js";
+import importPlugin from "eslint-plugin-import";
+import jsdoc from "eslint-plugin-jsdoc";
+import globals from "globals";
+
+
+/** @type {import('eslint').Linter.Config[]} */
+export default [
+    {
```
danloveg, line 10: I suggest adding these paths to be ignored by ESLint:

```suggestion
export default [
    {
        ignores: [
            "env/",
            "dist/",
            "node_modules/",
            "bagitobjecttransfer/static/",
        ],
    },
    {
```

I had trouble getting this to work, but it's working the way I have it writtne. I thought `ignores` would go inside each configuration, but it's supposed to go inside it's own configuration like I have it above. See:

- https://eslint.org/docs/latest/use/configure/ignore#ignoring-directories
- https://github.com/eslint/eslint/discussions/18304#discussioncomment-9069706

danloveg (changes requested):
Just had one issue with the linting script - this should be resolved with my suggested changes but I'd suggest you verify on your end whether it works as expected.

package.json
yenaing-oo: > Running `npm run lint` on Windows fails with this error:
> 
> ```
> > bagitobjecttransfer@2024.10.22 lint
> > eslint '**/*.{js,cjs,mjs}'
> 
> 
> Oops! Something went wrong! :(
> 
> ESLint: 9.17.0
> 
> No files matching the pattern "'**/*.{js,cjs,mjs}'" were found.
> Please check for typing mistakes in the pattern.
> ```

I think this could be because single quotes are not recognized as a way to wrap strings in Windows. It I change them to escaped double quotes as shown below, it should work:

```
"lint": "eslint \"**/*.{js,cjs,mjs}\""
```
I agree with your suggested change because it looks like `eslint` already lints files with the above 3 extensions by default, so specifying them explicitly is redundant:
https://eslint.org/docs/latest/use/configure/configuration-files#specifying-files-without-extension

yenaing-oo: Including these directories in `ignores` makes sense to me:
> * `node_modules/`
> * `env/` (JS files in Python virtual environment)

However, I think these two would only be relevant to ignore if we want to be able run `eslint` in-container. Do you think this would be useful to have?
> * `dist/` (Built static resources)
> * `bagitobjecttransfer/static/` (Collected static resources)

If so, we'll have to include `eslint.config.mjs` as well when copying over files to the container in `Containerfile`, and list `eslint` dependencies under `dependencies` instead of `devDependencies` in `package.json`.

danloveg: > However, I think these two would only be relevant to ignore if we want to be able run `eslint` in-container. Do you think this would be useful to have?
> 
> > * `dist/` (Built static resources)
> > * `bagitobjecttransfer/static/` (Collected static resources)
> 
> If so, we'll have to include `eslint.config.mjs` as well when copying over files to the container in `Containerfile`, and list `eslint` dependencies under `dependencies` instead of `devDependencies` in `package.json`.

I ran into issues with those two directories *outside* of a container. Consider these two situations that would necessitate those two directories being ignored:

1. Someone runs `npm run build` locally. I've personally done this to test out the webpack config, so it's conceivable that someone else might as well.
  - This would require `dist/` being in the ignore list
2. Someone runs `python manage.py collectstatic` locally. This is also something I've done when testing out the webpack config.
  - This would require `bagitobjecttransfer/static/` being in the ignore list

I think those directories are a safe bet to include in the case someone might possibly run ESLint after executing either of those two commands locally.

yenaing-oo: Ah I see, I didn't think of that - thanks for clarifying!

yenaing-oo:
> Just had one issue with the linting script - this should be resolved with my suggested changes but I'd suggest you verify on your end whether it works as expected.

I've tested `npm run lint` with your suggested changes and made sure all the ignored directories were being ignored as expected.

I also included a few more common directory names for the Python virtual environment: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/287/commits/82ece7067a28ae7fb21c7f21a8be992f38be9374

danloveg (approved):
Works great now. Good catch about the different possible names for virtual environments, too.

Issue #237, opened 2024-11-08
Unable to return to previous step in final step of the transfer form


yenaing-oo:
![image](https://github.com/user-attachments/assets/4f495f88-e933-4f45-88b2-8fd09f0c57ae)

danloveg:
The session token is set when all the files have finished being uploaded here:

https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/6d9e5f148f0ced7ffa57e526e2e646e786c0e7c6/bagitobjecttransfer/recordtransfer/static/recordtransfer/js/dropzone/dropzoneForm.js#L279

yenaing-oo:
I think making the `session_token` field optional, and checking that `session_token` exists and is valid in the `done()` method of `TransferFormWizard` upon submission could be a potential way to fix this. 

This would allow the user to return to a previous step without the current step (Upload Files) failing validation.

danloveg:
That sounds like a valid approach. You should also check whether that field is set when the request reaches the backend, too

yenaing-oo:
I ran into this error after setting the `session_token` field's `required` argument to `False` and attempting to go to the Previous Step from the final step in the form:
![image](https://github.com/user-attachments/assets/f9098b89-68af-405c-9973-996ffea822e8)


Upon debugging the source code, it looks like the `WizardView` is not taking me to the `grouptransfer` step because `wizard_goto_step` does not exist in `self.request.GET`: 
![image](https://github.com/user-attachments/assets/58a21d6b-faf5-41e7-b8e8-679be26fad28)

However, `wizard_goto_step` does exist when returning to the previous step on other steps of the form:
![image](https://github.com/user-attachments/assets/1ef37426-94b4-4fef-a07a-cf911427ee1e)
In the above scenario, I am returning to the `otheridentifiers` step from the `grouptransfer` step.

To recreate and view this issue:
- modify `session_token` field's `required` argument to `False` in `transferforms.UploadFilesForm`
- go to your debugging configuration and set `justMyCode` to `false`
- Set a breakpoint on line 367 of `views.py` in `recordtransfer`
- This should take you to the step in the above image

yenaing-oo:
It looks like the lack of the `wizard_goto_step` value in the `POST` request originates from the form submission on the client-side. This is the form that gets submitted when the "Previous Step" button is pressed on the final step:

[...]

PR #290, opened 2024-12-27
Fix invisible reCAPTCHA breaking Previous Step button on final step

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/237

It turns out that the `captcha/js_v2_invisible.html` template was not actually being used to properly override `django_recaptcha/includes/js_v2_invisible.html`. The original template was still being used despite this in `base.html`:

```html
{% if FILE_UPLOAD_ENABLED %}
    {% include "captcha/js_v2_invisible.html" %}
{% endif %}
```

I've renamed that file (`captcha/js_v2_invisible.html`) to match the same folder structure as the actual `include` used in the source code so that Django renders our custom template instead.

It looks like there was difficulty in overriding this exact same template in a past issue https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/92.

### DIFF
--- a/bagitobjecttransfer/recordtransfer/forms/transfer_forms.py
+++ b/bagitobjecttransfer/recordtransfer/forms/transfer_forms.py
@@ -783,7 +783,7 @@ class UploadFilesForm(TransferForm):
         label=gettext("Other notes"),
     )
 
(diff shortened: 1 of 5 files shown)


danloveg (approved):
Excellent detective work - this works flawlessly now.

Issue #291, opened 2025-01-02
File upload overhaul


yenaing-oo:
1. Dropzone.js lacks proper documentation and is not being actively maintained. I believe it would it sensible to replace Dropzone with a different package, such as [Uppy](https://uppy.io/). 
2. Currently, all uploaded files are stored in the same location (`UPLOAD_STORAGE_FOLDER`), including files which are part of an in-progress submission or files which were never saved as part of an in-progress submission. I think it would a good idea to have a separate folder to store temporary uploaded files and then move them to a permanent storage folder once they become part of a completed submission.
3. Currently, there is no capability in the back-end for a user fetch the files that they've uploaded (or at least the metadata of the files) as part of an upload session. I think it would be useful to have this so that the user can go back and forth from the Upload Step of the form, without losing progress of the files that they have uploaded so far.

yenaing-oo:
Trying to fix the Dropzone code and convert jQuery to plain JS in [72-show-summary-of-form](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/tree/72-show-summary-of-form) was what led to the concluding point (1).

Issue #72, opened 2022-03-17
Show Summary of Form / Show Links to All Form Pages


danloveg:
It would be useful to be able to show users all the fields they entered on the final page of the form, or else a way to go back to any page of the form.

whikloj:
I just got the request to allow a user to save a form and come back to it later, it's not quite the same as you use-case but I thought there is a bit of overlap. Not sure if you have a use case for saving/resuming form entries.

PR #292, opened 2025-01-02
File Upload Overhaul

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/291


yenaing-oo:
TODO:
- Removing an already uploaded file should make a call to the backend to delete that file from storage

To consider:
- How long should an upload session be valid? Indefinitely until the session becomes a part of a completed submission? or perhaps since we don't want to hold uploaded files forever if the user never ends up submitting, we might want to invalidate the session and delete uploaded files in that session.
- If invalidating session, would need to include field in response to indicate session has timed out or was invalid, so that client can update their copy.
- Would it be better to store the upload session token in cookies rather than as a hidden form input?

yenaing-oo:
@danloveg 

This PR should be ready for a draft review. I have several failing tests which I can't debug right now because there's something wrong with my VSCode Testing + debugging configuration, which I'm hoping you can help me fix in our next meeting. I'd also like to discuss the items in the Notes section in our meeting as well.

Change summary:
- File upload settings context is no longer passed using a context process, but passed directly to the template to be used in JS

- Uploaded files are initially stored a the TEMP_STORAGE_FOLDER, then moved to UPLOAD_STORAGE_FOLDER upon submission of the transfer form 
(note that a new method `models.UploadSession.move_session_uploads` was used instead of `move_session_uploads` because the `storage` attribute used for an `UploadedFile`'s `file_upload` field has to also change in order to generate the right URL, in addition to the file relocation)

- A path for Django to serve uploaded temp files directly without nginx in development is added to `urls.py`

- A bare-bones final step (Review Step) is added to the transfer form to prepare for the incoming feature https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/72

- Dropzone is replaced with Uppy for file upload in the front-end

Notes:

- I'd like to get your thoughts on this. I think these would be better to address in separate issues.
> How long should an upload session be valid? Indefinitely until the session becomes a part of a completed submission? Or perhaps since we don't want to hold uploaded files forever if the user never ends up submitting, we might want to invalidate the session and delete uploaded files in that session.

> Would it be better to store the upload session token in cookies rather than as a hidden form input?

- Currently, uploaded temp files are served to the client without issue in development, but not in production. I keep getting a 404 error even though everything in `views.media_request` and `nginx.conf` looks correct. I'm not sure how to debug that endpoint in the production container, since my logging statements don't show up.

danloveg:
Morning. This is looking really good, this will be great inclusion to improve how robust file uploads are. I haven't been able to test it out quite yet since I'm still looking at the code so I'll wait for our meeting to see it in action.

> I have several failing tests which I can't debug right now because there's something wrong with my VSCode Testing + debugging configuration, which I'm hoping you can help me fix in our next meeting.

The tests are running in my environment (albeit a few are failing). We can debug it to try and figure out what's going on in your environment.

> I'd like to get your thoughts on this. I think these would be better to address in separate issues.

Yes, please open two separate issues and we can discuss there. For this pull request, let's not worry about making either of those changes.

> - Currently, uploaded temp files are served to the client without issue in development, but not in production. I keep getting a 404 error even though everything in views.media_request and nginx.conf looks correct. I'm not sure how to debug that endpoint in the production container, since my logging statements don't show up.

This sounds like it could still be an issue with `media_request`. I can take a look later today.

yenaing-oo:
@danloveg 
There are a few things I don't understand about this fix (https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/292/commits/70e2fa59e6de4f6e73b7d4e5a911c39b1eac466b):
1) Why does this work, whereas the `media_request` endpoint which also uses the "X-Accel-Redirect" header as well, does not? Perhaps the URL to the file was not being constructed correctly?

2) Does this endpoint, `uploaded_file`, make `media_request` redundant?

3) In `development`, the status code returned upon accessing an uploaded file is `200 OK` and not a `302`, even though the return statement in `uploaded_file` is a `HTTPResponseRedirect`. Is this because Django itself intercepts that request and serves it directly due to this in `urls.py`:?

```python
# Serve media files directly without nginx during development
if settings.DEBUG:
    urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
```

4) The final URL after accessing an uploaded file in production looks like this:
`http://localhost:1337/transfer/upload-session/vehsSsk1f9iPVps3Slbd8tsy2eGbwCGh/files/dcc.pdf/`
whereas the final URL in development looks like this:
`http://localhost:8000/media/temp/Q75UlTYyOdLl7uFW1zN5kRFcli6PUepZ/dcc.pdf`.
Why does the URL is development differ the one in production? Is it because in development, a redirect occurs?

danloveg:
>  2. Does this endpoint, `uploaded_file`, make `media_request` redundant?

No, the `media_request` endpoint will still be hit when someone downloads a BagIt bag from the admin site in production.

>  3. In `development`, the status code returned upon accessing an uploaded file is `200 OK` and not a `302`, even though the return statement in `uploaded_file` is a `HTTPResponseRedirect`. Is this because Django itself intercepts that request and serves it directly due to this in `urls.py`:?
> 
> 
> ```python
> # Serve media files directly without nginx during development
> if settings.DEBUG:
>     urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
> ```

That's a good question. It probably comes down to how Django handles media files, yes.

>  4. The final URL after accessing an uploaded file in production looks like this:
>      `http://localhost:1337/transfer/upload-session/vehsSsk1f9iPVps3Slbd8tsy2eGbwCGh/files/dcc.pdf/`
>      whereas the final URL in development looks like this:
>      `http://localhost:8000/media/temp/Q75UlTYyOdLl7uFW1zN5kRFcli6PUepZ/dcc.pdf`.
>      Why does the URL in development differ from the one in production? Is it because in development, a redirect occurs?

In development, I return the raw path to the media file. Django then does its magic and sends the media file to the browser. In production, I return a path to an API endpoint. Inside that endpoint, I tell NGINX to send the requester the file located at the file's media path by [setting the X-Accel-Redirect header to the path to the file](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/7f35789ced32e71edf0ac26127d42de2c47c77fe/bagitobjecttransfer/recordtransfer/views.py#L1222). This ends up hitting [this location in the NGINX config](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/7f35789ced32e71edf0ac26127d42de2c47c77fe/docker/nginx/nginx.conf#L22). Since the location is `internal`, it looks for the `X-Accel-Redirect` header. If it's there, it serves the file, otherwise, it returns a 404.

danloveg:
> EDIT: I've realised that the `media_request` endpoint is not being hit at all in production. I verified this by placing logging statements at the top of `media_request` and attempting to access a file using a URL such as: `http://localhost:1337/media/temp/EkSaE4LwBI6SFWF70Cuems3yIbu3pVeA/files/dcc.pdf`. My logging statements did not appear, but a 404 request does show up on the `nginx` container. Perhaps `nginx` is intercepting ANY request with a URL beginning with `/media/`, and since the request did not come internally, it sent back a 404.

Forgot to respond to this. Yes that's right, NGINX will intercept all request to `/media/`. If a request is made to `/media/`, and it didn't come from internally, then it will return a 404. Also note, if the request *does* come from internally, but the `X-Accel-Redirect` header is missing, it will also return a 404.

yenaing-oo:
@danloveg 
Thanks for providing clarification on my questions. It makes much more sense now that I know NGINX intercepts all requests made to `/media/`. I'll take a look at how `media_request` is being used for allowing downloads of bags.

I added test classes for the two new views:
- [Add test class for list_uploaded_files view](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/292/commits/a1b5f1644a0d52a7015adf6ba22e1433285bd4a2)
- [Add test class for uploaded_file view](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/292/commits/16620d395a9c44a3da37069cdf3afcddda2cf6e7)

I tried making a test class for the `UploadedFile` model, but I wasn't sure about the right way to write unit tests for some methods that dealt with the file system such as:
```python
class UploadedFile(models.Model):
...
  def copy(self, new_path):
          """Copy this file to a new path.
  
          Args:
              new_path: The new path to copy this file to
          """
          if self.file_upload:
              shutil.copy2(self.file_upload.path, new_path)
  
      def move(self, new_path):
          """Move this file to a new path. Marks this file as removed post-move.
  
          Args:
              new_path: The new path to move this file to
          """
          if not self.file_upload:
              return
  
          directory = os.path.dirname(new_path)
          os.makedirs(directory, exist_ok=True)
  
          shutil.move(self.file_upload.path, new_path)
          self.remove()
 ```
Would you happen to know the appropriate way to test methods like this? If not, I can look further into this.

In addition, I made a change in the way I am constructing URLs for the uploaded files such that I don't use any hardcoded strings such as `temp`, although I'm unsure if this is the best solution:
[Avoid hardcoding URL paths](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/292/commits/7f35789ced32e71edf0ac26127d42de2c47c77fe)
In my method, I am assuming that the `TMP_STORAGE_FOLDER` and `UPLOAD_STORAGE_FOLDER` are both subdirectories of `MEDIA_ROOT`, although this might not always be the case. Is there a better way to do this?

**Comment by danloveg** [comment]
> I tried making a test class for the `UploadedFile` model, but I wasn't sure about the right way to write unit tests for some methods that dealt with the file system such as:
> 
> ```python
> class UploadedFile(models.Model):
> ...
>   def copy(self, new_path):
>           """Copy this file to a new path.
>   
>           Args:
>               new_path: The new path to copy this file to
>           """
>           if self.file_upload:
>               shutil.copy2(self.file_upload.path, new_path)
>   
>       def move(self, new_path):
>           """Move this file to a new path. Marks this file as removed post-move.
>   
>           Args:
>               new_path: The new path to move this file to
>           """
>           if not self.file_upload:
>               return
>   
>           directory = os.path.dirname(new_path)
>           os.makedirs(directory, exist_ok=True)
>   
>           shutil.move(self.file_upload.path, new_path)
>           self.remove()
> ```
> 
> Would you happen to know the appropriate way to test methods like this? If not, I can look further into this.

You can create a temporary directory in the test and use that to do some actual copies. Something like this: https://stackoverflow.com/a/78856008

> In addition, I made a change in the way I am constructing URLs for the uploaded files such that I don't use any hardcoded strings such as `temp`, although I'm not sure if this is the best solution: [Avoid hardcoding URL paths](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/292/commits/7f35789ced32e71edf0ac26127d42de2c47c77fe) In my method, I am assuming that the `TMP_STORAGE_FOLDER` and `UPLOAD_STORAGE_FOLDER` are both subdirectories of `MEDIA_ROOT`, although this might not always be the case. Is there a better way to do this?

Temp storage has to be in `/media/` or else NGINX won't be able to serve the temporarily uploaded file. It's fine to assume that the temp storage and upload storage are in `/media/`. Regarding upload storage: we're getting to a point where it might make sense for people to see their files after a transfer has been submitted. If we do end up doing that, then the upload storage will need to be in media as well.

**Comment by yenaing-oo** [comment]
@danloveg 
> You can create a temporary directory in the test and use that to do some actual copies/moves. Something like this: https://stackoverflow.com/a/78856008

This was exactly what I needed - temp folders to perform test moves and deletes in. It's pretty neat that Python has this module built in. 

I'm overriding the storage folder settings in `test.py` rather than how the stackoverflow post is using the `override_settings()` decorator because that doesn't seem to take effect. I believe it's because the settings used in `FileStorageSystem` classes are initialized at the start and cannot be changed later on.

All my tests in `TestUploadedFile` pass in my environment, but I can't figure out why they're failing in the workflow. Could you please a look?

**Comment by danloveg** [comment]
> All my tests in `TestUploadedFile` pass in my environment, but I can't figure out why they're failing in the workflow. Could you please a look?

I'm taking a look now, will let you know what I find.

**Comment by danloveg** [comment]
> > All my tests in `TestUploadedFile` pass in my environment, but I can't figure out why they're failing in the workflow. Could you please a look?
> 
> I'm taking a look now, will let you know what I find.

@yenaing-oo, `SimpleUploadedFile` inherits from `InMemoryUploadedFile` (see [here](https://github.com/django/django/blob/b0d309c9eb802cbc652595e2d413bb451e37f124/django/core/files/uploadedfile.py#L127)), meaning that the file may not exist on the disk.

I believe you should be using Django's file API rather than `pathlib` whenever possible for operations on uploaded files. For example, you had the right idea here when checking if a file exists: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/9fbfc49a2c64bb66b3cd62e3bbbade35af36e2e3/bagitobjecttransfer/recordtransfer/models.py#L185

And another example, you had the right idea to delete files like this instead of calling `unlink`: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/9fbfc49a2c64bb66b3cd62e3bbbade35af36e2e3/bagitobjecttransfer/recordtransfer/models.py#L266

**Comment by danloveg** [comment]
@yenaing-oo these tests also fail on my computer due to how paths are constructed in Windows:

- `TestUploadedFile.test_get_temp_file_media_url`

```
E       AssertionError: '\\media\\temp3aMMY1CQTegbXz4QMZJ6sinJBW78eHSa/test.pdf' != '/media/\\temp3aMMY1CQTegbXz4QMZJ6sinJBW78eHSa/test.pdf'
E       - \media\temp3aMMY1CQTegbXz4QMZJ6sinJBW78eHSa/test.pdf
E       ? ^
E       + /media/\temp3aMMY1CQTegbXz4QMZJ6sinJBW78eHSa/test.pdf
E       ? ^     +
```

- `TestUploadedFile.test_get_permanent_file_media_url`

```
E       AssertionError: '\\media\\uploadsQJKhiZvGi03jZIAsrY3edXYEIM2AvFHk/test.pdf' != '/media/\\uploadsQJKhiZvGi03jZIAsrY3edXYEIM2AvFHk/test.pdf'
E       - \media\uploadsQJKhiZvGi03jZIAsrY3edXYEIM2AvFHk/test.pdf
E       ? ^
E       + /media/\uploadsQJKhiZvGi03jZIAsrY3edXYEIM2AvFHk/test.pdf
E       ? ^     +
```

- `TestUploadedFileView.test_get_uploaded_file_in_debug`

```
E       AssertionError: '%5Cmedia%5CtempIUlSCJQLU0uWXafTl9yBzy7ocoLjh7Rx/testfile.txt' != '\\media\\tempIUlSCJQLU0uWXafTl9yBzy7ocoLjh7Rx/testfile.txt'
E       - %5Cmedia%5CtempIUlSCJQLU0uWXafTl9yBzy7ocoLjh7Rx/testfile.txt
E       ? ^^^     ^^^
E       + \media\tempIUlSCJQLU0uWXafTl9yBzy7ocoLjh7Rx/testfile.txt
E       ? ^     ^
```

**Comment by yenaing-oo** [comment]
> @yenaing-oo, `SimpleUploadedFile` inherits from `InMemoryUploadedFile` (see [here](https://github.com/django/django/blob/b0d309c9eb802cbc652595e2d413bb451e37f124/django/core/files/uploadedfile.py#L127)), meaning that the file may not exist on the disk.
> 
> I believe you should be using Django's file API rather than `pathlib` whenever possible for operations on uploaded files. 

Thanks for taking a look at this.

I see what you mean about how `SimpleUploadedFile` could be living in memory instead of on disk. I was under the impression that once I pass a `SimpleUploadedFile` to the `UploadedFile` model to make an instance [here](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/9fbfc49a2c64bb66b3cd62e3bbbade35af36e2e3/bagitobjecttransfer/recordtransfer/tests/test_models.py#L215) and then save it, the file would be persisted somewhere on disk. 

I do get back an actual location on disk when I call `self.uploaded_file.file_upload.path` like I am doing [here](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/9fbfc49a2c64bb66b3cd62e3bbbade35af36e2e3/bagitobjecttransfer/recordtransfer/tests/test_models.py#L235), but perhaps using the OS's file system to perform file operations on that path directly might not be appropriate because it might mess with how Django is managing the files internally.

Does this mean I should also avoid using `shutil` in `UploadedFile.copy()` and `UploadedFile.move()`? Unfortunately, I don't see any method for doing copies and moves on a Django `FileField`: https://docs.djangoproject.com/en/4.1/ref/models/fields/#filefield

Perhaps the only way safe way to copy a model's `FileField` would be to create a new instance of a different model to represent an uploaded file that is not temporary, similar to this: https://stackoverflow.com/a/21482851/15193287

**Comment by yenaing-oo** [comment]
> @yenaing-oo these tests also fail on my computer due to how paths are constructed in Windows:

I'll try building the paths differently to account for the OS.

**Comment by yenaing-oo** [comment]
@danloveg 
Should the `accept_file` view and endpoint be removed, since we're already performing the underlying `_accept_file()` check in `upload_file`?

**Comment by danloveg** [comment]
> @danloveg Should the [`accept_file` view](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/443e5a3eb1dad47f964a7220e14a18624f248f71/bagitobjecttransfer/recordtransfer/views.py#L941) and endpoint be removed, since we're already performing the underlying `_accept_file()` check in `upload_file`?

That makes sense to me, wherever we can simplify things, we should.

File: bagitobjecttransfer/recordtransfer/storage.py [hunk 1]
```diff
@@ -1,25 +1,53 @@
...
+        location_path = Path(location).resolve()
+
+        if not location_path.is_relative_to(media_root):
+            raise ValueError("The location must be a subdirectory of MEDIA_ROOT")
+        kwargs["location"] = str(location)
         super().__init__(**kwargs)
+
+    def url(self, name: str) -> str:
+        """Generate the URL based on MEDIA_URL and the relative path."""
+        return f"/{Path(self.base_location).relative_to(settings.BASE_DIR) / name}"
```
danloveg, line 39: This is technically a URL, so we should use `urljoin` here. As a benefit, this is platform-agnostic and won't create a path with backslasehs like the current code is doing in Windows.

```suggestion
        return urljoin(self.base_url, name)
```

bagitobjecttransfer/recordtransfer/storage.py
```diff
@@ -1,25 +1,53 @@
...
+
+    def __init__(self, location: str, **kwargs):
+        """Set the location of the storage. The location has to be a subdirectory of MEDIA_ROOT."""
+        # Check if the location is a subdirectory of MEDIA_ROOT
+        media_root = Path(settings.MEDIA_ROOT).resolve()
+        location_path = Path(location).resolve()
+
+        if not location_path.is_relative_to(media_root):
+            raise ValueError("The location must be a subdirectory of MEDIA_ROOT")
+        kwargs["location"] = str(location)
```
danloveg, line 34: Django storage has [a way to specify a base_url](https://docs.djangoproject.com/en/5.1/ref/files/storage/#django.core.files.storage.FileSystemStorage.base_url), we should use that. I also recommend using `MEDIA_URL` as the basis of forming the base_url.

```suggestion
        if not location_path.is_relative_to(media_root):
            raise ValueError("The location must be a subdirectory of MEDIA_ROOT")

        rel_path = Path(settings.UPLOAD_STORAGE_FOLDER).relative_to(Path(settings.MEDIA_ROOT))
        rel_path = rel_path.as_posix()
        base_url = urljoin(settings.MEDIA_URL, rel_path)

        kwargs["location"] = str(location)
        kwargs["base_url"] = base_url
```

bagitobjecttransfer/recordtransfer/storage.py
```diff
@@ -1,25 +1,53 @@
+from pathlib import Path
```
danloveg, line 1: Need urljoin for a change I highlighted below.

```suggestion
from pathlib import Path
from urllib.parse import urljoin
```

bagitobjecttransfer/recordtransfer/tests/unit/test_models.py
```diff
@@ -113,59 +139,167 @@ def test_delete_files_in_session(self, uploadedfile_set_mock):
...
+        self.assertFalse(self.uploaded_file.temp)
+
+    def test_get_temp_file_media_url(self) -> None:
+        """Test that the file media URL is returned."""
+        self.assertEqual(
+            self.uploaded_file.get_file_media_url(),
+            f"""/{
+                Path(settings.TEMP_STORAGE_FOLDER).relative_to(settings.BASE_DIR)
+                / self.uploaded_file.file_upload.name
+            }"""
```
danloveg, line 267:
```suggestion
            urljoin(
                Path(settings.TEMP_STORAGE_FOLDER).relative_to(settings.BASE_DIR).as_posix(),
                self.uploaded_file.file_upload.name
            )
```

bagitobjecttransfer/recordtransfer/tests/unit/test_models.py
```diff
@@ -113,59 +139,167 @@ def test_delete_files_in_session(self, uploadedfile_set_mock):
...
+
+    def test_get_permanent_file_media_url(self) -> None:
+        """Test that the file media URL is returned."""
+        self.uploaded_file.move_to_permanent_storage()
+        self.assertEqual(
+            self.uploaded_file.get_file_media_url(),
+            f"""/{
+                Path(settings.UPLOAD_STORAGE_FOLDER).relative_to(settings.BASE_DIR)
+                / self.uploaded_file.file_upload.name
+            }"""
```
danloveg, line 278:
```suggestion
            urljoin(
                Path(settings.UPLOAD_STORAGE_FOLDER).relative_to(settings.BASE_DIR).as_posix()
                self.uploaded_file.file_upload.name,
            )
```

bagitobjecttransfer/bagitobjecttransfer/settings/test.py
```diff
@@ -105,5 +105,10 @@
...
 
-BAG_STORAGE_FOLDER = config("BAG_STORAGE_FOLDER", default=os.path.join(BASE_DIR, "media/bags/"))
-UPLOAD_STORAGE_FOLDER = config("UPLOAD_STORAGE_FOLDER", default=os.path.join(BASE_DIR, "media/uploaded_files/"))
+BASE_DIR = tempfile.mkdtemp()
+MEDIA_ROOT = Path(BASE_DIR) / "media"
+TEMP_STORAGE_FOLDER = Path(MEDIA_ROOT) / "temp"
+TEMP_STORAGE_FOLDER.mkdir(parents=True, exist_ok=True)
+UPLOAD_STORAGE_FOLDER = Path(MEDIA_ROOT) / "uploads"
+UPLOAD_STORAGE_FOLDER.mkdir(parents=True, exist_ok=True)
+BAG_STORAGE_FOLDER = Path(MEDIA_ROOT) / "bags"
```
danloveg, line 114: These are strings in the other files, we should make these strings to keep consistent with other settings files. I can imagine a case where a test passes while testing if a Path operation is used on one of these, whereas it'll fail in production or development if that same Path operation is used (because these are strings in dev/prod).

danloveg (changes requested):
This is my first pass of reviewing this PR. First, I'm looking at storage and media URLs. We should be using `urljoin` to join media URLs instead of Pathlib. The changes I recommended in this review will likely cause a bunch of the other tests to fail - I didn't include fixes to most of those tests here.

bagitobjecttransfer/recordtransfer/tests/unit/test_models.py
yenaing-oo: The reason why I avoided using `urljoin` is because it works differently from the way I expected it to. For e.g.:

image
If the second argument of `urljoin` does not begin with a `/` (see above), it as treated as a relative path and replaces everything after **the first `/`** of the first argument.

image

If the second argument of `urljoin` does begin with a '/' (see above), it is treated as an absolute path and overrides the entire final path.

I'm trying to find a good solution for this, but the best I can come up with so far is to use `pathlib` then `as_posix()` on the path to get a path with single forward-slashes only.

danloveg: That makes sense, didn't realize `urljoin` had that behaviour. You could even do this:

```
Path(settings.TEMP_STORAGE_FOLDER).relative_to(settings.BASE_DIR).as_posix() +
"/" +
self.uploaded_file.file_upload.name
```

bagitobjecttransfer/bagitobjecttransfer/settings/test.py
yenaing-oo: Addressed by:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/292/commits/2f2f69c930409018a2b4e32bcd3e7f3af69401e1

bagitobjecttransfer/recordtransfer/tests/unit/test_models.py
yenaing-oo: Using `as_posix()` on `base_url` `Path` instead of using `url_join`

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/292/commits/152ecf548e49cc79804dcac70cb6b3d58491026b

bagitobjecttransfer/recordtransfer/tests/unit/test_models.py
yenaing-oo: Using `as_posix()` on `base_url` `Path` instead of using `url_join`

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/292/commits/152ecf548e49cc79804dcac70cb6b3d58491026b

bagitobjecttransfer/recordtransfer/storage.py
yenaing-oo: Using `as_posix()` on `base_url` `Path` instead of using `url_join`

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/292/commits/152ecf548e49cc79804dcac70cb6b3d58491026b

bagitobjecttransfer/recordtransfer/storage.py
yenaing-oo: Using `as_posix()` on `base_url` `Path` instead of using `url_join`

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/292/commits/152ecf548e49cc79804dcac70cb6b3d58491026b

bagitobjecttransfer/recordtransfer/storage.py
yenaing-oo: Using `as_posix()` on `base_url` `Path` instead of using `url_join`

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/292/commits/152ecf548e49cc79804dcac70cb6b3d58491026b

danloveg (changes requested):
The tests all pass on my machine now.

Next up, I noticed a few issues with submitting files and downloading them:

1. If you click "Next Step" on the upload page without adding any files or filling out the general notes, you get the confusing message in the following screenshot. I believe this is due to the required `session_token` field not passing validation. Can the error message be moved down to the file upload box, and have a better message like "you must upload at least one file"?

Screenshot 2025-01-15 141337

2. Related to 1. If you upload a file, and then delete it, you can go to the next step without having any fields filled out. There should be a `clean` method in `recordtransfer.forms.transfer_forms.UploadFilesForm` that checks whether the session has one or more files.

3. The "Copy Link" button in Uppy doesn't have much use. Can it be disabled?

Screenshot 2025-01-15 141542

4. I get errors when downloading a BagIt bag for a submission in the admin. It seems to be using the old temp paths when grabbing the files. Here's a sample error log:

```
INFO 2025-01-15 14:09:52,262 jobs: Creating zipped bag from /app/bagitobjecttransfer/media/bags/danloveg/danloveg_20250115-140952_my-images
INFO 2025-01-15 14:09:52,291 jobs: No bag exists at /app/bagitobjecttransfer/media/bags/danloveg/danloveg_20250115-140952_my-images, creating it now.
INFO 2025-01-15 14:09:52,300 models: Created new user folder at "/app/bagitobjecttransfer/media/bags/danloveg"
INFO 2025-01-15 14:09:52,307 models: Created new bag folder at "/app/bagitobjecttransfer/media/bags/danloveg"
INFO 2025-01-15 14:09:52,317 models: Copying 3 temp files to /app/bagitobjecttransfer/media/bags/danloveg/danloveg_20250115-140952_my-images
ERROR 2025-01-15 14:09:52,320 models: File "/app/bagitobjecttransfer/media/temp/02dDxNQAJ1jL2G1wvEgJ2SE0sEpwJfTU/golddisc.png" was moved or deleted
ERROR 2025-01-15 14:09:52,322 models: File "/app/bagitobjecttransfer/media/temp/02dDxNQAJ1jL2G1wvEgJ2SE0sEpwJfTU/conch_ascii.png" was moved or deleted
ERROR 2025-01-15 14:09:52,325 models: File "/app/bagitobjecttransfer/media/temp/02dDxNQAJ1jL2G1wvEgJ2SE0sEpwJfTU/SFU_Transfer3_Utilities.pdf" was moved or deleted
ERROR 2025-01-15 14:09:52,325 models: One or more uploaded files is missing!
INFO 2025-01-15 14:09:52,325 models: Removing bag at "/app/bagitobjecttransfer/media/bags/danloveg/danloveg_20250115-140952_my-images" due to missing files
```

yenaing-oo:
> 1. If you click "Next Step" on the upload page without adding any files or filling out the general notes, you get the confusing message in the following screenshot. I believe this is due to the required `session_token` field not passing validation. Can the error message be moved down to the file upload box, and have a better message like "you must upload at least one file"?

I don't think there's a good way to change the location of the error message (red text) that gets sent back as part of the cleaning of the form, since the fields and the errors are being rendered through a template.

I can, however, make sure that there is a BE (addressed in 2.) and FE check with an Uppy error message that prevents the user from attempting form submission without a session_token.

Addressed by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/292/commits/ccb9e39822d468455f9336a10a206cc1578a5655 and 
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/292/commits/37816c4f6d52e8a4e3a1c184e7363e5b0a947002

> 2. Related to 1. If you upload a file, and then delete it, you can go to the next step without having any fields filled out. There should be a `clean` method in `recordtransfer.forms.transfer_forms.UploadFilesForm` that checks whether the session has one or more files.

This makes sense.
Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/292/commits/f79d22619882a9f2bfd188215443f1090378569c


> 3. The "Copy Link" button in Uppy doesn't have much use. Can it be disabled?

There is no way hide just that button, since it comes with the Uppy's `showLinkToFileUploadResult` feature, but I can hide it through CSS.

Addressed by:
 https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/292/commits/8b3e8b00156790fc17c0b1331ab85c40b01ef330

> 4. I get errors when downloading a BagIt bag for a submission in the admin. It seems to be using the old temp paths when grabbing the files. Here's a sample error log:

I suspect that even if we change the storage class of the `file_upload` field in `move_to_permanent_storage()`, when the model instance is recreated, it uses the default attributes that was provided when the model was defined (i.e. it uses `TempFileStorage`):
```
class UploadedFile(models.Model):
    """Represent a file that a user uploaded during an upload session."""

    name = models.CharField(max_length=256, null=True, default="-")
    session = models.ForeignKey(UploadSession, on_delete=models.CASCADE, null=False)
    file_upload = models.FileField(
        null=True, storage=TempFileStorage, upload_to=session_upload_location
    )
    temp = models.BooleanField(default=True)
```

We might have to try making a new model to separate temporary uploaded files and permanent ones.

bagitobjecttransfer/recordtransfer/tests/unit/test_utils.py
```diff
@@ -109,123 +127,326 @@ class FileCountingUtilityTests(TestCase):
...
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        """Set up class-level test data."""
+        super().setUpClass()
+        settings.ACCEPTED_FILE_FORMATS = {"Document": ["docx", "pdf"], "Spreadsheet": ["xlsx"]}
+        settings.MAX_TOTAL_UPLOAD_SIZE = 3  # MiB
+        settings.MAX_SINGLE_UPLOAD_SIZE = 1  # MiB
+        settings.MAX_TOTAL_UPLOAD_COUNT = 4  # Number of files
+
```
danloveg, line 256: I recommend patching these settings in this class instead:

```suggestion

@patch(
    "django.conf.settings.ACCEPTED_FILE_FORMATS",
    {"Document": ["docx", "pdf"], "Spreadsheet": ["xlsx"]},
)
@patch("django.conf.settings.MAX_TOTAL_UPLOAD_SIZE", 3)  # MiB
@patch("django.conf.settings.MAX_SINGLE_UPLOAD_SIZE", 1)  # MiB
@patch("django.conf.settings.MAX_TOTAL_UPLOAD_COUNT", 4)  # Number of files
class TestAcceptFile(TestCase):
    """Tests for accept_file method."""

```

bagitobjecttransfer/recordtransfer/tests/unit/test_utils.py
```diff
@@ -109,123 +127,326 @@ class FileCountingUtilityTests(TestCase):
...
```
danloveg, line 344: I recommend patching these settings here instead:

```suggestion

@patch(
    "django.conf.settings.ACCEPTED_FILE_FORMATS",
    {"Document": ["docx", "pdf"], "Spreadsheet": ["xlsx"]},
)
@patch("django.conf.settings.MAX_TOTAL_UPLOAD_SIZE", 3)  # MiB
@patch("django.conf.settings.MAX_SINGLE_UPLOAD_SIZE", 1)  # MiB
@patch("django.conf.settings.MAX_TOTAL_UPLOAD_COUNT", 4)  # Number of files
class TestAcceptSession(TestCase):
    """Tests for accept_session method."""

```

bagitobjecttransfer/recordtransfer/tests/unit/test_utils.py
```diff
@@ -109,123 +127,326 @@ class FileCountingUtilityTests(TestCase):
...
+                    result = accept_session(name, "1024", self.session_1)
+                    self.assertFalse(result["accepted"])
+                    self.assertIn(
+                        "A file with the same name has already been uploaded",
+                        result["error"],
+                    )
+
+        finally:
+            for file_ in files:
+                file_.delete()
```
danloveg, line 452: You can simplify these test by using a [tearDown method](https://docs.python.org/3.10/library/unittest.html#unittest.TestCase.tearDown).

```suggestion
    def tearDown(self) -> None:
        """Remove all uploaded files after each test."""
        UploadedFile.objects.all().delete()

    def test_session_has_room(self) -> None:
        """Test that a session with room for more files is accepted."""
        # 2 MiB of files (one MiB x 2)
        for name in ("File 1.docx", "File 2.docx"):
            UploadedFile.objects.create(
                session=self.session_1,
                file_upload=SimpleUploadedFile(name, self.one_mib),
                name=name,
            )

        for size in ("1024", len(self.one_mib)):
            with self.subTest():
                result = accept_session("My File.pdf", size, self.session_1)
                self.assertTrue(result["accepted"])

    def test_session_file_count_full(self) -> None:
        """Test that a session with the maximum number of files is rejected."""
        # 2 MiB of files (half MiB x 4)
        # Max file count is 4
        for name in ("File 1.docx", "File 2.pdf", "File 3.pdf", "File 4.pdf"):
            UploadedFile.objects.create(
                session=self.session_1,
                name=name,
                file_upload=SimpleUploadedFile(name, self.half_mib),
            )

        result = accept_session("My File.pdf", "1024", self.session_1)
        self.assertFalse(result["accepted"])
        self.assertIn("You can not upload anymore files", result["error"])

    def test_file_too_large_for_session(self) -> None:
        """Test that a file that is too large for the session is rejected."""
        # 2.5 MiB of files (1 Mib x 2, 0.5 MiB x 1)
        for name, content in (
            ("File 1.docx", self.one_mib),
            ("File 2.pdf", self.one_mib),
            ("File 3.pdf", self.half_mib),
        ):
            UploadedFile.objects.create(
                session=self.session_1,
                name=name,
                file_upload=SimpleUploadedFile(name, content),
            )

        result = accept_session("My File.pdf", len(self.one_mib), self.session_1)
        self.assertFalse(result["accepted"])
        self.assertIn("Maximum total upload size (3 MiB) exceeded", result["error"])

    def test_duplicate_file_name(self) -> None:
        """Test that a file with the same name as an existing file is rejected."""
        names = ("File.1.docx", "File.2.pdf")
        for name in names:
            UploadedFile.objects.create(
                session=self.session_1,
                name=name,
                file_upload=SimpleUploadedFile(name, self.half_mib),
            )

        for name in names:
            with self.subTest():
                result = accept_session(name, "1024", self.session_1)
                self.assertFalse(result["accepted"])
                self.assertIn(
                    "A file with the same name has already been uploaded",
                    result["error"],
                )
```

bagitobjecttransfer/recordtransfer/views.py
```diff
@@ -824,323 +855,181 @@ def get_in_progress_submission(user: User, uuid: str) -> Optional[InProgressSubm
...
-            the session is valid, or False if not. The dictionary also contains
-            an 'error' and 'verboseError' key if 'accepted' is False.
+        JsonResponse: A JSON response containing the list of uploaded files and their details,
+        or an error message if the session is not found.
     """
+    session = UploadSession.objects.filter(token=session_token).first()
+    if not session:
+        # Create a new session if one is not found
+        session = UploadSession.new_session()
+        session.save()
```
danloveg, line 961: Is there a need for this behaviour? A GET request shouldn't create resources on the backend. I don't think the uploadSessionToken from this request is even being used.

danloveg (changes requested):

bagitobjecttransfer/recordtransfer/static/recordtransfer/js/transferform/utils.js
```diff
@@ -0,0 +1,163 @@
...
+    }
+};
+
+/**
+ * Retrieves file upload settings from context passed from view. Note that this function
+ * should only be called after the DOM has loaded, otherwise the settings element may not exist
+ * yet.
+ * @returns {?object} The parsed JSON settings object if element exists, null otherwise.
+ */
+export const getSettings = () => {
```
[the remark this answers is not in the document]
yenaing-oo: Addressed by:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/292/commits/9175789d347d810262f6c4425ac30a94e71213eb

bagitobjecttransfer/recordtransfer/tests/unit/test_utils.py
yenaing-oo: Addressed by:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/292/commits/4056271a440453918fc63e5482ed5609c528319c

bagitobjecttransfer/recordtransfer/tests/unit/test_utils.py
yenaing-oo: Addressed by:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/292/commits/7022bf6a87fa86a5e97c3b59f6fd272b4f96a937

danloveg (approved):
I'm really happy with these changes. The new Javascript for Uppy is much cleaner and easier to understand (and thanks for not including any more jQuery).

Besides the two bugs we know we're including with this pull request (#313 and #302), everything looks good to me.

Issue #293, opened 2025-01-03
Retain Unsaved Form Data When Navigating to a Previous Step


yenaing-oo:
Navigating back to a previous step in the form results in losing all unsaved data from the current step. Saving the form before going back is inconvenient, as it redirects the user to the profile page.

Unsaved form data should be retained when moving back to a previous step for a better user experience.

danloveg:
That sounds like a good improvement for the user experience. Do you have any ideas for how we can accomplish this? I don't think you can submit the form on the page if it's incomplete. Might make sense to cache it on the client side?

yenaing-oo:
> Navigating back to a previous step in the form results in losing all unsaved data from the current step. Saving the form before going back is inconvenient, as it redirects the user to the profile page.
> 
> Unsaved form data should be retained when moving back to a previous step for a better user experience.

This was what I was thinking too. We would only have to cache the last step we reached that hadn't been validated yet on the client side.

yenaing-oo:
> That sounds like a good improvement for the user experience. Do you have any ideas for how we can accomplish this? I don't think you can submit the form on the page if it's incomplete. Might make sense to cache it on the client side?

This was what I was thinking too. We would only have to cache the last step we reached that hadn't been validated yet on the client side.

danloveg:
> > That sounds like a good improvement for the user experience. Do you have any ideas for how we can accomplish this? I don't think you can submit the form on the page if it's incomplete. Might make sense to cache it on the client side?
> 
> This was what I was thinking too. We would only have to cache the last step we reached that hadn't been validated yet on the client side. Submitting the incomplete form would not be possible with the current setup.

Sounds like a job for local storage. We'll need to think about how the form field cache would get invalidated or expire it so that the cached form doesn't get applied when it shouldn't. For example, if Step 3 gets cached, and I close the browser and try to submit another form, those cached fields shouldn't get applied.

yenaing-oo:
https://stackoverflow.com/questions/5523140/are-there-any-benefits-to-session-storage-over-local-storage

> the data stored in localStorage persists until explicitly deleted. Changes made are saved and available for all current and future visits to the site.
> 
> For sessionStorage, changes are only available per tab. Changes made are saved and available for the current page in that tab until it is closed. Once it is closed, the stored data is deleted.

[...]

PR #303, opened 2025-01-10
Save current step of form data before going to previous step

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/293

Initial solution:
This solution makes use of the existing logic in `TransferFormWizard.post` to handle the case where both `SAVE_FORM_ACTION` and `GO_TO_STEP_ACTION` is included in the submitted form data. Hidden inputs for these fields are added onto the form when the user clicks on the previous step button.

I'm not using `sessionStorage` like we discussed in https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/293, but perhaps this is a cleaner solution.

Additional change:
- An error message is shown if the form fails to save its progress when the user clicks on Previous step
image

EDIT:
New solution:
Override `SessionWizardView`'s `render_goto_step()` method to save the current step's data before rendering the desired step.


danloveg (changes requested):
This doesn't appear to be working correctly. I've noticed two issues:

1. The link "Click to save form and resume later" takes you to the next page of the form instead of saving the transfer.
2. Going to the previous step and back to the next step does not add the semi-filled form data.

An example to reproduce issue 2:

- Fill out steps 1-3 of the form.
- On step 4, add the title "Minutes"
- Click Previous Step
- Click Next Step

"Minutes" won't be in the title.

I'm confused about where the previous form data is supposed to be added to the current step of the form. I can see where you save the current data but not where it gets added to the form when you click Next Step.

yenaing-oo:
> 2. Going to the previous step and back to the next step does not add the semi-filled form data.
I figured out the issue, and why it seemed to work fine when I tested it. When I tested this, it was using a resumed in-progress submission. This meant that that the view had access to the in-progress submission model instance. 

On a fresh transfer form, going back to a previous step does indeed save the current step data along with previous step data, but every time this happens, a new `InProgressSubmission` was created:

In `save_transfer()`:
```python

        if submission:
            submission.last_updated = timezone.now()
        else:
            submission = InProgressSubmission()
```
Since the URL of the transfer form page does not include the query parameter `transfer_uuid` of the newly created InProgressSubmission, it's not able to access the saved data.

I found a more elegant solution that makes use of `django-formtools`'s own class methods: https://stackoverflow.com/questions/19039246/django-form-wizard-save-and-go-to-previous-step

I made use of that solution here: [Save current step data before going back to a previous step](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/303/commits/49422fcac48872e5b075300c656b49db11842f51). This should now save non-valid form data before rendering a previous step.

yenaing-oo:
> 1. When the form gets partially filled upon clicking "Next Step", error messages are rendered even though the user hasn't submitted that part of the form yet:

I don't know why the errors didn't appear for me when I tested it initially, I'll be more thorough next time. 
It looks like a form's data is validated automatically when initial data is passed to it, so my workaround to this is to clear the errors before rendering the next step's form to the user.
Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/303/commits/dcdde8e8384862c16a6340dcb72cd949d8105c20

> 2. If you click Previous Step on page 2, and then click Next Step to go back to page 2, a KeyError is raised:
This was due to a lack of initial value for the "province_or_state" field data, and the way its value was being looked up using `[]` rather than `.get()`.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/303/commits/d889f2db639918cecf2c3eb8b29272f13a234442 and https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/303/commits/393720f047c009245cc3030bcc622813b684621b

> 3. Can you add a couple tests to prove that this behaviour works as intended? We've not tested the form wizard yet I don't think so I don't know how easily those tests will be to write.

I added an E2E test using Selenium to test that going back to a previous step retains the data in the current step.
Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/303/commits/020515b76ef5b75a6178bae1ba6abad84660b1ca

danloveg:
> EDIT: Looks like I also need to install Chrome and the necessary drives on the workflow for the test to work.

If this isn't possible then I'm okay with the E2E tests being excluded from that workflow.

yenaing-oo:
> If this isn't possible then I'm okay with the E2E tests being excluded from that workflow.

Looks like it is possible, although it did take around 15 seconds for the E2E test class to run. Do we still want to keep it in the workflow?

danloveg:
> > If this isn't possible then I'm okay with the E2E tests being excluded from that workflow.
> 
> Looks like it is possible, although it did take around 15 seconds for the E2E test class to run. Do we still want to keep it in the workflow? If so, I'll go ahead and mark all the unit tests with the decorator `@pytest.mark.unit` so we can run unit tests separately from E2E tests.
> 
> EDIT: It might be actually longer, looking at the logs. It took 31 or 49 seconds for each of the workflows respectively.

Thanks for testing that out. Let's exclude e2e tests for now and focus on running unit tests only in the workflow.

danloveg (approved):
The incomplete form saving works great, no issues there. This is a clean implementation that didn't require any JS, which is a plus.

The new end-to-end test work as well. Thanks for setting that up, having that example will be very helpful if/when we need to write more end-to-end tests in the future.

Issue #297, opened 2025-01-08
Run Asset Build in Containerfile


danloveg:
JS/CSS/Image assets should be built in the Containerfile as opposed to in the entrypoint. Running the asset bundling in the entrypoint slows the app/rq container start time, something we should aim to minimize as much as possible.

For production, asset bundling only needs to happen once, so it can happen before the container starts up (i.e., it can be run in the Containerfile).

For development, asset bundling does need to happen more than once, but with the `npm run watch` command the bundling can be executed whenever a change to a file is made.

PR #304, opened 2025-01-13
Run asset build in Containerfile instead of entrypoint

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/297

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -9,16 +9,16 @@ RUN python -m pip install --user pipx && \
 
 WORKDIR /app/
 
+# Copy poetry-related files, and install Python dependencies
+COPY pyproject.toml poetry.lock README.md /app/
+RUN poetry config virtualenvs.create false && poetry install
+
 # Copy Node-related files, and install NodeJS dependencies
 COPY package*.json webpack.config.js /app/
 # Copy script files
 COPY ./scripts /app/scripts
 
 RUN npm install --no-color
-
-# Copy poetry-related files, and install Python dependencies
-COPY pyproject.toml poetry.lock README.md /app/
-RUN poetry config virtualenvs.create false && poetry install
 
 # Copy entrypoint script to image
 COPY ./docker/entrypoint.sh /app/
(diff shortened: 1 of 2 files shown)


docker/entrypoint.sh
```diff
@@ -16,9 +16,6 @@ if [ "$IS_RQ" = 'yes' ]; then
 
 # Bundle, minify, and collect static assets in the Django container
```
danloveg, line 17: I don't think this comment is necessary anymore.

Containerfile
```diff
@@ -29,4 +29,7 @@ COPY ./bagitobjecttransfer /app/bagitobjecttransfer
 
 WORKDIR /app/bagitobjecttransfer/
 
+# Run webpack to bundle and minify assets
+RUN echo ">> Running webpack to bundle + minify assets." && npm run build
```
danloveg, line 33: The `package.json` file is in `/app/`, I think it makes more sense to run the npm command while in `/app/` before moving into `/app/bagitobjecttransfer/`.

I also don't think the `echo` is necessary, if someone wants to see what's going on when building the image, they'll open the `Containerfile` up and see the comment there.

```suggestion
# Run webpack to bundle and minify assets
RUN npm run build

WORKDIR /app/bagitobjecttransfer/
```

danloveg (changes requested):
The container starts running much faster for me now. I made a couple of small comments but otherwise this works great.

docker/entrypoint.sh
yenaing-oo: Resolved by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/304/commits/07f1be85e3d896762731c006cd1335873f17bae5

Containerfile
yenaing-oo: Resolved by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/304/commits/07f1be85e3d896762731c006cd1335873f17bae5

yenaing-oo:
Thanks for the feedback, all suggested changes make sense to me and have been applied.

danloveg (approved):

Issue #300, opened 2025-01-08
NGINX client_max_body_size does not match MAX_SINGLE_UPLOAD_SIZE


danloveg:
The `nginx.conf` file should be changed to a template. This template should set the `client_max_body_size` to the same value as `MAX_SINGLE_UPLOAD_SIZE`.

Currently, the front-end may report that a file is too big, regardless of what the MAX_SINGLE_UPLOAD_SIZE is set to.

yenaing-oo:
@danloveg 
The size value of `client_max_body_size` uses megabytes as the unit. However, the environment variable `MAX_SINGLE_UPLOAD_SIZE` is in mebibytes. How should we deal with this? 

I think it would be better to use MB consistently throughout the app. I recall you mentioning that the reason we are using MiB  is because Windows uses MiB but displays sizes as MB. Perhaps displaying the uploaded file size counter in MB in the Upload Step is good enough to address this? It appears to me that this is the only place where MiB is used, which has forced the entire app to use MiB as well.

yenaing-oo:
Sounds good, I've opened up an issue for that here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/305

yenaing-oo:
@danloveg 
According to one of the comments on this [StackOverflow post](https://stackoverflow.com/a/28476755/15193287), the NGINX's documentation on file sizes and how it interprets file sizes in its source code are inconsistent.

It looks like in the docs, the file sizes are implied to be decimal-based because "megabytes" are used instead of "mebibytes". But for bytes, the gave an example value of "1024":
https://nginx.org/en/docs/syntax.html
This might be a hint that the actual units are binary-based.

In the [source code](https://github.com/nginx/nginx/blob/release-1.23.1/src/core/ngx_parse.c#L13) from the comment, they're definitely using binary-based unit conversions.

I think we can keep the current configuration the way it is if this is the case.

yenaing-oo:
Since `client_max_body_size` is actually confirmed to be in `MiB`, we can go ahead and use `MAX_SINGLE_UPLOAD_SIZE` to configure `client_max_body_size` in a template.

PR #306, opened 2025-01-13
Make NGINX Config into Template

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/300

Passes `MAX_SINGLE_UPLOAD_SIZE` as an environment variable within `.prod.env` to the nginx container, where it is used within an nginx configuration template.


docker/nginx/templates/nginx.conf.template
```diff
@@ -11,17 +11,15 @@ server {
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header Host $host;
         proxy_redirect off;
-
-        # client_max_body_size 1024M;
     }
```
danloveg, line 15: Since we only need the body size large when uploading files, I think we can add a new location like this for uploading files.

```suggestion
    location /transfer/uploadfile/ {
        proxy_pass http://record_transfer_app;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header Host $host;
        proxy_redirect off;
        client_max_body_size ${MAX_SINGLE_UPLOAD_SIZE}m;
    }
```

docker/nginx/templates/nginx.conf.template
```diff
@@ -35,4 +33,6 @@ server {
         internal;
         alias /var/www/html/500.html;
     }
+
+    client_max_body_size ${MAX_SINGLE_UPLOAD_SIZE}m;
```
danloveg, line 37: Can remove if the new location is added.

danloveg (changes requested):

docker/nginx/templates/nginx.conf.template
yenaing-oo: Since `/transfer/uploadfile` is a path under `/`, I think it is sufficient to define just `client_max_body_size` under `/transfer/uploadfile` as the other configurations will be shared.

```
location /transfer/uploadfile/ {
    client_max_body_size ${MAX_SINGLE_UPLOAD_SIZE}m;
}
```

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/306/commits/a3e68424a8c1cf0da0eac580339b15c8a147b742

yenaing-oo: Looks like nginx only supports directive inheritance from the parent level, not for parent paths. I think your suggestion is the best way to set the configuration for now.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/306/commits/c925ca73f56f1494886f4817a1bab8ee45084f5c

danloveg (approved):

PR #308, opened 2025-01-13
Convert compress_images.js to Webpack Plugin

danloveg:
This pull request removes `compress_images.js` in favour of using a Webpack plugin.


yenaing-oo (approved):
I didn't think of making the script into a plugin, that's pretty neat. This works flawlessly - the same images are being converted into webp as before.

PR #309, opened 2025-01-14
Fix Cases where Lazy Log Formatting is not Used

### LINKED ISSUE #242: Fix Cases Where Lazy Log Formatting is not Used
There are a few places where lazy log formatting is not used:

- `recordtransfer.models.UploadSession`
- `recordtransfer.utils.count_file_types`
- `recordtransfer.views.uploadfiles`

These should be changed to use "%s" in the log string rather than calling `.format()` on the string.

danloveg:
Removes all identified cases where non-lazy log formatting is used.

This pull request would conflict with #292 in `UploadSession` and `uploadfiles`; this pull request is being left as Draft until that pull request is resolved.

Closes #242

### DIFF
--- a/bagitobjecttransfer/recordtransfer/tests/test_utils.py
+++ b/bagitobjecttransfer/recordtransfer/tests/test_utils.py
@@ -1,60 +1,55 @@
+from unittest.mock import patch
+
 from django.test import TestCase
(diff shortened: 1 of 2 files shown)


danloveg:
Just going to restart on a different branch, this was getting out of hand.

Issue #310, opened 2025-01-14
WEBPACK_MODE is undefined in Containerfile


danloveg:
The environment variable WEBPACK_MODE is not defined when `npm run build` is run when the container is built. A way needs to be added to set the webpack mode when the container is built. If there's no simple way to accomplish setting WEBPACK_MODE to dev/prod, then production mode should be used as the default.

PR #312, opened 2025-01-15
Fix undefined WEBPACK_MODE in Containerfile

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/310

Passes `WEBPACK_MODE` as an `arg` from the compose files to the Containerfile, and sets it as an environment variable  there, which `npm run build` has access to.

### DIFF
--- a/.github/workflows/publish-to-ghcr.yml
+++ b/.github/workflows/publish-to-ghcr.yml
@@ -49,6 +49,8 @@ jobs:
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+          build-args: |
+            WEBPACK_MODE=production
 
       - name: Generate artifact attestation
         uses: actions/attest-build-provenance@v1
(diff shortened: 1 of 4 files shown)


danloveg (changes requested):
Works great! I have no issues with these changes, but the build arguments will need to be passed when the container is built in `.github/workflows/publish-to-ghcr.yml` as well.

I think this should work (see build-args at the end):

```yaml
      - name: Build and push Docker image
        id: push
        uses: docker/build-push-action@v6
        with:
          context: .
          file: Containerfile
          push: true
          tags: ${{ steps.meta.outputs.tags }}
          labels: ${{ steps.meta.outputs.labels }}
          build-args: |
            WEBPACK_MODE=production
```

yenaing-oo:
> ... build arguments will need to be passed when the container is built in .github/workflows/publish-to-ghcr.yml as well.

This makes sense. I've addressed it here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/312/commits/b89642ff02e097ca4b6cc008032302c12a0e3ff0

danloveg (approved):

PR #317, opened 2025-01-16
Fix non-lazy logging calls

danloveg:
Fixes all remaining non-lazy log calls, and removes the non-standard `msg=()` syntax.


yenaing-oo (approved):
Looks like those are the last of the non-lazy log calls. Everything looks good!

Issue #313, opened 2025-01-16
Separate UploadedFile into Two Models to Represent Temporary and Permanent


yenaing-oo:
This issue is dependent on https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/292 being merged in.

There are errors thrown when creating a downloadable bag from a submission from the admin page, causing the bagging job to fail. This is because the files are being looked for in their old temp paths, rather than their new one. 

Here's a sample error log:

```
INFO 2025-01-15 14:09:52,262 jobs: Creating zipped bag from /app/bagitobjecttransfer/media/bags/danloveg/danloveg_20250115-140952_my-images
INFO 2025-01-15 14:09:52,291 jobs: No bag exists at /app/bagitobjecttransfer/media/bags/danloveg/danloveg_20250115-140952_my-images, creating it now.
INFO 2025-01-15 14:09:52,300 models: Created new user folder at "/app/bagitobjecttransfer/media/bags/danloveg"
INFO 2025-01-15 14:09:52,307 models: Created new bag folder at "/app/bagitobjecttransfer/media/bags/danloveg"
INFO 2025-01-15 14:09:52,317 models: Copying 3 temp files to /app/bagitobjecttransfer/media/bags/danloveg/danloveg_20250115-140952_my-images
ERROR 2025-01-15 14:09:52,320 models: File "/app/bagitobjecttransfer/media/temp/02dDxNQAJ1jL2G1wvEgJ2SE0sEpwJfTU/golddisc.png" was moved or deleted
ERROR 2025-01-15 14:09:52,322 models: File "/app/bagitobjecttransfer/media/temp/02dDxNQAJ1jL2G1wvEgJ2SE0sEpwJfTU/conch_ascii.png" was moved or deleted
ERROR 2025-01-15 14:09:52,325 models: File "/app/bagitobjecttransfer/media/temp/02dDxNQAJ1jL2G1wvEgJ2SE0sEpwJfTU/SFU_Transfer3_Utilities.pdf" was moved or deleted
ERROR 2025-01-15 14:09:52,325 models: One or more uploaded files is missing!
INFO 2025-01-15 14:09:52,325 models: Removing bag at "/app/bagitobjecttransfer/media/bags/danloveg/danloveg_20250115-140952_my-images" due to missing files
```

[...]

PR #318, opened 2025-01-17
Separate UploadedFile into two models

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/313
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/324


yenaing-oo:
@danloveg 
This PR is still a draft, since I have yet to fix existing tests and add new ones.

But could you please take a rough look at my changes to spot any issues? In particular, I'd like your feedback on:

1.  The way I am splitting the `UploadedFile` model into two separate ones `TempUploadedFile` and `PermUploadedFile`, both of which are child model classes of the abstract class `BaseUploadedFile`
2. The way I am dealing with uploaded files, which could be either temporary or permanent, in `UploadSession`, based on whether the session is expired
3. Whether you foresee any issues moving forward with this sort of design pattern in 1. and 2.
4. The display of information and the different functionalities offered to an admin user on the admin page.

The "Generate downloadable bag" feature should be working fine now. Please feel free to provide feedback on anything else.

danloveg:
Reviewing now, will have some feedback for you shortly

danloveg (commented):
Rather than comment on one specific section of code, I have some general comments:

The purpose of an `UploadSession` is to track both the temporary files someone uploaded, and whether those files were moved to permanent storage. There are a number of states that the session can be in:

- Created (no files yet)
- Temp files only (one or more temp files, zero perm files)
- Copying in progress (temp files being moved to permanent storage)
- Perm files only (zero temp files, one or more perm files)
- Copying failed (files could not be moved to perm storage for some reason)
- Expired (no files, and not accepting any more files)

Note that a session may transition into expired if and only if the session is in the Created or Temp files only state.

The purpose of a `BaseUploadedFile` is to track a file, but also which storage it's in. The session + the name of a file uniquely identifies that file, so that there is no file with the same session + name in both the `TempUploadedFile` and `PermUploadedFile` tables.

The way the `UploadSession` is structured now hides the temp files once `expired=True`. I think `UploadSession` should look like this:

```python

    class SessionStatus(models.TextChoices):
        """The status of the session."""

        CREATED = "CR", _("Session Created")
        EXPIRED = "EX", _("Session Expired")
        TEMPORARY_FILES = "TF", _("All Files in Temporary Storage")
        COPYING_IN_PROGRESS = "CP", _("Copying Files to Permanent Storage")
        PERMANENT_FILES = "PF", _("All Files in Permanent Storage")
        COPYING_FAILED = "FD", _("Copying Failed")

    token = models.CharField(max_length=32)
    started_at = models.DateTimeField()

    status = models.CharField(
        max_length=2, choices=SessionStatus.choices, default=SessionStatus.CREATED
    )

    @classmethod
    def new_session(cls) -> "UploadSession":
        """Start a new upload session."""

    @property
    def upload_size(self) -> int:
        """Get total size (in bytes) of all files in this session.

        This includes the size of both temporary and permanent files.
        """

    @property
    def file_count(self) -> int:
        """Get the total count of temporary + permanent uploaded files."""

    def get_temporary_uploads(self) -> list[TempUploadedFile]:
        """Get a list of temporary uploaded files associated with this session.

        May be empty if temp uploads have already been moved to permanent storage.
        """

    def get_permanent_uploads(self) -> list[PermUploadedFile]:
        """Get a list of permanent uploaded files associated with this session.

        May be empty if temp uploads have not been moved.
        """

    def remove_uploads(self) -> None:
        """Remove all uploaded files associated with this session."""

    def make_uploads_permanent(self, logger: Optional[logging.Logger] = None) -> None:
        """Make all temporary uploaded files associated with this session permanent."""

    def copy_session_uploads(
        self, destination: str, logger: Optional[logging.Logger] = None
    ) -> tuple[list[str], list[str]]:
        """Copy permanent files associated with this session to a destination directory."""

    def __str__(self):
        """Return a string representation of this object."""
```

The session's state should be updated and checked based on what's going on in these functions. For example, you can skip `make_uploads_permanent` if `session.status in (SessionStatus.PERMANENT_FILES, SessionStatus.COPYING_IN_PROGRESS)`. Once uploads have been moved to permanent storage, you can set status to PERMANENT_FILES and save the session.

And one final comment, `PermUploadedFile` does not need a `move_to_permanent_storage` function. It's fine to implement that function only on the `TempUploadedFile` model.

yenaing-oo:
> Rather than comment on one specific section of code, I have some general comments:
> 
> The purpose of an `UploadSession` is to track both the temporary files someone uploaded, and whether those files were moved to permanent storage. There are a number of states that the session can be in:
> 
> * Created (no files yet)
> * Temp files only (one or more temp files, zero perm files)
> * Copying in progress (temp files being moved to permanent storage)
> * Perm files only (zero temp files, one or more perm files)
> * Copying failed (files could not be moved to perm storage for some reason)
> * Expired (no files, and not accepting any more files)
> 
> Note that a session may transition into expired if and only if the session is in the Created or Temp files only state.
> 
> The purpose of a `BaseUploadedFile` is to track a file, but also which storage it's in. The session + the name of a file uniquely identifies that file, so that there is no file with the same session + name in both the `TempUploadedFile` and `PermUploadedFile` tables.
> 
> The way the `UploadSession` is structured now hides the temp files once `expired=True`. I think `UploadSession` should look like this:
> 
> The session's state should be updated and checked based on what's going on in these functions. For example, you can skip `make_uploads_permanent` if `session.status in (SessionStatus.PERMANENT_FILES, SessionStatus.COPYING_IN_PROGRESS)`. Once uploads have been moved to permanent storage, you can set status to PERMANENT_FILES and save the session.
> 

Thank you for the detailed feedback. As you pointed out, I was indeed trying to hide the temp files once `expire=True`. My intention was to create an interface that made it so that the user does not have to worry about the type of `BaseUploadedFile` they're getting back. But your approach makes sense since we might want more control, given that file operations aren't instantaneous and are also prone to failure. 


> And one final comment, `PermUploadedFile` does not need a `move_to_permanent_storage` function. It's fine to implement that function only on the `TempUploadedFile` model.

I see. If we know exactly the type of `BaseFileUpload` we're getting, it makes sense to have `move_to_permanent_storage` only on the `TempUploadedFile` model.

I'll apply these changes.

yenaing-oo:
@danloveg 

>The purpose of an UploadSession is to track both the temporary files someone uploaded, and whether those files were moved to permanent storage. There are a number of states that the session can be in:
> - Created (no files yet)
> - Temp files only (one or more temp files, zero perm files)
> - Copying in progress (temp files being moved to permanent storage)
> - Perm files only (zero temp files, one or more perm files)
> - Copying failed (files could not be moved to perm storage for some reason)
> - Expired (no files, and not accepting any more files)

> ```python
> class SessionStatus(models.TextChoices):
>         """The status of the session."""
> 
>         CREATED = "CR", _("Session Created")
>         EXPIRED = "EX", _("Session Expired")
>         TEMPORARY_FILES = "TF", _("All Files in Temporary Storage")
>         COPYING_IN_PROGRESS = "CP", _("Copying Files to Permanent Storage")
>         PERMANENT_FILES = "PF", _("All Files in Permanent Storage")
>         COPYING_FAILED = "FD", _("Copying Failed")
> ```

After giving this more thought, I have some follow-up questions:
1. Can a `TEMPORARY_FILES` session transition to `EXPIRED`, and if so, how are we dealing with the already uploaded temp files? When you say "Expired (no files..)", does this mean the files associated with the session have been hidden or deleted?
2. If some temp files fail to be copied to perm storage, do we allow the session to be associated with a mix of temp and perm uploaded files, or do we roll back the changes we made? 
3. Could you provide some state transition flows that include `EXPIRED` and `COPYING_FAILED`? The only flow that is obvious to me is  CREATED -> TEMPORARY_FILES -> COPYING_IN_PROGRESS -> PERMANENT_FILES

danloveg:
>     1. Can a `TEMPORARY_FILES` session transition to `EXPIRED`, and if so, how are we dealing with the already uploaded temp files? When you say "Expired (no files..)", does this mean the files associated with the session have been hidden or deleted?

For sure, a session may be expired after creation and no files were uploaded, or after some files were uploaded, but the user did not submit the form.

When a session is expired it essentially becomes unusable. The temporary files associated with that session should be deleted.

>     2. If some temp files fail to be copied to perm storage, do we allow the session to be associated with a mix of temp and perm uploaded files, or do we roll back the changes we made?

I intended the `COPYING_FAILED` state to describe that situation. Focus on the different states right now and we can look at how to recover from a failed copy after the states are sorted out.

>     3. Could you provide some state transition flows that include `EXPIRED` and `COPYING_FAILED`? The only flow that is obvious to me is  CREATED -> TEMPORARY_FILES -> COPYING_IN_PROGRESS -> PERMANENT_FILES

For sure, here you go:

Screenshot 2025-01-20 110016

danloveg:
> @danloveg
> 
> Thanks for the clarification. I have a few more follow-up questions.
> 
>  1. Do we need additional states for:
> 
>  * when files are being removed
>       I can imagine a situation where `remove_uploads` is called and while this is running, one of `copy_session_uploads`, `make_uploads_permanent`, `get_temporary_uploads`, `get_permanent_uploads` is called

Yes, a state for removing does make sense to include. It follows that we would want a DELETED state too.

>  * when files are being copied from permanent storage to a destination directory
>       Similar to before, it is possible for `remove_uploads` to be called while `copy_session_uploads` is taking place.

Interesting question. Yes, this state should be included, but is it possible to reuse the same copying state for this as well as for temp->perm?

>  2. Should we prevent `get_temporary_uploads` and `get_permanent_uploads` from being called (perhaps by raising an error) if `remove_uploads` or `make_uploads_permanent` is taking place?

Yes, that makes sense to me. If a temp file gets returned mid-copy, that temp file may not exist the next time it gets accessed.

I think these changes point to this new flowchart:

uploadsession flowchart 2

yenaing-oo:
> Interesting question. Yes, this state should be included, but is it possible to reuse the same copying state for this as well as for temp->perm?

Yes, it makes sense to me to reuse the same copying state for this as well.

I have another question:
Do we want detailed error messages for different cases [like this](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/944e9c1e36eea6a8295d910bf81468ddbbcf875f/bagitobjecttransfer/recordtransfer/models.py#L127-L147), or is it enough to simply state that a method failed because it was not in the right session status [like this](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/944e9c1e36eea6a8295d910bf81468ddbbcf875f/bagitobjecttransfer/recordtransfer/models.py#L180-L184)?

danloveg:
> Unrelated, I didn't know you could add flowcharts here like that - that's very neat!

I also didn't know, but figured I'd just try it out! I use Mermaid for generating flowcharts in Markdown, and GitHub supports it apparently!

> I have another question: Do we want detailed error messages for different cases [like this](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/944e9c1e36eea6a8295d910bf81468ddbbcf875f/bagitobjecttransfer/recordtransfer/models.py#L127-L147), or is it enough to simply state that a method failed because it was not in the right session status [like this](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/944e9c1e36eea6a8295d910bf81468ddbbcf875f/bagitobjecttransfer/recordtransfer/models.py#L180-L184)?

I personally prefer those detailed error messages. Will the number of different error messages become unwieldy with the number of states we now have, or is it a manageable number?

yenaing-oo:
> I personally prefer those detailed error messages. Will the number of different error messages become unwieldy with the number of states we now have, or is it a manageable number?

I see, that's good to know. Just wanted to check what your preference was. I like detailed error messages as well. I think it should be manageable if we group them together such as how I grouped the "copying in progress" and "removal in progress" together:
```python
"""Cannot get temporary uploaded files from session {self.token} while copy or
removal of files is in progress"""
``` 
One drawback is that we would also have to be diligent is updating the codebase in multiple places if we introduce another `SessionStatus`, since we're catching each case individually rather than using `if self.status !=`

yenaing-oo:
I think we should make a manager class for adding/removing and fetching files from a session.

Right now, I'm having to update the status of the session myself after creating a `TempUploadedFile` in `views.py` [here](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/944e9c1e36eea6a8295d910bf81468ddbbcf875f/bagitobjecttransfer/recordtransfer/views.py#L930-L931)

Another place I see this manager coming into play is [here](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/944e9c1e36eea6a8295d910bf81468ddbbcf875f/bagitobjecttransfer/recordtransfer/views.py#L1019). We probably want to avoid deleting a temp file if `UploadSession.make_uploads_permanent` is already running. Instead of checking the status here ourselves, we could let a manager class handle that.

Additional notes:
- avoid adding/removing a temp file if session is in `REMOVING_IN_PROGRESS` or `COPYING_IN_PROGRESS` status
- `get_temp_uploads` handles cases where the status is not `TEMPORARY_FILES`, should the manager handle this instead? Same goes for all the other state checks in the `UploadSession`'s methods.

danloveg:
> I think we should make a manager class for adding/removing and fetching files from a session.
> 
> Right now, I'm having to update the status of the session myself after creating a `TempUploadedFile` in `views.py` [here](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/944e9c1e36eea6a8295d910bf81468ddbbcf875f/bagitobjecttransfer/recordtransfer/views.py#L930-L931)

The upload session is kind of your file manager already. I do see the issue you're talking about, but it could be solved with this function:

```python
    def add_temp_file(self, file):
        if not self.state in (self.SessionStatus.CREATED, self.TEMPORARY_FILES):
            raise ValueError("some descriptive error message here")

        temp_file = TempUploadedFile(session=self, file_upload=file, name=file.name)
        temp_file.save()

        if self.state == self.SessionStatus.CREATED:
            self.state = self.SessionStatus.TEMPORARY_FILES
            self.save()
```

> Another place I see this manager coming into play is [here](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/944e9c1e36eea6a8295d910bf81468ddbbcf875f/bagitobjecttransfer/recordtransfer/views.py#L1019). We probably want to avoid deleting a temp file if `UploadSession.make_uploads_permanent` is already running. Instead of checking the status here ourselves, we could let a manager class handle that.

Similar to my above example, you could add some logic to the upload session for doing that checking.

```python
    def remove_temp_file_by_name(self, name: str):
        try:
            temp_file = self.tempuploadedfile_set.get(name=name)
        except TempUploadedFile.DoesNotExist:
            raise FileNotFoundError(f"There is no temporary file with the name '{name}' in this session")

        if self.state != self.SessionStatus.TEMPORARY_FILES:
            raise ValueError("Cannot remove temp files unless in TEMPORARY_FILES state")

        temp_file.delete()
```

danloveg:
> For the temp file removal method, should we update the status to `CREATED` if the `file_count` is back to zero? This makes sense to me since `CREATED` does not represent anything more than a state that has zero files. If so, I believe the state diagram would have to be updated like this:

That makes sense to me, yes.

I was thinking about this last night, and what do you think about renaming TEMPORARY_FILES to UPLOADING? I think it needs to be given a better name that reflects the fact that a user is currently in the process of uploading files to that session.

yenaing-oo:
> > For the temp file removal method, should we update the status to `CREATED` if the `file_count` is back to zero? This makes sense to me since `CREATED` does not represent anything more than a state that has zero files. If so, I believe the state diagram would have to be updated like this:
> 
> That makes sense to me, yes.
> 
> I was thinking about this last night, and what do you think about renaming TEMPORARY_FILES to UPLOADING? I think it needs to be given a better name that reflects the fact that a user is currently in the process of uploading files to that session.

I agree, `UPLOADING` would describe the session better and is more consistent with how the other enums are describing the session rather than what the session holds.

danloveg:
> > > For the temp file removal method, should we update the status to `CREATED` if the `file_count` is back to zero? This makes sense to me since `CREATED` does not represent anything more than a state that has zero files. If so, I believe the state diagram would have to be updated like this:
> > 
> > 
> > That makes sense to me, yes.
> > I was thinking about this last night, and what do you think about renaming TEMPORARY_FILES to UPLOADING? I think it needs to be given a better name that reflects the fact that a user is currently in the process of uploading files to that session.
> 
> I agree, `UPLOADING` would describe the session better and is more consistent with how the other enums are describing the session rather than what the session holds.

Okay great. I want to mention now too that there are going to be a bunch of migrations made for this pull request since we've been making small tweaks to the `SessionStatus` enum, but we'll want to squash them later with [`squashmigrations`](https://docs.djangoproject.com/en/4.2/topics/migrations/#squashing-migrations). Not something that needs to be done now but something to keep in mind for later.

yenaing-oo:
> Okay great. I want to mention now too that there are going to be a bunch of migrations made for this pull request since we've been making small tweaks to the `SessionStatus` enum, but we'll want to squash them later with [`squashmigrations`](https://docs.djangoproject.com/en/4.2/topics/migrations/#squashing-migrations). Not something that needs to be done now but something to keep in mind for later.

I see, it's good to know there's a built-in way to do that in Django.

bagitobjecttransfer/recordtransfer/models.py
```diff
@@ -2,6 +2,7 @@
 import os
```
danloveg, line 2: Add `from __future__ import annotations` at top of file.

bagitobjecttransfer/recordtransfer/models.py
```diff
@@ -39,138 +42,364 @@ def get_full_name(self):
...
+        of both temporary and permanent files.
+        """
+        if self.status in (
+            self.SessionStatus.EXPIRED,
+            self.SessionStatus.DELETED,
+        ):
+            raise ValueError(
+                f"""Cannot get upload size from session {self.token} because the session has
+                expired or has been deleted"""
+            )
```
danloveg, line 84: Using a triple """ preserves the extra whitespace on the indented line (starting with "expired") and preserves the newline after the word "has". This should be changed to a string on a single line:

```suggestion
            raise ValueError(
                f"Cannot get upload size from session {self.token} because the session has "
                "expired or has been deleted"
            )
```

It's not as clean looking but doesn't have that unwanted side effect that """ does.

bagitobjecttransfer/recordtransfer/models.py
```diff
@@ -39,138 +42,364 @@ def get_full_name(self):
...
+        elif self.status in (
+            self.SessionStatus.COPYING_IN_PROGRESS,
+            self.SessionStatus.REMOVING_IN_PROGRESS,
+        ):
+            raise ValueError(
+                f"""Cannot get upload size from session {self.token} while copying or removing
+                files is in progress"""
+            )
```
danloveg, line 92:
```suggestion
            raise ValueError(
                f"Cannot get upload size from session {self.token} while copying or removing "
                "files is in progress"
            )
```

bagitobjecttransfer/recordtransfer/models.py
```diff
@@ -39,138 +42,364 @@ def get_full_name(self):
...
+    def file_count(self) -> int:
+        """Get the total count of temporary + permanent uploaded files."""
+        if self.status in (
+            self.SessionStatus.EXPIRED,
+            self.SessionStatus.DELETED,
+        ):
+            raise ValueError(
+                f"""Cannot get file count from session {self.token} because the session has
+                expired or has been deleted"""
+            )
```
danloveg, line 109:
```suggestion
            raise ValueError(
                f"Cannot get file count from session {self.token} because the session has "
                "expired or has been deleted"
            )
```

bagitobjecttransfer/recordtransfer/models.py
```diff
@@ -39,138 +42,364 @@ def get_full_name(self):
...
+        elif self.status in (
+            self.SessionStatus.COPYING_IN_PROGRESS,
+            self.SessionStatus.REMOVING_IN_PROGRESS,
+        ):
+            raise ValueError(
+                f"""Cannot get file count from session {self.token} while copying or removing files
+                is in progress"""
+            )
```
danloveg, line 117:
```suggestion
            raise ValueError(
                f"Cannot get file count from session {self.token} while copying or removing files "
                "is in progress"
            )
```

bagitobjecttransfer/recordtransfer/models.py
```diff
@@ -39,138 +42,364 @@ def get_full_name(self):
...
-        """Get all files from the uploadedfile_set where the file exists"""
-        return [f for f in self.uploadedfile_set.all() if f.exists]
+    def add_temp_file(self, file: UploadedFile) -> "TempUploadedFile":
+        """Add a temporary uploaded file to this session."""
+        if self.status not in (self.SessionStatus.CREATED, self.SessionStatus.UPLOADING):
+            raise ValueError(
+                f"""Cannot add temporary uploaded file to session {self.token} because the session
+                status is {self.status} and not {self.SessionStatus.CREATED} or
+                {self.SessionStatus.UPLOADING}"""
+            )
```
danloveg, line 130: Similar to my other suggestions, """ should be changed to ", but I'm also suggesting you use the names from the enum rather than their values.

This has the effect of outputting "... and not CREATED or UPLOADING" versus "... and not CR or UG". The human readable names should be preferred in error messages.

```suggestion
            raise ValueError(
                f"Cannot add temporary uploaded file to session {self.token} because the session "
                f"status is {self.status.name} and not {self.SessionStatus.CREATED.name} or "
                f"{self.SessionStatus.UPLOADING.name}"
            )
```

bagitobjecttransfer/recordtransfer/models.py
```diff
@@ -39,138 +42,364 @@ def get_full_name(self):
...
+        return sum(
+            f.exists
+            for f in chain(self.permuploadedfile_set.all(), self.tempuploadedfile_set.all())
+        )
 
-    def get_existing_file_set(self):
-        """Get all files from the uploadedfile_set where the file exists"""
-        return [f for f in self.uploadedfile_set.all() if f.exists]
+    def add_temp_file(self, file: UploadedFile) -> "TempUploadedFile":
```
danloveg, line 123: Since we're going to include `from __future__ import annotations` at the top of this file, you can refer to the class name without quotes.

```suggestion
    def add_temp_file(self, file: UploadedFile) -> TempUploadedFile:
```

bagitobjecttransfer/recordtransfer/models.py
```diff
@@ -39,138 +42,364 @@ def get_full_name(self):
...
+            raise ValueError(
+                f"""Cannot remove temporary uploaded file from session {self.token} because the
+                session status is {self.status} and not {self.SessionStatus.UPLOADING}"""
+            )
+        try:
+            temp_file = self.tempuploadedfile_set.get(name=name)
+        except TempUploadedFile.DoesNotExist:
+            raise FileNotFoundError(
+                f"No temporary file with name {name} exists in session {self.token}"
+            ) from None
```
danloveg, line 153:
```suggestion
        except TempUploadedFile.DoesNotExist as exc:
            raise FileNotFoundError(
                f"No temporary file with name {name} exists in session {self.token}"
            ) from exc
```

bagitobjecttransfer/recordtransfer/models.py
```diff
@@ -39,138 +42,364 @@ def get_full_name(self):
...
+
+        return temp_file
+
+    def remove_temp_file_by_name(self, name: str) -> None:
+        """Remove a temporary uploaded file from this session by name."""
+        if self.status != self.SessionStatus.UPLOADING:
+            raise ValueError(
+                f"""Cannot remove temporary uploaded file from session {self.token} because the
+                session status is {self.status} and not {self.SessionStatus.UPLOADING}"""
+            )
```
danloveg, line 147:
```suggestion
            raise ValueError(
                f"Cannot remove temporary uploaded file from session {self.token} because the "
                f"session status is {self.status.name} and not {self.SessionStatus.UPLOADING.name}"
            )
```

bagitobjecttransfer/recordtransfer/models.py
```diff
@@ -39,138 +42,364 @@ def get_full_name(self):
...
 
-    def remove_session_uploads(self, logger=None):
-        """Remove uploaded files associated with this session.
+        temp_file.delete()
+
+        if self.file_count == 0:
+            self.status = self.SessionStatus.CREATED
+            self.save()
+
+    def get_temp_file_by_name(self, name: str) -> Optional["TempUploadedFile"]:
```
danloveg, line 161: What do you think about raising a FileNotFoundError instead of returning None here?

bagitobjecttransfer/recordtransfer/models.py
```diff
@@ -39,138 +42,364 @@ def get_full_name(self):
...
+
+        if self.status != self.SessionStatus.UPLOADING:
+            raise ValueError(
+                f"""Can only get temporary uploaded files from session {self.token} when the
+                session status is {self.SessionStatus.UPLOADING}"""
+            )
+
+        return self.tempuploadedfile_set.filter(name=name).first()
+
+    def get_temporary_uploads(self) -> list["TempUploadedFile"]:
```
danloveg, line 179:
```suggestion
    def get_temporary_uploads(self) -> list[TempUploadedFile]:
```

bagitobjecttransfer/recordtransfer/models.py
```diff
@@ -39,138 +42,364 @@ def get_full_name(self):
...
+            self.SessionStatus.COPYING_IN_PROGRESS,
+            self.SessionStatus.REMOVING_IN_PROGRESS,
+        ):
+            raise ValueError(
+                f"""Cannot get temporary uploaded files from session {self.token} while copy or
+                removal of files is in progress"""
+            )
+        return [f for f in self.tempuploadedfile_set.all() if f.exists]
+
+    def get_permanent_uploads(self) -> list["PermUploadedFile"]:
```
danloveg, line 204:
```suggestion
    def get_permanent_uploads(self) -> list[PermUploadedFile]:
```

bagitobjecttransfer/recordtransfer/models.py
```diff
@@ -39,138 +42,364 @@ def get_full_name(self):
...
+            self.SessionStatus.COPYING_IN_PROGRESS,
+            self.SessionStatus.REMOVING_IN_PROGRESS,
+        ):
+            raise ValueError(
+                f"""Cannot get permanent uploaded files from session {self.token} while copy or
+                removal of files is in progress"""
+            )
+        return [f for f in self.permuploadedfile_set.all() if f.exists]
+
+    def get_uploads(self) -> Union[list["TempUploadedFile"], list["PermUploadedFile"]]:
```
danloveg, line 232:
```suggestion
    def get_uploads(self) -> Union[list[TempUploadedFile], list[PermUploadedFile]]:
```

bagitobjecttransfer/recordtransfer/models.py
```diff
@@ -39,138 +42,364 @@ def get_full_name(self):
...
+            )
+
+    def remove_uploads(self) -> None:
+        """Remove all uploaded files associated with this session."""
+        if self.status == self.SessionStatus.REMOVING_IN_PROGRESS:
+            LOGGER.info("File removal is already in progress for session %s", self.token)
+            return
+        elif self.status == self.SessionStatus.CREATED:
+            LOGGER.info("There are no uploaded files in the session %s to remove", self.token)
+            return
```
danloveg, line 257:
```suggestion
        if self.status == self.SessionStatus.REMOVING_IN_PROGRESS:
            LOGGER.warning("File removal is already in progress for session %s", self.token)
            return
        elif self.status == self.SessionStatus.CREATED:
            LOGGER.warning("There are no uploaded files in the session %s to remove", self.token)
            return
```

bagitobjecttransfer/recordtransfer/models.py
```diff
@@ -39,138 +42,364 @@ def get_full_name(self):
...
+                f"""Cannot remove uploaded files from session {self.token} while copying files is
+                in progress"""
+            )
+
+        self.status = self.SessionStatus.REMOVING_IN_PROGRESS
+        self.save()
 
-    def move_uploads_to_permanent_storage(self, logger: Optional[logging.Logger] = None) -> None:
-        """Move uploaded files from temporary to permanent storage.
+        files = list(self.permuploadedfile_set.all()) + list(self.tempuploadedfile_set.all())
```
danloveg, line 272: I liked your use of `chain` above, would that work here too?

bagitobjecttransfer/recordtransfer/tests/unit/test_views.py
```diff
@@ -287,11 +272,13 @@ def test_list_uploaded_files_empty_session(self):
...
     def test_list_uploaded_files_with_files(self):
         """Session has one file."""
-        uploaded_file = UploadedFile(
+        uploaded_file = TempUploadedFile(
             session=self.session,
             file_upload=SimpleUploadedFile("testfile.txt", self.one_kib),
             name="testfile.txt",
         )
+        self.session.status = UploadSession.SessionStatus.UPLOADING
+        self.session.save()
```
danloveg, line 281: What do you think about doing this instead?

```suggestion
        self.session.add_temp_file(SimpleUploadedFile("testfile.txt", self.one_kib))
```

bagitobjecttransfer/recordtransfer/tests/unit/test_views.py
```diff
@@ -323,12 +310,13 @@ def setUpTestData(cls):
...
         self.session = UploadSession.new_session()
-        self.session.save()
-        self.uploaded_file = UploadedFile(
+        self.uploaded_file = TempUploadedFile(
             session=self.session,
             file_upload=SimpleUploadedFile("testfile.txt", self.one_kib),
             name="testfile.txt",
         )
+        self.session.status = UploadSession.SessionStatus.UPLOADING
+        self.session.save()
```
danloveg, line 319: Similar to my other suggestion in this file.

```suggestion
        self.session.add_temp_file(SimpleUploadedFile("testfile.txt", self.one_kib))
```

danloveg (changes requested):
This is working really well. Nothing is *broken*, but I do have a number of comments regarding code style, especially regarding error messages.

I stopped commenting about the use of """ partway through models.py but I still would like to see all the error messages using """ changed to ". Similarly, all cases where a SessionStatus is shown in the error message should use the `.name` property of the enum member instead of the value (which is what's used by default).

Finally, can you run this command to squash migration 0043 to 0049?

```shell
python manage.py squashmigrations --squashed-name split_file_into_two_models recordtransfer 0043_storeduploadedfile_tempuploadedfile_and_more 0049_alter_uploadsession_status
```

You can delete the old un-squashed migrations afterwards.

bagitobjecttransfer/recordtransfer/models.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/27a5f765eb73a3faa444e2a40977b38728ce2c2d

bagitobjecttransfer/recordtransfer/models.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/27a5f765eb73a3faa444e2a40977b38728ce2c2d

bagitobjecttransfer/recordtransfer/models.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/27a5f765eb73a3faa444e2a40977b38728ce2c2d

bagitobjecttransfer/recordtransfer/models.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/f216239bd7f9c5e521579aaa15340758e0bcb71f

bagitobjecttransfer/recordtransfer/models.py
yenaing-oo: Addressed by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/8ace49c7759f1c1f5fb089f9d652b71a54497865

bagitobjecttransfer/recordtransfer/models.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/27a5f765eb73a3faa444e2a40977b38728ce2c2d

bagitobjecttransfer/recordtransfer/models.py
yenaing-oo: Addressed by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/8ace49c7759f1c1f5fb089f9d652b71a54497865

bagitobjecttransfer/recordtransfer/models.py
yenaing-oo: Addressed by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/8ace49c7759f1c1f5fb089f9d652b71a54497865

bagitobjecttransfer/recordtransfer/models.py
yenaing-oo: Addressed by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/8ace49c7759f1c1f5fb089f9d652b71a54497865

bagitobjecttransfer/recordtransfer/models.py
yenaing-oo: Addressed bv: [Include error traceback](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/15a158004bc96133a69f441536cb60cfbd6e1000)

bagitobjecttransfer/recordtransfer/models.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/6b657e3ac893ef49d6a0aaacb0e4852585d51ccd

bagitobjecttransfer/recordtransfer/models.py
yenaing-oo: Addressed by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/f6fc71c7837b84bb846082cdc0cf31de9d17ac94

bagitobjecttransfer/recordtransfer/tests/unit/test_views.py
yenaing-oo: Addressed by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/b590646a934d658a3e69aabf9a36df474355a883

yenaing-oo:
@danloveg 
As you suggested, I've moved up the code block handling a `DELETE` request in the `uploaded_file` view endpoint, so that 
resources aren't wasted looking for the temp file by name twice.
Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/f0d9239495010ebc88c74d635f10cb42bb378271

yenaing-oo:
@danloveg 

Do you think it would be useful to include the UploadSession state diagram somewhere in the documentation for developers? If so, where would you say it is suitable to place it?

danloveg:
> @danloveg
> 
> Do you think it would be useful to include the UploadSession state diagram somewhere in the documentation for developers? If so, where would you say it is suitable to place it?

Good idea. If you put it here, then it will be included in the docs:

https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/572d503e0ba80ffb3fcefbec62b3e55624bc0828/bagitobjecttransfer/recordtransfer/models.py#L47

That would get rendered here: https://secure-record-transfer.readthedocs.io/en/latest/code/models.html#recordtransfer.models.UploadSession

I don't know if Sphinx can render Mermaid diagrams, you may need to use an image.

yenaing-oo:
Additional considerations:

I noticed that the `views.uploaded_file` endpoint performs `GET` and `DELETE` on **temporary uploaded files** only, whereas `views.list_uploaded_files` serves file metadata for **all uploaded files** (regardless of whether they are permanent or temporary). 

Should this be changed so that `views.list_uploaded_files` also only lists metadata for temp files only as well? Unless, we also want to support displaying of permanent uploaded files in the future (but perhaps this is for a later issue).

bagitobjecttransfer/recordtransfer/models.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/8ace49c7759f1c1f5fb089f9d652b71a54497865

danloveg:
> > @danloveg
> > Do you think it would be useful to include the UploadSession state diagram somewhere in the documentation for developers? If so, where would you say it is suitable to place it?
> 
> Good idea. If you put it here, then it will be included in the docs:
> 
> https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/572d503e0ba80ffb3fcefbec62b3e55624bc0828/bagitobjecttransfer/recordtransfer/models.py#L47
> 
> That would get rendered here: https://secure-record-transfer.readthedocs.io/en/latest/code/models.html#recordtransfer.models.UploadSession
> 
> I don't know if Sphinx can render Mermaid diagrams, you may need to use an image.

@yenaing-oo I wonder if it would be possible to render the diagram in text so that it can be viewed in the code and the docs the same way.

Something like this:

```
+---------+      +---------+
| CREATED | ---> | EXPIRED |
+---------+      +---------+
```

yenaing-oo:
> > > @danloveg
> @yenaing-oo I wonder if it would be possible to render the diagram in text so that it can be viewed in the code and the docs the same way.
> EDIT: I found this, which seems like a drop-in module that can render Mermaid: https://sphinxcontrib-mermaid-demo.readthedocs.io/en/latest/#

I think I got it working! (see https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/f95aafcf0f9ca9a40d82a7a3f97f9549fb151320 and below):

image

Although, it looks like this in code:
```python
    """Represents a file upload session. This model is used to track the files that a
    user uploads during a session.

    .. mermaid::
       :caption: UploadSession State Diagram

       flowchart TD
       CREATED
       EXPIRED
       UPLOADING
       COPYING_IN_PROGRESS
       COPYING_FAILED
       STORED
       REMOVING_IN_PROGRESS
       DELETED

       CREATED --> EXPIRED
       CREATED --> UPLOADING
       UPLOADING --> CREATED
       UPLOADING --> EXPIRED
       UPLOADING --> COPYING_IN_PROGRESS
       UPLOADING --> REMOVING_IN_PROGRESS
       COPYING_IN_PROGRESS --> COPYING_FAILED
       COPYING_IN_PROGRESS --> STORED
       STORED --> COPYING_IN_PROGRESS
       STORED --> REMOVING_IN_PROGRESS
       REMOVING_IN_PROGRESS --> DELETED
    """
```
Is that fine?

danloveg:
> Although, it looks like this in code:
> 
> ```python
>     """Represents a file upload session. This model is used to track the files that a
>     user uploads during a session.
> 
>     .. mermaid::
>        :caption: UploadSession State Diagram
> 
>        flowchart TD
>        CREATED
>        EXPIRED
>        UPLOADING
>        COPYING_IN_PROGRESS
>        COPYING_FAILED
>        STORED
>        REMOVING_IN_PROGRESS
>        DELETED
> 
>        CREATED --> EXPIRED
>        CREATED --> UPLOADING
>        UPLOADING --> CREATED
>        UPLOADING --> EXPIRED
>        UPLOADING --> COPYING_IN_PROGRESS
>        UPLOADING --> REMOVING_IN_PROGRESS
>        COPYING_IN_PROGRESS --> COPYING_FAILED
>        COPYING_IN_PROGRESS --> STORED
>        STORED --> COPYING_IN_PROGRESS
>        STORED --> REMOVING_IN_PROGRESS
>        REMOVING_IN_PROGRESS --> DELETED
>     """
> ```
> 
> Is that fine?

Looks good to me, though I think you can simplify the chart like this:

```
.. mermaid::
    :caption: UploadSession State Diagram

    flowchart TD
    CREATED --> EXPIRED
    CREATED --> TEMPORARY_FILES
    TEMPORARY_FILES --> EXPIRED
    TEMPORARY_FILES --> COPYING_IN_PROGRESS
    COPYING_IN_PROGRESS --> COPYING_FAILED
    COPYING_IN_PROGRESS --> STORED
    STORED --> COPYING_IN_PROGRESS
    STORED --> REMOVING_IN_PROGRESS
    REMOVING_IN_PROGRESS --> DELETED
```

Also, if you haven't already written one, that flowchart will need a bit of an introduction in the docstring.

yenaing-oo:
@danloveg 

> Looks good to me, though I think you can simplify the chart like this:

```
.. mermaid::

:caption: UploadSession State Diagram

    flowchart TD
    CREATED --> EXPIRED
    CREATED --> TEMPORARY_FILES
    TEMPORARY_FILES --> EXPIRED
    TEMPORARY_FILES --> COPYING_IN_PROGRESS
    COPYING_IN_PROGRESS --> COPYING_FAILED
    COPYING_IN_PROGRESS --> STORED
    STORED --> COPYING_IN_PROGRESS
    STORED --> REMOVING_IN_PROGRESS
    REMOVING_IN_PROGRESS --> DELETED

```

Did you use `TEMPORARY_FILES` by accident here?

Also, I see that `UPLOADING --> CREATED` and `UPLOADING --> REMOVING_IN_PROGRESS` are no longer included. Is this intentional?

danloveg:
> Did you use `TEMPORARY_FILES` by accident here?
> 
> Also, I see that `UPLOADING --> CREATED` and `UPLOADING --> REMOVING_IN_PROGRESS` are no longer included. Is this intentional?

Sorry, I used an outdated version of the flowchart. My point was the you can remove the definitions of each box before specifying the arrows between them.

i.e., these are equivalent:

```
.. mermaid::
    flowchart TD
    A
    B
    C

    A --> B
    B --> C
```

```
.. mermaid::
    flowchart TD
    A --> B
    B --> C
```

yenaing-oo:
> > Did you use `TEMPORARY_FILES` by accident here?
> > Also, I see that `UPLOADING --> CREATED` and `UPLOADING --> REMOVING_IN_PROGRESS` are no longer included. Is this intentional?
> 
> Sorry, I used an outdated version of the flowchart. My point was the you can remove the definitions of each box before specifying the arrows between them.
> 
> i.e., these are equivalent:
> 
> ```
> .. mermaid::
>     flowchart TD
>     A
>     B
>     C
> 
>     A --> B
>     B --> C
> ```
> 
> ```
> .. mermaid::
>     flowchart TD
>     A --> B
>     B --> C
> ```

Ah I see what you mean, thanks for clarifying.

yenaing-oo:
I've added an introduction and simplified the diagram in https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/da5c5d3dfbc3cf23ccc7b455dc6ef7d266a71463:

image

yenaing-oo:
@danloveg 

> This is working really well. Nothing is _broken_, but I do have a number of comments regarding code style, especially regarding error messages.
> 
> I stopped commenting about the use of """ partway through models.py but I still would like to see all the error messages using """ changed to ". Similarly, all cases where a SessionStatus is shown in the error message should use the `.name` property of the enum member instead of the value (which is what's used by default).
> 
> Finally, can you run this command to squash migration 0043 to 0049?
> 
> ```shell
> python manage.py squashmigrations --squashed-name split_file_into_two_models recordtransfer 0043_storeduploadedfile_tempuploadedfile_and_more 0049_alter_uploadsession_status
> ```
> 
> You can delete the old un-squashed migrations afterwards.

I forgot to do this at the end- I've done it now.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/f564001a7cb23c7523d2e017a3e16c798a51b46c and https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/d03e204d3dd20484552f98b17c98269d6b2da34a

yenaing-oo:
Also, in case my question got lost in all the other comments:


> Additional considerations:
> 
> I noticed that the `views.uploaded_file` endpoint performs `GET` and `DELETE` on **temporary uploaded files** only, whereas `views.list_uploaded_files` serves file metadata for **all uploaded files** (regardless of whether they are permanent or temporary).
> 
> Should this be changed so that `views.list_uploaded_files` also lists metadata for temp files only as well? Unless, we also want to support displaying of permanent uploaded files in the future (but perhaps this is for a later issue)

danloveg:
> Also, in case my question got lost in all the other comments:
> 
> > Additional considerations:
> > I noticed that the `views.uploaded_file` endpoint performs `GET` and `DELETE` on **temporary uploaded files** only, whereas `views.list_uploaded_files` serves file metadata for **all uploaded files** (regardless of whether they are permanent or temporary).
> > Should this be changed so that `views.list_uploaded_files` also lists metadata for temp files only as well? Unless, we also want to support displaying of permanent uploaded files in the future (but perhaps this is for a later issue)

Good catch, `uploaded_file` should also be able to handle GET requests for permanent files, yes. DELETE requests shouldn't be allowed if it's permanent, though.

Is this a quick change or does it deserve a different issue?

yenaing-oo:
> > Also, in case my question got lost in all the other comments:
> > > Additional considerations:
> > > I noticed that the `views.uploaded_file` endpoint performs `GET` and `DELETE` on **temporary uploaded files** only, whereas `views.list_uploaded_files` serves file metadata for **all uploaded files** (regardless of whether they are permanent or temporary).
> > > Should this be changed so that `views.list_uploaded_files` also lists metadata for temp files only as well? Unless, we also want to support displaying of permanent uploaded files in the future (but perhaps this is for a later issue)
> 
> Good catch, `uploaded_file` should also be able to handle GET requests for permanent files, yes. DELETE requests shouldn't be allowed if it's permanent, though.
> 
> Is this a quick change or does it deserve a different issue?

It'll require me to make another method , perhaps `get_perm_file_by_name` in `UploadSession`, and create additional tests for the model and the view. So this wouldn't be quick change, but it's not too lengthy either.

danloveg:
> > Good catch, `uploaded_file` should also be able to handle GET requests for permanent files, yes. DELETE requests shouldn't be allowed if it's permanent, though.
> > Is this a quick change or does it deserve a different issue?
> 
> It'll require me to make another method , perhaps `get_perm_file_by_name` in `UploadSession`, and create additional tests for the model and the view. So this wouldn't be quick change, but it's not too lengthy either.

Let's do that in a separate issue. I just opened up #326, we can do that work in there.

yenaing-oo:
I didn't know you made a separate issue for this: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/324
While I was adding the `sphinxcontrib-mermaid` dependency, I noticed it and addressed it at the same time: [Set selenium dependency to optional](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/f4846d2ad512d968632c75a4a80e8b530397b136)

danloveg:
> I didn't know you made a separate issue for this: #324 While I was adding the `sphinxcontrib-mermaid` dependency, I noticed it and addressed it at the same time: [Set selenium dependency to optional](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/318/commits/f4846d2ad512d968632c75a4a80e8b530397b136)

Awesome, addressing it here works, too.

bagitobjecttransfer/recordtransfer/models.py
```diff
@@ -39,138 +44,388 @@ def get_full_name(self):
...
+                f"Cannot remove uploaded files from session {self.token} while copying files is "
+                "in progress"
+            )
+
+        self.status = self.SessionStatus.REMOVING_IN_PROGRESS
+        self.save()
+
+        files = list(chain(self.permuploadedfile_set.all(), self.tempuploadedfile_set.all()))
+        for f in files:
+            f.remove()
```
danloveg, line 300: Calling `list()` here is redundant since `chain` already produces an iterator.

```suggestion
        for f in chain(self.permuploadedfile_set.all(), self.tempuploadedfile_set.all()):
            f.remove()
```

danloveg (changes requested):
I noticed one very small fix, but other than that I don't see anything else needing to be changed.

danloveg (approved):

PR #321, opened 2025-01-17
Associate UploadSession to User

yenaing-oo:


danloveg:
@yenaing-oo I'm going to hold off reviewing this until #318 is merged in. Reason being, the migration 0043 that was generated here would conflict with the migration 0043 generated in your other branch.

danloveg (changes requested):
`master` needs to be pulled into this branch and any conflicts there need to be resolved.

The migration `0043_uploadsession_user` needs to be changed to `0044_uploadsession_user`. You can change migrations by hand but I'd recommend removing the file and then asking Django to `makemigrations` again for that change.

yenaing-oo:
> `master` needs to be pulled into this branch and any conflicts there need to be resolved.
> 
> The migration `0043_uploadsession_user` needs to be changed to `0044_uploadsession_user`. You can change migrations by hand but I'd recommend removing the file and then asking Django to `makemigrations` again for that change.

I've pulled changes from `master` and remade the migration. This PR should be ready for a review now.

bagitobjecttransfer/recordtransfer/models.py
```diff
@@ -88,6 +88,7 @@ def __str__(self) -> str:
     status = models.CharField(
         max_length=2, choices=SessionStatus.choices, default=SessionStatus.CREATED
     )
+    user = models.ForeignKey(User, on_delete=models.SET_NULL, null=True)
 
     @classmethod
     def new_session(cls) -> UploadSession:
```
danloveg, line 94: Could you make it so that `new_session` can accept a user object when the session is created?

```suggestion
    def new_session(cls, user: Optional[User] = None) -> UploadSession:
```

danloveg (changes requested):
Looks good, this wasn't as big of a change as I was expecting it to be.

I think we can tighten up creating new sessions by allowing an optional User to be specified when the session is created.

Also, can you add a link to the User object in the admin here?

Screenshot 2025-01-23 131700

Let me know if you need help with that, but you can probably refer to the Submission admin for how to do that, since it includes the User in the list in the admin too.

bagitobjecttransfer/recordtransfer/models.py
yenaing-oo: I like this idea - it's certainly neater than setting the user after the session is created.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/321/commits/54c10848d8030c940eddf4567d37a6de2a10d35e

yenaing-oo:
> Also, can you add a link to the User object in the admin here?
> 
> Screenshot 2025-01-23 131700
> 
> Let me know if you need help with that, but you can probably refer to the Submission admin for how to do that, since it includes the User in the list in the admin too.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/321/commits/0fc5f70f6ab375544858b97894245d22b1ea3e88"

Also, I was playing around with the *_Admin classes and removing the form for an admin class that is read-only does not seem to affect anything. I don't think including the a form, such as `UploadSessionForm` on the `UploadSessionAdmin` class in necessary since it's no editing capabilities are enabled.

danloveg:
> Also, I was playing around with the *_Admin classes and removing the form for an admin class that is read-only does not seem to affect anything. I don't think including a form, such as [`UploadSessionForm`](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/0fc5f70f6ab375544858b97894245d22b1ea3e88/bagitobjecttransfer/recordtransfer/admin.py#L229) on the `UploadSessionAdmin` class, is necessary if no editing capabilities are enabled for that model.

I'm all for removing code when possible. Feel free to open up an issue for that (tag it with the cleanup tag), that sounds like a good change to have.

danloveg (approved):
Looks great, really glad to have this protection in now to prevent users from seeing/deleting each others files.

PR #327, opened 2025-01-23
Improve navbar burger menu close button

yenaing-oo:
I didn't want to spend too much time on CSS for making a fancy animation for the burger menu button -> close button, so I've opted for a simple icon change instead. I think the positioning of the navbar container has been improved though.


danloveg (approved):
Looks great!

Issue #333, opened 2025-01-27
Past Transfers in this Group Table is Empty


danloveg:
The table on a submission group page for "Past Transfers in this Group" is empty, regardless of how many submissions are actually part of that submission.

![Image](https://github.com/user-attachments/assets/d8092219-7ece-492a-9ca0-2f5f0670efe4)

This is most likely an issue with the submissions not being passed correctly to the template.

PR #335, opened 2025-01-27
Fix submissions not showing for a given group in Group Details page

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/333

### DIFF
--- a/bagitobjecttransfer/recordtransfer/templates/includes/submission_table.html
+++ b/bagitobjecttransfer/recordtransfer/templates/includes/submission_table.html
@@ -9,7 +9,7 @@
             <th>{% trans "View Report" %}</th>
             <th>{% trans "Download CSV" %}</th>
         </tr>
-        {% for submission in submissions_page_obj %}
+        {% for submission in submissions %}
             <tr>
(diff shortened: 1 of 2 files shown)


danloveg (changes requested):
The profile page isn't showing submissions now. I think it's because the queryset is called `submissions_page_obj` on the profile page, and then it's called `submissions` on the group page.

Screenshot 2025-01-27 154929

yenaing-oo:
> The profile page isn't showing submissions now. I think it's because the queryset is called `submissions_page_obj` on the profile page, and then it's called `submissions` on the group page.
> 

Thanks for catching that. I think this is a good lesson to explicitly pass objects to includes by name like this:
`include "includes/submission_table.html" with submissions=submissions_page_obj` to prevent issues with change in variable naming.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/335/commits/92efb8afb03269594eb9a701116ce582a9d02d92

danloveg (approved):
Perfect!

Issue #328, opened 2025-01-23
Check if admin forms are needed for admin classes of read-only models


yenaing-oo:
It looks like removing the form from admin classes for models that are read-only doesn’t change anything. For example, the [`UploadSessionForm`](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/0fc5f70f6ab375544858b97894245d22b1ea3e88/bagitobjecttransfer/recordtransfer/admin.py#L229) might not be needed in `UploadSessionAdmin` since editing isn’t allowed.

To do:
- Check if admin forms are actually needed for read-only models.
- If they aren't, remove them to make code cleaner

PR #337, opened 2025-01-27
Remove redundant forms from recordtransfer admin

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/328

Change notes:
- Removed `admin/readonly_change_form.html` as it was only being used to hide the "Close" button on several read-only `ModelAdmin` classes 
- Added a utility function to format file sizes of `BaseUploadedFile` instances and display them in `ModelAdmin` and `TabularInline` classes
- Created a `ReadOnlyInline` class for read-only `TabularInline`s, similar to `ReadOnlyAdmin`, and made use of it in most inline admin classes (overriding permissions where necessary)
- Removed usage of forms to control display of fields in read-only inline and model admin classes, defining the fields to display directly in the admin class instead


bagitobjecttransfer/recordtransfer/admin.py
```diff
@@ -113,16 +118,15 @@ class TempUploadedFileAdmin(ReadOnlyAdmin):
     class Media:
         js = ("admin_uploadedfile.bundle.js",)
 
-    change_form_template = "admin/readonly_change_form.html"
-
-    form = UploadedFileForm
-
     actions = [
         "clean_temp_files",
     ]
```
danloveg, line 123: This action can be removed - it should be the UploadSession that manages files. We also don't want to be able to delete files from the backend like this.

bagitobjecttransfer/recordtransfer/admin.py
```diff
@@ -113,16 +118,15 @@ class TempUploadedFileAdmin(ReadOnlyAdmin):
...
 
+    fields = ["id", "name", format_upload_size, "exists", linkify("session"), "file_upload"]
+
```
danloveg, line 126: Since we're looking at these admins now, let's add some [search_fields](https://docs.djangoproject.com/en/4.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.search_fields) to make these files searchable.

I think these would be good:

```python
search_fields = [
    "name",
    "session__token",
    "session__user__username",
]
```

bagitobjecttransfer/recordtransfer/admin.py
```diff
@@ -102,7 +107,7 @@ def has_delete_permission(self, request, obj=None):
 
 @admin.register(TempUploadedFile)
 class TempUploadedFileAdmin(ReadOnlyAdmin):
```
danloveg, line 109: You can reuse this admin for both models:

```suggestion
@admin.register(TempUploadedFile)
@admin.register(PermUploadedFile)
class UploadedFileAdmin(ReadOnlyAdmin):
```

bagitobjecttransfer/recordtransfer/admin.py
```diff
@@ -113,16 +118,15 @@ class TempUploadedFileAdmin(ReadOnlyAdmin):
...
     ]
 
+    fields = ["id", "name", format_upload_size, "exists", linkify("session"), "file_upload"]
+
     list_display = [
         "name",
+        format_upload_size,
         "exists",
         linkify("session"),
     ]
```
danloveg, line 132: I can't comment on changes that aren't in this pull request, but I think the ordering should be `["-pk"]` instead. That way, files that have been most recently uploaded get uploaded first.

bagitobjecttransfer/recordtransfer/admin.py
```diff
@@ -151,31 +155,27 @@ class PermUploadedFileAdmin(ReadOnlyAdmin):
...
 
     ordering = ["-session", "name"]
 
 
-class TempUploadedFileInline(admin.TabularInline):
-    """Inline admin for the BaseUploadedFile model. Used to view the files
-    associated with an upload session.
+class ReadOnlyInline(admin.TabularInline):
```
danloveg, line 170: Can you move this class to below `ReadOnlyAdmin`?

bagitobjecttransfer/recordtransfer/admin.py
```diff
@@ -224,23 +229,24 @@ class UploadSessionAdmin(ReadOnlyAdmin):
         - delete: Not allowed
     """
 
-    change_form_template = "admin/readonly_change_form.html"
+    @display(description=gettext("Upload Size"))
+    def upload_size(self, obj):
+        return get_human_readable_size(obj.upload_size, 1000, 2)
 
-    form = UploadSessionForm
+    fields = ["token", linkify("user"), "started_at", "file_count", "upload_size", "status"]
```
danloveg, line 236: Like I recommended for the uploaded file admin, some [search_fields](https://docs.djangoproject.com/en/4.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.search_fields) would be good here to be able to search upload sessions.

How about:

```python
search_fields = ["token", "user__username"]
```

danloveg (changes requested):
I like how this is shaping up. I have a few comments about organizing the code and making the admin pages for the uploaded files/upload session more interactive.

bagitobjecttransfer/recordtransfer/admin.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/337/commits/0d80505c9396e77c4b45255bcd0c232d8a867aea

bagitobjecttransfer/recordtransfer/admin.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/337/commits/c23f7fe83774ec5e26e14a8d2dd275f751244791

bagitobjecttransfer/recordtransfer/admin.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/337/commits/caa772330e003fedfec045e14d9ed554890e5ca8

bagitobjecttransfer/recordtransfer/admin.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/337/commits/f0c0b27e5e9de361d5ff7d1dd0114b23c0f2cbc3

bagitobjecttransfer/recordtransfer/admin.py
yenaing-oo: Just a slight syntax adjustment since stacking decorators like that does not work.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/337/commits/deccec1809d2bcebd9329ef47b60670dea794fac

danloveg (approved):
I love when functionality is added and more lines of code were removed than added. Looks great!

PR #339, opened 2025-01-29
Make File Transfer REST-ful

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/32

### DIFF
--- a/bagitobjecttransfer/recordtransfer/tests/unit/test_views.py
+++ b/bagitobjecttransfer/recordtransfer/tests/unit/test_views.py
@@ -1,6 +1,6 @@
 import logging
 from unittest import skipIf
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
 
 from django.contrib.messages import get_messages
 from django.core.files.uploadedfile import SimpleUploadedFile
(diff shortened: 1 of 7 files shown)


yenaing-oo:
Change notes:
- The endpoint to upload a file is now being built dynamically on the FE, based on the current upload session token
- The views `upload_file` and `list_uploaded_files` are combined into one: `upload_or_list_uploaded_files` to cater to both POST and GET requests to the same endpoint
- The test class for the views (`upload_file` and `list_uploaded_files`) have been combined, and a new test class for testing the`create_upload_session` view has been added

bagitobjecttransfer/recordtransfer/tests/unit/test_views.py
```diff
@@ -58,98 +58,134 @@ def setUp(self) -> None:
...
+        file_to_upload = SimpleUploadedFile("testfile.txt", self.one_kib)
+        temp_file = self.session.add_temp_file(file_to_upload)
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 200)
+        response_json = response.json()
+        responseFiles = response_json.get("files")
+        self.assertEqual(len(responseFiles), 1)
+        self.assertEqual(responseFiles[0]["name"], "testfile.txt")
+        self.assertEqual(responseFiles[0]["size"], file_to_upload.size)
+        self.assertEqual(responseFiles[0]["url"], temp_file.get_file_access_url())
```
danloveg, line 119: snake_case!

```suggestion
        response_files = response_json.get("files")
        self.assertEqual(len(response_files), 1)
        self.assertEqual(response_files[0]["name"], "testfile.txt")
        self.assertEqual(response_files[0]["size"], file_to_upload.size)
        self.assertEqual(response_files[0]["url"], temp_file.get_file_access_url())
```

bagitobjecttransfer/recordtransfer/views.py
```diff
@@ -953,33 +953,71 @@ def get_in_progress_submission(user: User, uuid: str) -> Optional[InProgressSubm
...
+    except Exception as exc:
+        LOGGER.error("Error creating upload session: %s", str(exc), exc_info=exc)
+        return JsonResponse(
+            {"error": gettext("There was an internal server error. Please try again.")},
+            status=500,
+        )
+
+
+@require_http_methods(["GET", "POST"])
+def upload_or_list_files(request: HttpRequest, session_token: str) -> JsonResponse:
```
danloveg, line 978: This might sound a bit backwards, but I think this would be a bit easier to read if the `_list_uploaded_files` and `_upload_file` functions were removed and incorporated into this function.

I like the docstring for the POST functionality, if you could combine that docstring into this function's docstring I think that'd be great.

danloveg (changes requested):
The endpoints are so much easier to understand now. This setup will be great for if/when we enable looking at file uploads post-submission.

I made a suggestion to make one big function for listing files and uploading a file - I know a lot of common advice is to make small functions (which is sound advice to follow usually) but I think in this case it'd actually be easier to read the view code if it was all in one function.

bagitobjecttransfer/recordtransfer/views.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/339/commits/f22f23d12aff7321c0d649d2ccecc630351a4b96

bagitobjecttransfer/recordtransfer/tests/unit/test_views.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/339/commits/d91f206b09d09a1de568bbb2a52d264a0306f8c0

bagitobjecttransfer/recordtransfer/views.py
```diff
@@ -953,97 +953,129 @@ def get_in_progress_submission(user: User, uuid: str) -> Optional[InProgressSubm
...
+    """Upload a single file to the server list the files uploaded in a given upload session. The
+    file is added to the upload session using the session token passed as a parameter in the
+    request. If an session token is invalid, an error message is returned.
```
danloveg, line 981: One tiny change needed in this docstring:

```suggestion
    request. If a session token is invalid, an error message is returned.
```

danloveg (changes requested):

danloveg (approved):

PR #340, opened 2025-01-30
Switch to mysqlclient for MySQL Connections

danloveg:
This pull request switches out the `mysql-connector-python` dependency for Django's recommended `mysqlclient`.

Since `mysqlclient` has some C dependencies that need to be built, I've modified how the container is built.

The Containerfile now uses a multi-stage build. The `base` stage is shared between the new `dev` and `prod` targets. A way to visualize the Containerfile is this:

```mermaid
---
title: Target Flowchart for new Containerfile
---

flowchart TD
    base("`base
    (based on *nikolaik/python-nodejs:python3.10-nodejs22-slim*)
    `")

    dev("`**dev**
    (based on *nikolaik/python-nodejs:python3.10-nodejs22-slim*)
    `")

    builder("`builder
    (based on *nikolaik/python-nodejs:python3.10-nodejs22-slim*)
    `")

    prod("`**prod**
    (based on *python:3.10-slim*)
    `")

    base --> dev
    base --> builder
    builder --> prod
```

I've switched the `prod` image to use `python:3.10-slim` which has better support than `nodejs:python3.10-nodejs22-slim`. This means that `node` and `npm` are no longer available in the production image. This is fine because the assets are built in the `base` image and do not need to rebuilt after that for prod.

Another notable change is that I'm now using a virtual environment in the container. Since I needed to be able to copy Python dependencies between different containers, using a virtual environment was the obvious solution.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,34 +1,61 @@
-FROM nikolaik/python-nodejs:python3.10-nodejs22-slim
+FROM nikolaik/python-nodejs:python3.10-nodejs22-slim AS builder
 
 ENV PYTHONUNBUFFERED=1
+
 # Make arg passed from compose files into environment variable
 ARG WEBPACK_MODE
 ENV WEBPACK_MODE $WEBPACK_MODE
 
-# Install poetry
-RUN python -m pip install --user pipx && \
-    python -m pipx ensurepath && \
-    python -m pipx install poetry
-
 WORKDIR /app/
+
+# Install build dependencies and poetry
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        build-essential \
+        default-libmysqlclient-dev \
+        pkg-config \
+    && \
+    curl -sSL https://install.python-poetry.org | python3 - --version 1.8.5
 
 # Copy poetry-related files, and install Python dependencies
 COPY pyproject.toml poetry.lock README.md /app/
-RUN poetry config virtualenvs.create false && poetry install
+RUN poetry config virtualenvs.create true && \
+    poetry config virtualenvs.in-project true && \
+    poetry install -E prod
 
 # Copy Node-related files, and install NodeJS dependencies
 COPY package*.json webpack.config.js /app/
 RUN npm install --no-color
-
-# Copy entrypoint script to image
-COPY ./docker/entrypoint.sh /app/
-RUN chmod +x /app/entrypoint.sh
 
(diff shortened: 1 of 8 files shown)


yenaing-oo:
@danloveg 
Did you want me to review this PR by any chance? I didn't see a request for one and it looks like it is ready for review.

danloveg:
> @danloveg Did you want me to review this PR by any chance? I didn't see a request for one and it looks like it is ready for review.

It's ready to go! I hadn't opened up a request yet because I wanted to focus on the January 2025 milestone.

Containerfile
```diff
@@ -1,35 +1,90 @@
...
+################################################################################
+#
+# PRODUCTION IMAGE
+#
+
+FROM base AS builder
+
+# Install build tools for production dependencies
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
```
yenaing-oo, line 61: Looking at `mysqlclient`'s Installation Instructions for Linux, I see that `python3-dev` is included as one of the build dependencies there. Everything seems to work fine without it - I was just wondering why that is.

Containerfile
```diff
@@ -1,35 +1,90 @@
...
-
-# Install poetry
-RUN python -m pip install --user pipx && \
-    python -m pipx ensurepath && \
-    python -m pipx install poetry
 
 WORKDIR /app/
 
+# Install poetry
+RUN curl -sSL https://install.python-poetry.org | python3 - --version 1.8.5
```
yenaing-oo, line 8: Could you explain why you're installing poetry directly using `curl`, in favor of using `pipx`?

yenaing-oo (commented):
I have no feedback since everything looks good and works fine, but I do have some questions for educational purposes.

Containerfile
danloveg: I didn't realize this when first setting up Poetry, but `curl` is the [official method of installing Poetry](https://python-poetry.org/docs/#installing-with-the-official-installer). I also notice that it installs much faster than using `pipx` so I decided to switch to this method.

Containerfile
yenaing-oo: I see, no worries. It appears to work perfectly fine on the production container so this isn't an issue.

yenaing-oo (approved):
Everything looks good to me, thanks for the clarifications on my questions!

Issue #343, opened 2025-01-30
Automate Static Asset Build for E2E Tests


yenaing-oo:
E2E tests depend on static assets being built first, requiring `npm run build` to be executed beforehand. This process needs to be automated somehow.

PR #344, opened 2025-01-30
Automate Static Asset Build for E2E Tests

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/343

Uses a pytest hook to call `npm run build` before any E2E test is run:
- https://docs.pytest.org/en/stable/how-to/writing_plugins.html#making-your-plugin-installable-by-others
- https://docs.pytest.org/en/stable/how-to/writing_hook_functions.html

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,21 @@
+"""Contains pytest hooks. The hooks defined here are used to modify the behavior of
+pytest during test collection and execution.
+"""
+
+import subprocess
+from pathlib import Path
+
+import pytest
+
+
+def pytest_collection_modifyitems(session, config, items) -> None:
+    """Run npm build only if e2e tests are being run. `items` is a list of test items to be
+    run.
+    """
+    e2e_tests = any("e2e" in str(Path(item.fspath)) for item in items)
+
+    if e2e_tests:
+        try:
+            subprocess.run(["npm", "run", "build"], check=True)
+        except subprocess.CalledProcessError as e:
+            pytest.exit(f"Failed to run npm build: {e}")
--- a/conftest.py
+++ b/conftest.py
@@ -12,9 +12,8 @@ def pytest_collection_modifyitems(session, config, items) -> None:
     """Run npm build only if e2e tests are being run. `items` is a list of test items to be
     run.
     """
-    e2e_tests = any("e2e" in str(Path(item.fspath)) for item in items)
-
-    if e2e_tests:
+    # Check if running unit tests via -k parameter
+    if config.getoption("-k") == "e2e":
         try:
             subprocess.run(["npm", "run", "build"], check=True)
         except subprocess.CalledProcessError as e:
--- a/conftest.py
+++ b/conftest.py
@@ -14,6 +14,19 @@ def pytest_collection_modifyitems(session, config, items) -> None:
     # Check if running unit tests via -e2e parameter
     if config.getoption("-k") == "e2e":
         try:
-            subprocess.run(["npm", "run", "build"], check=True)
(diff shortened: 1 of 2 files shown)


conftest.py
```diff
@@ -0,0 +1,19 @@
...
+def pytest_collection_modifyitems(session, config, items) -> None:
+    """Run npm build only if e2e tests are being run. `items` is a list of test items to be
+    run.
+    """
+    # Check if running unit tests via -e2e parameter
+    if config.getoption("-k") == "e2e":
+        try:
+            subprocess.run(["npm", "run", "build"], check=True)
+        except subprocess.CalledProcessError as e:
+            pytest.exit(f"Failed to run npm build: {e}")
```
danloveg, line 19: This is failing to run on my Windows machine with a `FileNotFoundError`. I found out it's because the npm command should be `npm.cmd` in Windows.

I recommend using `npm.cmd` in Windows, and being explicit with what mode Webpack is to be run in.

(These changes require an `import os` at the top of the file)

```suggestion
    # Check if running unit tests via -e2e parameter
    if config.getoption("-k") == "e2e":
        try:
            npm_cmd = "npm.cmd" if os.name == "nt" else "npm"
            subprocess.run(
                [npm_cmd, "run", "build"],
                check=True,
                env={
                    **os.environ,
                    "WEBPACK_MODE": "development",
                },
            )
        except subprocess.CalledProcessError as e:
            pytest.exit(f"Failed to run npm build: {e}")
        except FileNotFoundError as e:
            pytest.exit(
                f"npm command not found (got: {e}). npm must be installed to run e2e tests."
            )

```

pyproject.toml
```diff
@@ -42,3 +42,6 @@ build-backend = "poetry.core.masonry.api"
 pythonpath = "bagitobjecttransfer"
 python_files = "test_*.py"
 DJANGO_SETTINGS_MODULE = "bagitobjecttransfer.settings.test"
+
+[tool.poetry.plugins."pytest11"]
```
danloveg, line 46: The `conftest` runs fine on my machine without the quotes around `pytest11`. Can you double check whether they're required on your machine too?

```suggestion
[tool.poetry.plugins.pytest11]
```

danloveg (changes requested):

conftest.py
yenaing-oo: I see, I didn't know that `npm.cmd` is actually the command that gets called for npm on Windows. It turns out `npm` does not work on Windows using `subprocess.run` because it does not use shell resolution by default. 

Would it be better to use `shutil.which("npm")` instead to look for the path of the `npm` executable on any given platform?

Addressed by:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/344/commits/39739825fb63b7ba26674bcfb21814439e744482

pyproject.toml
yenaing-oo: Addressed by:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/344/commits/c682701d7aaee98bd437da98554a8b09ccc381a7

yenaing-oo:
I also added this to the if-condition, so that assets are also built if no option "-k" is specified, which means all tests (inlcuding E2E) will run.

https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/344/commits/1499e77afaaef24f52a8af887810993d74b5434b

danloveg (approved):
I like how you switched it to use `shutil.which`, that makes more sense. Looks good!

Issue #319, opened 2025-01-17
Rename bagitobjecttransfer


danloveg:
I am proposing we rename the Django project `bagitobjecttransfer` to `app`. The name `bagitobjecttransfer` is a bit misleading - it seems to indicate that people transfer objects in BagIt bags using this app. For one, objects do not need to be in BagIt bags when uploaded. And another thing, file transfers can be turned off and the app can run in a mode that doesn't even accept objects.

Naming it `app` (or something equally generic) is a more accurate representation of what's in that directory. `core` might also be a good option.

If I was coming across this repository for the first time, having a directory named `app` would make it crystal clear that that's where the application code lives.

---

This question has some answers on how to actually do the rename: https://stackoverflow.com/questions/18293875/is-there-an-easy-way-to-rename-a-django-project

yenaing-oo:
Do you mind clarifying what you expect the project structure to look like after this change?

I'm looking at what a standard Django project's hierarchy looks like for comparison:

```
project_name/
├── manage.py
├── project_name/
│   ├── __init__.py
│   ├── asgi.py
│   ├─ settings.py
│   ├─ urls.py
│   ├── wsgi.py
├── app1/
├── app2/
  ...
├── static/
├── media/
├── templates/
```

Were you thinking of:
1) Moving the contents of the outer `bagitobjecttransfer` out removing that directory  AND
2) Renaming the one `bagitobjecttransfer` directory that's left (the one that contains the project settings) to `app`?

danloveg:
> Do you mind clarifying what you expect the project structure to look like after this change?
> 
> I'm looking at what a standard Django project's hierarchy looks like for comparison:
> 
> ```
> project_name/
> ├── manage.py
> ├── project_name/
> │   ├── __init__.py
> │   ├── asgi.py
> │   ├─ settings.py
> │   ├─ urls.py
> │   ├── wsgi.py
> ├── app1/
> ├── app2/
>   ...
> ├── static/
> ├── media/
> ├── templates/
> ```
> 
> Were you thinking of:
> 
>     1. Moving the contents of the outer `bagitobjecttransfer` out removing that directory  AND
> 
>     2. Renaming the one `bagitobjecttransfer` directory that's left (the one that contains the project settings) to `app`?

The structure should remain as it is, but every instance of `bagitobjecttransfer` should be replaced with `app`.

PR #345, opened 2025-01-30
Split Views Into Separate Modules

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/311

Additional changes:
- Updates the docs to separate each new view module into its own section (https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/345/commits/168330268b672f1e1518b53011eeb163919f3715)
- Splits tests for views into separate test files by module

PR #346, opened 2025-01-31
Rename bagitobjecttransfer

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/319


yenaing-oo:
@danloveg

After replacing every instance of `bagitobjecttransfer` with `app`, this is the project structure we get:

image

I think the top level directory `app` is redundant since it acts as a container folder for another folder of the same name and purpose. Is there perhaps a reason why we might want to keep it as is?

It looks like the app runs without any issues after this change.

danloveg:
> @danloveg
> 
> After replacing every instance of `bagitobjecttransfer` with `app`, this is the project structure we get:
> 
> image
> 
> I think the top level directory `app` is redundant since it acts as a container folder for another folder of the same name and purpose. Is there perhaps a reason why we might want to keep it as is?
> 
> It looks like the app runs without any issues after this change.

Awesome! The top-level `/app/` folder is still needed since we don't want to copy all the files into the root `/` path of the container. I think it would make sense to copy the app to `/opt/secure-record-transfer/` when the container is built.

This line:

https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/1bfbb39bcd08ce1ad9a195f4118a41d332bedc7b/Containerfile#L13

Should be `WORKDIR /opt/secure-record-transfer/`

This line:

https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/1bfbb39bcd08ce1ad9a195f4118a41d332bedc7b/Containerfile#L33

Should be `WORKDIR /opt/secure-record-transfer/app/`

yenaing-oo:
I think it would make sense to hold this off until https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/345 is merged in

yenaing-oo:
This PR will no longer be used to address https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/319, as it's branch has fallen too far behind `master` and will be difficult to resolve conflicts.

yenaing-oo:
- `PROJ_DIR="/opt/secure-record-transfer/"` is now being used in place of `/app/` to store built static assets, Python virtual environment, node modules, webpack configuration, etc.
- ENV APP_DIR="/opt/secure-record-transfer/app/" is now being used in place of `/app/bagitobjecttransfer` to hold the app files

Containerfile
```diff
@@ -40,14 +42,16 @@ RUN chmod +x /app/entrypoint.sh
 FROM base as dev
 
 ENV PYTHONUNBUFFERED=1
+ENV PROJ_DIR="/opt/secure-record-transfer/"
+ENV APP_DIR="/opt/secure-record-transfer/app/"
```
danloveg, line 46: Something I learned this morning is that an image that extends another one in Docker inherits the environment variables. So these three lines aren't necessary.

See here, I removed these three lines from the `Containerfile` on my machine and they are still set correctly in the `dev` image:

Screenshot 2025-02-07 091131

The `ENV`s in the prod image are still required though since that's a completely new image, it doesn't inherit from the `base`

app/app/settings/docker_prod.py
danloveg, line 1: The media and static files aren't working in the prod configuration, because they're still set to be relative to `/app/`. Instead of setting the explicit absolute directory to the MEDIA_ROOT and STATIC_ROOT, would this work?

```python
# file: docker_prod.py

MEDIA_ROOT = os.path.join(os.path.dirname(BASE_DIR), "media")
STATIC_ROOT = os.path.join(os.path.dirname(BASE_DIR), "static")
```

In the prod configuration, the media and static files are located relative to one directory *above* the `WORKDIR`, but in the dev configuration, the media and static files are located relative to the `WORKDIR`. I think it makes sense to set the MEDIA_ROOT and STATIC_ROOT like my snippet above but let me know if there are any drawbacks to that method (or if it would even work).

danloveg (changes requested):

Containerfile
yenaing-oo: That's good to know. I've addressed that here:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/346/commits/ef9c454bde5178556a5716b253838dc17c6fa115

app/app/settings/base.py
```diff
@@ -256,3 +256,15 @@
...
 )
+
+# Media file storage locations
+BAG_STORAGE_FOLDER = config(
+    "BAG_STORAGE_FOLDER", default=os.path.join(MEDIA_ROOT, "bags")
+)
+UPLOAD_STORAGE_FOLDER = config(
+    "UPLOAD_STORAGE_FOLDER", default=os.path.join(MEDIA_ROOT, "uploaded_files")
+)
+print("UPLOAD_STORAGE_FOLDER: ", UPLOAD_STORAGE_FOLDER)
```
danloveg, line 267: This random `print` message can probably go.

docs/settings/index.rst
```diff
@@ -410,11 +410,11 @@ BAG_STORAGE_FOLDER
...
-        ====================================  ================  ======
-        Default in Dev                        Default in Prod   Type
-        ====================================  ================  ======
-        /app/bagitobjecttransfer/media/bags/  /app/media/bags/  string
-        ====================================  ================  ======
+        ===========================================  ================  ======
+        Default in Dev                               Default in Prod   Type
+        ===========================================  ================  ======
+        /opt/secure-record-transfer/app/media/bags/  /app/media/bags/  string
+        ===========================================  ================  ======
```
danloveg, line 417: This and other mentions of `/app/media/` should be change to the correct prod default.

danloveg (changes requested):
In addition to the changes I highlighted, the `example.prod.env` needs to be updated. The `BAG_STORAGE_FOLDER` and `UPLOAD_STORAGE_FOLDER` should both be removed from that file.

yenaing-oo:
> In addition to the changes I highlighted, the `example.prod.env` needs to be updated. The `BAG_STORAGE_FOLDER` and `UPLOAD_STORAGE_FOLDER` should both be removed from that file.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/346/commits/ffd12c5535d0426733d502f6177435b385ea45de

docs/settings/index.rst
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/346/commits/ce77a602dcf034e0c638d3c06a2f76f5134db63e

compose.prod.yml
```diff
@@ -80,14 +80,14 @@ services:
...
+      - DJANGO_SETTINGS_MODULE=app.settings.docker_prod
     depends_on:
       - db
       - rq
       - clamav
     volumes:
-      - static-volume:/app/static/
-      - media-volume:/app/media/
+      - static-volume:/opt/secure-record-transfer/static/
+      - media-volume:/opt/secure-record-transfer/media/
```
danloveg, line 90: This directory doesn't have the static or media files anymore, it should be changed to:

```suggestion
      - static-volume:/opt/secure-record-transfer/app/static/
      - media-volume:/opt/secure-record-transfer/app/media/
```

You can leave the path in the NGINX container as-is, since it's fine for this volume to be mounted at a different path in that container. But this container does need the proper path since it has to be the path to the `STATIC_ROOT`.

docs/settings/index.rst
```diff
@@ -433,11 +433,11 @@ UPLOAD_STORAGE_FOLDER
...
 
-        ==============================================  ==========================  ======
-        Default in Dev                                  Default in Prod             Type
-        ==============================================  ==========================  ======
-        /app/bagitobjecttransfer/media/uploaded_files/  /app/media/uploaded_files/  string
-        ==============================================  ==========================  ======
+        ======================================================  ======================================================  ======
+        Default in Dev                                          Default in Prod                                         Type
+        ======================================================  ======================================================  ======
+        /opt/secure-record-transfer/app//media/uploaded_files/  /opt/secure-record-transfer/app//media/uploaded_files/  string
```
danloveg, line 439: Double forward slashes here:

```suggestion
        /opt/secure-record-transfer/app/media/uploaded_files/  /opt/secure-record-transfer/app/media/uploaded_files/  string
```

danloveg (changes requested):
There's more consistency between the dev and prod containers now regarding static and media files, which is good.

I did notice an issue with the volume being set incorrectly in the prod compose file. Once that's fixed up that should be the last thing.

compose.prod.yml
yenaing-oo: Could you explain why it's fine for the static and media volumes to be mounted at a different path in the `nginx` container a bit further? Doesn't `nginx` need access to the files in `/opt/secure-record-transfer/app/static/` and `/opt/secure-record-transfer/app/media/` to be able to serve them?

yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/346/commits/a40bac9188e8a68d670122a302e4d1ccdc5d0887

danloveg: > Could you explain why it's fine for the static and media volumes to be mounted at a different path in the `nginx` container a bit further? Doesn't `nginx` need access to the files in `/opt/secure-record-transfer/app/static/` and `/opt/secure-record-transfer/app/media/` to be able to serve them?

The filesystems for the `nginx` and `app` containers are completely separate, the only thing similar between them is the `static-volume` and `media-volume`. When `collecstatic` is run, the static files are all copied to the path `/opt/secure-record-transfer/app/static/`, which is in the `static-volume` volume. That volume could be mounted anywhere in the `nginx` container, so long as it knows where to find them.

So for example, this (partial) setup would be valid:

```yml
app:
  volumes:
    - static-volume:/opt/secure-record-transfer/app/static/
    - media-volume:/opt/secure-record-transfer/app/media/

nginx:
  volumes:
    - static-volume:/usr/share/nginx/html/static/
    - media-volume:/usr/share/nginx/html/media/
  environment:
    - STATIC_ROOT=/usr/share/nginx/html/static/
    - MEDIA_ROOT=/usr/share/nginx/html/media/
```

The static files are shared between the containers, but the path to them doesn't matter so much in the NGINX container.

yenaing-oo: Ah this makes sense. I forgot about how shared volumes between containers worked. Thanks for clarifying!

danloveg (approved):

PR #347, opened 2025-01-31
Write Tests for SubmissionGroupForm, SubmissionGroupCreateView and SubmissionGroupDetailView

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/239

### DIFF
--- a/bagitobjecttransfer/recordtransfer/tests/unit/test_forms.py
+++ b/bagitobjecttransfer/recordtransfer/tests/unit/test_forms.py
@@ -452,46 +452,45 @@ class SubmissionGroupFormTest(TestCase):
         form = SubmissionGroupForm(data=form_data, user=self.user)
         assert form.is_valid()
         group = form.save()
-        assert group.name == "New Group"
-        assert group.description == "A new submission group"
-        assert group.created_by == self.user
+        self.assertEqual(group.name, "New Group")
+        self.assertEqual(group.description, "A new submission group")
+        self.assertEqual(group.created_by, self.user)
 
     def test_form_empty_data(self) -> None:
         """Case where the form is empty."""
         form = SubmissionGroupForm(data={}, user=self.user)
-        assert not form.is_valid()
+        self.assertFalse(form.is_valid())
 
     def test_form_missing_name(self) -> None:
         """Case where the name is missing."""
         form_data = {"description": "A new submission group"}
         form = SubmissionGroupForm(data=form_data, user=self.user)
-        assert not form.is_valid()
-        assert "name" in form.errors
+        self.assertFalse(form.is_valid())
+        self.assertIn("name", form.errors)
 
     def test_form_duplicate_name(self) -> None:
         """Case where the name is a duplicate of an existing group name."""
         form_data = {"name": "Existing Group", "description": "A duplicate submission group"}
         form = SubmissionGroupForm(data=form_data, user=self.user)
-        assert not form.is_valid()
-        assert "name" in form.errors
+        self.assertFalse(form.is_valid())
+        self.assertIn("name", form.errors)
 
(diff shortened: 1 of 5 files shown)


bagitobjecttransfer/recordtransfer/tests/unit/test_forms.py
```diff
@@ -433,3 +434,79 @@ def test_form_no_files_uploaded(self) -> None:
...
+        form = SubmissionGroupForm(data=form_data, user=self.user)
+        assert not form.is_valid()
+        assert "name" in form.errors
+
+    def test_form_no_changes(self) -> None:
+        """Case where no changes are detected."""
+        form_data = {"name": "Existing Group", "description": "A new description"}
+        form = SubmissionGroupForm(data=form_data, user=self.user)
+        form.is_valid()  # Trigger the clean method
+        assert not form.is_valid()
```
danloveg, line 483: I believe the first `is_valid()` is redundant. If it's not, can you add a slightly more detailed comment as to why it's needed?

```suggestion
        self.assertFalse(form.is_valid())
```

bagitobjecttransfer/recordtransfer/tests/unit/test_views.py
```diff
@@ -701,3 +708,180 @@ def test_empty_form_submission(self):
...
+        self.client.logout()
+        response = self.client.get(self.url)
+        self.assertRedirects(response, f"{reverse('login')}?next={self.url}")
+
+    def test_access_non_creator_user(self) -> None:
+        """Test that a non-creator user cannot access the view."""
+        non_creator_user = User.objects.create_user(username="noncreator", password="password")
+        self.client.login(username="noncreator", password="password")
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 403)
```
danloveg, line 843: This is an interesting case you've found here. I think the view should actually return 404 in this test. 403 implies the resource exists, but the user can't access it. I would prefer the user to be given a 404 so that they don't even know whether that's a valid resource URL (for security purposes).

danloveg (changes requested):
These tests look clean and self-contained, just a couple small changes I'm recommending. I especially like that you found `self.assertTemplateUsed` - I didn't know that was a function.

bagitobjecttransfer/recordtransfer/tests/unit/test_forms.py
```diff
@@ -433,3 +434,79 @@ def test_form_no_files_uploaded(self) -> None:
...
+
+    def test_form_valid_data(self) -> None:
+        """Case where the form is valid."""
+        form_data = {"name": "New Group", "description": "A new submission group"}
+        form = SubmissionGroupForm(data=form_data, user=self.user)
+        assert form.is_valid()
+        group = form.save()
+        assert group.name == "New Group"
+        assert group.description == "A new submission group"
+        assert group.created_by == self.user
```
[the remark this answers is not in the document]
yenaing-oo: I see, this makes sense.

Addressed by:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/347/commits/d11d93f3b65df1e19fa36a42c4cc0cf624b3929f

bagitobjecttransfer/recordtransfer/tests/unit/test_forms.py
yenaing-oo: Yes, the first one is redundant. I've fixed that here:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/347/commits/d11d93f3b65df1e19fa36a42c4cc0cf624b3929f

danloveg (approved):

Issue #332, opened 2025-01-24
Refactor TransferFormWizard.get_context_data


danloveg:
The method of getting context data is getting a bit out of hand for the transfer form.

I am proposing we make `get_context_data` call two function:

- `get_template_context()` - this will be used to create the context for each template
- `get_javascript_context()` - this will be used to create the context that gets passed to the Javascript files (via json_script)

If `get_javascript_context()` returns something other than an empty dict, that context should be automatically injected into the Javascript (by using json_script in the javascript block). The base transferform template should have this functionality.

Right now, if we want to include some new context for .js files, we have to change three things:
 
- `get_context_data`
- The template
- The .js file

This change would remove the step of having to edit the template.

PR #348, opened 2025-01-31
Refactor TransferFormWizard.get_context_data

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/332


bagitobjecttransfer/recordtransfer/templates/recordtransfer/transferform_base.html
```diff
@@ -10,6 +10,7 @@
     {{ block.super }}
     <script src="{% static 'transferform.bundle.js' %}"></script>
     <script>currentFormStep = {{ wizard.steps.step1 }}</script>
+    {{ js_context|json_script:"py_context" }}
```
danloveg, line 13: Is it possible to incorporate the name of the form or some information specific to the form in the variable `py_context`? Like on the source information form, if this variable could instead by `py_context_sourceinfo` that'd be ideal. The reason I mention this is that the existence of a *specific* `py_context_` element can then be used to tell the JavaScript which page of the form it's supposed to be managing.

danloveg (commented):
This will conflict with my merge request #331 but I like where this is going. Maybe hold off on making any further changes until that merge request is merged in.

bagitobjecttransfer/recordtransfer/templates/recordtransfer/transferform_base.html
yenaing-oo: I see what you mean by this. You want the JS to verify that the context is indeed for that specific step, like this:

```javascript
export function setupSourceInfoForm() {
    // Context passed from template to JS
    const sourceInfoContextElement = document.getElementById("py_context_source_info");

    if (!sourceInfoContextElement) {
        return;
    }
```

I'm looking into whether I can set the name of this context variable dynamically using the current form step within the base template.

yenaing-oo: @danloveg 
Would it suffice to simply check whether a generic `py_context` element is empty or not?

```javascript
export function setupSourceInfoForm() {
    // Context passed from template to JS
    const sourceInfoContextElement = document.getElementById("py_context");

    if (!sourceInfoContextElement || sourceInfoContextElement.textContent.trim() === "{}") {
        return;
    }
```

danloveg: I still think the context variable name needs to be specific to that page.

---

Consider this case:

A user has filled out the contact information form, and proceeds to the source info form. Assume we're sending the context for that page in a script element with the generic ID "py_context".

At this point, this function is called: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/ff21abd9b99c09dbc8d06d91290bee93c884c13d/bagitobjecttransfer/recordtransfer/static/recordtransfer/js/transferform/index.js#L17

Which calls both:

- [`setupSourceInfoForm`](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/ff21abd9b99c09dbc8d06d91290bee93c884c13d/bagitobjecttransfer/recordtransfer/static/recordtransfer/js/transferform/index.js#L21)
- [`setupRightsForm`](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/ff21abd9b99c09dbc8d06d91290bee93c884c13d/bagitobjecttransfer/recordtransfer/static/recordtransfer/js/transferform/index.js#L22)

The context that was injected into the `<head>` of the page is the same when each of these functions is called; they would both load the element with the ID `"py_context"`. The user had navigated to the source info page, so the code in `setupSourceInfoForm` gets called, which is great, but the code in `setupRightsForm` also all gets executed, because there's no distinguishing between the context for each.

danloveg: If it isn't possible to include the step name in the context element's ID, we *could* use the same `"py_context"` for each step, but the context should contain some identifying information as to what step it's for.

```javascript
export function setupSourceInfoForm() {
    // Context passed from template to JS
    const contextElement = document.getElementById("py_context");
    const context = JSON.parse(contextElement.textContent);

    if (!context["form_step"] || context["form_step"] !== "source_info") {
        return;
    }

```

This approach is a bit inefficient because the JSON would need to be parsed inside each of these functions, so we might want to pull the parsing of the JSON out to the function above these setup functions (above meaning inside the caller of those functions, or in other words, [here](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/ff21abd9b99c09dbc8d06d91290bee93c884c13d/bagitobjecttransfer/recordtransfer/static/recordtransfer/js/transferform/index.js#L17)).

yenaing-oo: I agree, I'd rather not parse the contents to check what step it's for. I think it would be better to pass the ID to be used for the context element from the view to the base template, which I did here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/348/commits/508461e69ecb1cb31de101aeb29a663f867b96a9

danloveg: That looks perfect, just what I had in mind.

danloveg (changes requested):
Looks clean, no comment besides my one tiny suggestion.

bagitobjecttransfer/recordtransfer/static/recordtransfer/js/transferform/index.js
```diff
@@ -14,7 +14,7 @@ import {
     setupHelpTooltips,
 } from "./widgets";
 
-document.addEventListener("DOMContentLoaded", function () {
+document.addEventListener("DOMContentLoaded", function () {
```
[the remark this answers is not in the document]
yenaing-oo: Can I ask how you managed to catch that? It did not show up on my linter.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/348/commits/0234ee5197cd2f5e3fa2a25a6c5c44905a3cf869

danloveg (approved):

Issue #360, opened 2025-02-05
Other Source Type and Role Not Showing On Source Information Step


yenaing-oo:
The "Other source type" and "Other source role" input fields are not showing on the Source Information step.

![Image](https://github.com/user-attachments/assets/d993bcea-345b-4387-9093-68621513acd1)

PR #361, opened 2025-02-05
Fix Other Source Type and Role Not Showing On Source Information Step

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/360

### DIFF
--- a/bagitobjecttransfer/recordtransfer/static/recordtransfer/js/transferform/otherField.js
+++ b/bagitobjecttransfer/recordtransfer/static/recordtransfer/js/transferform/otherField.js
@@ -27,8 +27,7 @@ export function setupSelectOtherToggle(selectFieldInputId, otherFieldInputId, ot
         }
 
         const container = otherInput.parentElement;
-
-        if (this.value === otherValue) {
+        if (Number(this.value) === Number(otherValue)) {
             container.classList.remove("hidden-item");
         }
         else {


danloveg (changes requested):
This fixes the issue, but I'm noticing one other issue. The Other field is not present when a user doesn't fill out the Other input.

Screenshot 2025-02-05 155224

If you change the select field to something else and then back to Other that brings the field back, but it should be present when the form first renders if possible.

yenaing-oo:
> This fixes the issue, but I'm noticing one other issue. The Other field is not present when a user doesn't fill out the Other input.
> 
> Screenshot 2025-02-05 155224
> 
> If you change the select field to something else and then back to Other that brings the field back, but it should be present when the form first renders if possible.

That's a good catch, I've addressed that here:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/361/commits/03654105e64864ec930ccbdd5955fa8999021a1a

danloveg (changes requested):
The new `updateOtherFieldVisibility` function works great. But I'm noticing another bug. This is probably something I introduced, but if you try to submit a form with errors on the rights page with an empty "Other" field, all but the very first form is removed from the page.

This appears to be because Django is rendering the formset with INITIAL_FORMS=0 for some reason. Then the code that runs the form removal removes everything but the first form.

Can you take a look at this? To re-produce, you can:

Make three formset rows on the Rights page. On each, select:

- Access
- **Other** (but don't fill out the other field)
- Access

And click **Next Step**. All but the first Access row is deleted. The rows *are* rendered at first but they get removed when the code [here](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/8c819204be5ce860ac03340496cc0f77edaef556/bagitobjecttransfer/recordtransfer/static/recordtransfer/js/transferform/formset.js#L34) runs.

Screenshot 2025-02-06 085221

danloveg:
> The new `updateOtherFieldVisibility` function works great. But I'm noticing another bug. This is probably something I introduced, but if you try to submit a form with errors on the rights page with an empty "Other" field, all but the very first form is removed from the page.

And funnily enough, this *doesn't* happen if you do this with a loaded in-progress submission. So if you save the current transfer form on the rights page in this weird state, and load it back in, and then try to go to the next step, that issue doesn't happen at all.

yenaing-oo:
> This appears to be because Django is rendering the formset with INITIAL_FORMS=0 for some reason. Then the code that runs the form removal removes everything but the first form.

I think this is because Django does not recalculate this value, but rather, it uses whatever value was submitted for `INITIAL_FORMS`. A fresh Rights step will have TOTAL_FORMS=1 and INITIAL_FORMS=0. After adding two more rights, only the TOTAL_FORMS is updated to 3, but INITIAL_FORMS remains 0. After the form submits, and fails validation due to the missing Other field, it is reloaded and the last two forms gets removed since `numForms > 1`. 

I've addressed that issue here by updating `INITIAL_FORMS` to match `TOTAL_FORMS`: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/361/commits/9f989b4637b47c27438f3fd146c604ae4a99ef4c

danloveg (approved):
Your latest change fixed it, it works flawlessly now.

Issue #351, opened 2025-02-03
Install node dependencies before building static assets for E2E tests


yenaing-oo:
`npm install` should be run before `npm run build` is executed in `conftest.py`.

PR #365, opened 2025-02-06
Run npm install Before Asset Build in conftest.py

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/351


danloveg (approved):
Perfect, nice and simple. I was wondering if we should check if `node_modules` already exists but I think this approach is the right way to do it since the packages needed may change over time. Figuring out if an `npm install` is actually needed is probably not worth worrying about.

Issue #367, opened 2025-02-07
Commit ruff.toml to repository


danloveg:
I recommend we commit `ruff.toml` directly to the root of the repository rather than include it in a Git hook.

We expected `ruff.toml` to change often, hence why we made the decision to pull the latest version inside a pre-commit hook. But this has two drawbacks:

1. `ruff.toml` *hasn't* changed, so we're making a useless web request for this file every commit.
2. If someone pulls the repository for the first time, writes some Python, then makes a commit, Ruff wouldn't have been linting or formatting prior to making that commit. That commit then might contain a bunch of poorly formatted Python.

yenaing-oo:
I agree with both points. I still like the idea of having an automating any changes to `ruff.toml`, so perhaps we can use a GH Action instead? I made a PR to demonstrate my solution:  https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/369

PR #369, opened 2025-02-07
Sync ruff.toml Using Github Action

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/367

### DIFF
--- a/.github/workflows/sync-ruff-config.yml
+++ b/.github/workflows/sync-ruff-config.yml
@@ -24,17 +24,17 @@ jobs:
 
       - name: Fetch ruff.toml from NCTR's Python Development Guide
         run: |
-          curl -o ruff.toml https://raw.githubusercontent.com/NationalCentreTruthReconciliation/Python-Development-Guide/main/ruff.toml
+          curl -o ./ruff_new.toml https://raw.githubusercontent.com/NationalCentreTruthReconciliation/Python-Development-Guide/main/ruff.toml
 
       - name: Set up Git
         run: |
           git config --global user.name "github-actions[bot]"
           git config --global user.email "github-actions[bot]@users.noreply.github.com"
 
-      - name: Compare and commit changes if different
+      - name: Update ruff.toml if different
         run: |
-          if ! cmp -s ruff.toml ./ruff.toml; then
-            mv ruff.toml ./ruff.toml
+          if ! cmp -s ./ruff_new.toml ./ruff.toml; then
+            mv ./ruff_new.toml ./ruff.toml
             git add ./ruff.toml
             git commit -m "Sync ruff.toml from NationalCentreTruthReconciliation/Python-Development-Guide"
           fi
--- a/.github/workflows/sync-ruff-config.yml
+++ b/.github/workflows/sync-ruff-config.yml
@@ -1,0 +1,32 @@
+name: Sync Ruff Configuration
+
+on:
+  schedule:
+    - cron: "0 0 * * 0" # Runs every week
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+
+jobs:
+  sync-file:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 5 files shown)


danloveg:
As I understand it, this would have a bot make a commit to the default branch (i.e., `master`) every week, given that the `ruff.toml` file has changed (feel free to correct me if that's not how the new action works).

While I like the idea of using GH actions to automate our housekeeping, I don't love the idea of a bot pushing to the master branch. Is it possible to change it so that it creates a pull request instead? I'm not sure how difficult that would be.

Also I know it's Friday, don't feel obligated to respond right now!

yenaing-oo:
> As I understand it, this would have a bot make a commit to the default branch (i.e., `master`) every week, given that the `ruff.toml` file has changed (feel free to correct me if that's not how the new action works).

Yes, this is how my original workflow would work.

> While I like the idea of using GH actions to automate our housekeeping, I don't love the idea of a bot pushing to the master branch. Is it possible to change it so that it creates a pull request instead? I'm not sure how difficult that would be.

I agree, I've changed the workflow so that it now makes a new branch `sync-ruff-config` from `master` if it doesn't exist, or resets an existing `sync-ruff-config` to the head of `master` if it does already exist. It'll make a commit with the update changes there, and create a pull request if any changes exist.

Those changes can be found here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/369/commits/5e2376370b70bc0693a3b2c7711fb003d10f9682

.github/workflows/sync-ruff-config.yml
```diff
@@ -0,0 +1,50 @@
...
+          if git ls-remote --exit-code --heads origin sync-ruff-config; then
+            git checkout sync-ruff-config
+            git reset --hard origin/master
+          else
+            git checkout -b sync-ruff-config
+          fi
+
+      - name: Fetch ruff.toml from NCTR's Python Development Guide
+        run: |
+          curl -o ruff.toml https://raw.githubusercontent.com/NationalCentreTruthReconciliation/Python-Development-Guide/main/ruff.toml
```
danloveg, line 27: These GH actions are a bit abstract to me, but I think the directory this command runs in is the same as the root of the repository, right?

If that's the case, I have another issue I flagged below.

.github/workflows/sync-ruff-config.yml
```diff
@@ -0,0 +1,50 @@
...
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+
+      - name: Compare and commit changes if different
+        run: |
+          if ! cmp -s ruff.toml ./ruff.toml; then
+            mv ruff.toml ./ruff.toml
+            git add ./ruff.toml
+            git commit -m "Sync ruff.toml from NationalCentreTruthReconciliation/Python-Development-Guide"
+          fi
```
danloveg, line 40: I believe `ruff.toml` and `./ruff.toml` are the same file, and that the `curl` command above would have overwritten the one already in the repository. `cmp` will never produce any output here since it is comparing the same file (which is identical to itself)

yenaing-oo: You're absolutely correct on this. I thought the downloaded `ruff.toml` would be stored in a different location and accessible by filename, but it is being downloaded into the checked out repoisitory.

I've used a different name for the downloaded file here to address this issue: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/369/commits/6e76f1eed2ccdf603057e6d416460c5d1b7ccb9a

danloveg (approved):
Looks good, let's get this in and test it out. I think you should be able to manually trigger it since `workflow_dispatch` is in the triggers.

danloveg:
@yenaing-oo I'm not going to delete this branch until we've tested this. If there are any changes needed you can use this same branch

PR #378, opened 2025-02-12
Standardize File Size Units to MB

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/305

PR #381, opened 2025-02-13
Refactor TransferFormWizard.set_quantity_and_unit_of_measure

yenaing-oo:


app/recordtransfer/utils.py
```diff
@@ -1,14 +1,16 @@
...
 
 from django.conf import settings
 from django.utils.html import strip_tags
 from django.utils.translation import gettext
 
 from recordtransfer.exceptions import FolderNotFoundError
-from recordtransfer.models import UploadSession
+
+if TYPE_CHECKING:
+    from recordtransfer.models import UploadSession
```
yenaing-oo, line 13: This is to avoid a circular import in `utils.py`, where `accept_session()` is using `UploadSession` in its type annotation.

danloveg: Can you add a comment to that effect?

app/recordtransfer/models.py
```diff
@@ -407,6 +408,42 @@ def copy_session_uploads(
...
+            self.SessionStatus.UPLOADING,
+            self.SessionStatus.STORED,
+        ):
+            raise ValueError(
+                f"Cannot get quantity and unit of measure from session {self.token} because the "
+                f"session status is {self.status} and not {self.SessionStatus.UPLOADING} or "
+                f"{self.SessionStatus.STORED}"
+            )
+
+        size = get_human_readable_size(self.upload_size, base=1024, precision=2)
```
danloveg, line 437: Change based on #378

```suggestion
        size = get_human_readable_size(self.upload_size, base=1000, precision=2)
```

danloveg (changes requested):

app/recordtransfer/utils.py
yenaing-oo: For sure, I've added one here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/381/commits/5c71118c0a5980517ef4252e0cf9a5d42f9ee07f

I didn't make it very specific where the circular import was happening on purpose, since the function name `accept_session` could change in the future

app/recordtransfer/models.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/381/commits/dd01e0df871b4f46d424e5f5b2fb84e120774bd5

danloveg (dismissed):

danloveg (approved):

Issue #380, opened 2025-02-13
npm commands are being run when debugging unit tests


danloveg:
The `npm` commands used to run Webpack are being run when a Django test is debugged from VS Code. This should be fixed so that those commands are not run when debugging unit tests.

To reproduce:

- Choose any test, and Debug it in VS Code.

Look in the Debug Console in VS Code. You should see the output from the Webpack command:

![Image](https://github.com/user-attachments/assets/6b055c98-e812-4263-890b-439b316f5ad5)

danloveg:
It looks like it's because pytest doesn't pass the `-k` option when debugging tests. On my machine, these are the arguments used:

```
Running pytest with args: ['-p', 'vscode_pytest', '--rootdir=c:\\Users\\lovegrod\\Documents\\NCTR_Code\\public\\Secure-Record-Transfer', '--capture=no', 'c:\\Users\\lovegrod\\Documents\\NCTR_Code\\public\\Secure-Record-Transfer\\app\\recordtransfer\\tests\\unit\\views\\test_home.py::TestAboutPage::test_accepted_file_types_shown']
```

I wonder if there's a better way to conditionally run those npm commands without relying on `-k`.

PR #385, opened 2025-02-13
Prevent npm commands being run when tests are run through VSCode

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/380

I replaced `pytest_collection_modifyitems(session, config, items)` with `pytest_runtestloop(session)`, the difference being that the former's `items` includes all tests that are discoverable, whereas the latter's `session.items` only includes tests which are selected to be run.

In this pytest hook, I'm looking through each test item's id (which looks like this: `app/recordtransfer/tests/e2e/test_e2e.py::TransferFormWizardTest::test_login`) to see if any of them include the substring "/e2e/". If any of them do, I run the npm commands.

This should now work as expected for both running tests from the command line and throguh VSCode's test explorer.


conftest.py
```diff
@@ -9,12 +9,10 @@
...
 
 
-def pytest_collection_modifyitems(session, config, items) -> None:
-    """Run npm build only if e2e tests are being run. `items` is a list of test items to be
-    run.
-    """
-    # Check if running unit tests via -e2e parameter or running all tests
-    if config.getoption("-k") == "e2e" or not config.getoption("-k"):
+def pytest_runtestloop(session: pytest.Session) -> None:
+    """Install node dependencies and isntall static assets only if E2E tests are selected to run."""
```
danloveg:
```suggestion
    """Install node dependencies and build static assets only if E2E tests are selected to run."""
```

danloveg (approved):
Works great! I accidentally submitted a review too quickly but it's skipping the npm commands now which is good.

Issue #384, opened 2025-02-13
Refactor TransferFormWizard


danloveg:
The transfer form wizard is one of the most complicated pieces of code in the codebase, and I want to keep it tidy to keep it easy to understand to avoid accruing technical debt.

There are a number of areas I found improvements for when reading the code with a critical eye.

These are a series of small refactors, they can all be addressed within the same PR.

# `validate_transfer_request`

This can be combined into `dispatch` and the function can be removed. This way, there doesn't need to be a `None` return type and the logic will be easier to follow.

# `get`

The call to the superclass should be changed to:

```python
return super().get(request, *args, **kwargs)
```

# `load_transfer_data`

This function should:

- Be renamed to `load_form_data`
- Be moved below `get`
- Access `self.in_progress_submission` directly rather than use it from the args

# `post`

The logic inside this function should be moved to the save transfer function (see next section).

The post method should end up looking like this:

```python
def post(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
    """Handle POST request to save a transfer."""
    if not request.POST.get("save_form_step"):
        return super().post(request, *args, **kwargs)
    self.save_form_data(request)
    return redirect("recordtransfer:userprofile")
```

# `save_transfer`

This function should:

- Be renamed to `save_form_data`
- Be moved below `post`
- Access `self` attributes directly rather than used them as they are passed to the method
- Accept only one argument: `request`

The function signature `save_transfer` uses a `data` argument instead of keyword arguments, which makes the function hard to read. Everything needed by that function can be accessed with `self` (except for request, which is why I'm recommending it be passed as an argument).

# `get_template_names`

This function is not used, it should be removed.

# `get_submission_group`

We may be able to refactor this into `transfer_forms.GroupTransferForm.clean()` by having that method return the actual group. The code may complain about a `SubmissionGroup` not being serializable, though, when the data is saved in a pickle.

If this is too complex to implement I recommend we leave it as-is. My goal with this issue is to make things simpler and easier to read, not more complex.

# `done`

This is a small one, but the `form_data` variable might be unbound, so it should be set outside of the `try` block.

yenaing-oo:
> # `get_template_names`
> This function is not used, it should be removed.

This method is actually overriding the superclass method to dynamically select a different template for each step.
https://django-formtools.readthedocs.io/en/latest/wizard.html#using-a-different-template-for-each-form

danloveg:
[...]

Issue #386, opened 2025-02-13
Contact name and email is overriden when resuming form


yenaing-oo:
The fields `contact_name` and `email` are overwritten as part of the form data re-population from resuming a form.

https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/74df516c63d1c922efc96d87bb72abc4d826d0f6/app/recordtransfer/views/transfer.py#L431-L433

Those fields should be populated only if it is a fresh form.

danloveg:
Closed by #387

PR #387, opened 2025-02-13
Refactor TransferFormWizard

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/384, https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/386 and https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/388

### DIFF
--- a/app/recordtransfer/forms/transfer_forms.py
+++ b/app/recordtransfer/forms/transfer_forms.py
@@ -889,23 +889,23 @@ class GroupTransferForm(TransferForm):
         choices = [(None, "Select a group")] + [
             (group.uuid, group.name) for group in user_submission_groups
         ]
-        self.fields["group_id"].choices = choices
-        self.fields["group_id"].initial = None
+        self.fields["group_uuid"].choices = choices
+        self.fields["group_uuid"].initial = None
 
     def clean(self) -> dict:
         """Check that chosen group exists and is owned by the user."""
         cleaned_data = super().clean()
-        group_id = cleaned_data.get("group_id")
+        group_uuid = cleaned_data.get("group_uuid")
 
-        group = SubmissionGroup.objects.filter(created_by=self.user, uuid=group_id).first()
+        group = SubmissionGroup.objects.filter(created_by=self.user, uuid=group_uuid).first()
         if not group:
-            raise ValidationError({"group_id": "Invalid group selected."})
+            raise ValidationError({"group_uuid": "Invalid group selected."})
 
         cleaned_data["submission_group"] = group
 
         return cleaned_data
 
-    group_id = forms.TypedChoiceField(
+    group_uuid = forms.TypedChoiceField(
         required=False,
         coerce=UUID,
         widget=forms.Select(
@@ -1049,19 +1049,19 @@ class ReviewForm(TransferForm):
     def _get_group_transfer_fields_data(form: GroupTransferForm, user: User) -> dict[str, Any]:
         """Handle group transfer specific field processing."""
         fields_data = ReviewForm._get_base_fields_data(form)
-        group_id = form.cleaned_data.get("group_id")
+        group_uuid = form.cleaned_data.get("group_uuid")
 
-        if not (group_id and user):
+        if not (group_uuid and user):
             return fields_data
 
-        group = SubmissionGroup.objects.filter(created_by=user, uuid=group_id).first()
+        group = SubmissionGroup.objects.filter(created_by=user, uuid=group_uuid).first()
         if not group:
             return fields_data
 
         # Replaces group UUID with group name
-        group_id_label = form.fields["group_id"].label
-        if group_id_label:
-            fields_data[group_id_label] = group.name
+        group_uuid_label = form.fields["group_uuid"].label
+        if group_uuid_label:
+            fields_data[group_uuid_label] = group.name
 
         return fields_data
 
--- a/app/recordtransfer/forms/transfer_forms.py
+++ b/app/recordtransfer/forms/transfer_forms.py
@@ -11,7 +11,7 @@ from caais.models import RightsType, SourceRole, SourceType
 from django import forms
 from django.conf import settings
 from django.db.models import Case, CharField, Value, When
-from django.forms import BaseForm, BaseFormSet
+from django.forms import BaseForm, BaseFormSet, ValidationError
 from django.utils.translation import gettext
 from django_countries.fields import CountryField
 from django_countries.widgets import CountrySelectWidget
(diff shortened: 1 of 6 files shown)


yenaing-oo:
- https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/387/commits/6e01eaba64d7984e84b6c41dde6eb8eda3b62a3e addresses https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/386
- https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/387/commits/1ece65beef044f329163957bad798e64bffa2dc4 addresses https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/388

app/recordtransfer/forms/transfer_forms.py
```diff
@@ -889,23 +889,23 @@ def __init__(self, *args, **kwargs):
...
+        group_uuid = cleaned_data.get("group_uuid")
 
-        if (
-            group_id
-            and not SubmissionGroup.objects.filter(created_by=self.user, uuid=group_id).exists()
-        ):
-            self.add_error("group_id", "Invalid group selected.")
+        group = SubmissionGroup.objects.filter(created_by=self.user, uuid=group_uuid).first()
+        if not group:
+            raise ValidationError({"group_uuid": "Invalid group selected."})
```
danloveg, line 902: Should be:

```suggestion
            self.add_error("group_uuid", "Invalid group selected.")
```

Though if there was a specific reason why you raised an error here you can ignore this suggestion.

app/recordtransfer/views/transfer.py
```diff
@@ -157,74 +155,77 @@ def current_step(self) -> TransferStep:
...
-            "save_form_step": self.current_step,
-            "form_data": {"past": past_data, "current": current_data},
-            "submission": self.in_progress_submission,
-            "title": title,
-        }
+        submission = self.in_progress_submission
+        if submission:
+            submission.last_updated = timezone.now()
+        else:
+            submission = InProgressSubmission()
```
danloveg, line 222: You can use `self.in_progress_submission` without having to assign it to a different variable:

```suggestion
        if self.in_progress_submission:
            self.in_progress_submission.last_updated = timezone.now()
        else:
            self.in_progress_submission = InProgressSubmission()
```

app/recordtransfer/views/transfer.py
```diff
@@ -157,74 +155,77 @@ def current_step(self) -> TransferStep:
...
-            self.save_transfer(data)
-            messages.success(request, gettext("Transfer saved successfully."))
-        except Exception:
-            messages.error(request, gettext("There was an error saving the transfer."))
-        return redirect("recordtransfer:userprofile")
+        submission.current_step = self.current_step.value
+        submission.user = cast(User, self.request.user)
+        submission.step_data = pickle.dumps(form_data)
+        submission.title = title
+        submission.save()
```
danloveg, line 228:
```suggestion
        self.in_progress_submission.current_step = self.current_step.value
        self.in_progress_submission.user = cast(User, self.request.user)
        self.in_progress_submission.step_data = pickle.dumps(form_data)
        self.in_progress_submission.title = title
        self.in_progress_submission.save()
```

danloveg (changes requested):
The GET and POST logic is much easier to follow now, which is great.

Besides the changes I suggested, I also have a bit of a weird edge case that we need to consider. Imagine a user does these things in this order:

- Selects a group on the Group page of the transfer form
- Goes to the next step
- Saves the form for later
- Deletes the group
  - Users can't do this yet, but it can be done from the backend
- Loads that same in-progress submission

The Group form is now invalid because the chosen group doesn't exist anymore. You can therefore run into an error like this:

Screenshot 2025-02-14 092251

Can the group form data be post-processed to make it seem like the user didn't select a group? (i.e., group_uuid="" and submission_group=None)

yenaing-oo:
> The GET and POST logic is much easier to follow now, which is great.
> 
> Besides the changes I suggested, I also have a bit of a weird edge case that we need to consider. Imagine a user does these things in this order:
> 
> * Selects a group on the Group page of the transfer form
> * Goes to the next step
> * Saves the form for later
> * Deletes the group
>   
>   * Users can't do this yet, but it can be done from the backend
> * Loads that same in-progress submission
> 
> The Group form is now invalid because the chosen group doesn't exist anymore. You can therefore run into an error like this:
> 
> Screenshot 2025-02-14 092251
> 
> Can the group form data be post-processed to make it seem like the user didn't select a group? (i.e., group_uuid="" and submission_group=None)

That is an interesting edge case I had not considered! To make it look as though the user never selected a group upon resuming a submission that is assigned a submission group that no longer exists, I would have to **not** raise any validation error in addition to clearing those fields. I think this should be fine, since the only way an invalid group_uuid could be passed from the client side is through direct manipulation of the form (e.g. via Postman).

I've addressed this here: 
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/387/commits/d26c6205bab660668aaa39355378288d482abfdd

app/recordtransfer/views/transfer.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/387/commits/1bf338f28ebfd80e9ea7d1e94dd778837581eed6

app/recordtransfer/views/transfer.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/387/commits/1bf338f28ebfd80e9ea7d1e94dd778837581eed6

app/recordtransfer/forms/transfer_forms.py
yenaing-oo: I raised a ValidationError on purpose there since I didn't want to include the `submission_group` in the `cleaned_data`. Showing an error is no longer necessary as per https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/387/commits/d26c6205bab660668aaa39355378288d482abfdd

app/recordtransfer/forms/transfer_forms.py
```diff
@@ -11,7 +11,7 @@
 from django import forms
 from django.conf import settings
 from django.db.models import Case, CharField, Value, When
-from django.forms import BaseForm, BaseFormSet
+from django.forms import BaseForm, BaseFormSet, ValidationError
```
danloveg, line 14: Unused import:

```suggestion
from django.forms import BaseForm, BaseFormSet
```

danloveg (changes requested):
I noted one last change before we're good to merge this in.

danloveg (approved):

PR #389, opened 2025-02-13
Fix running tests when using manage.py

### LINKED ISSUE #349: Fix TypeError from abstract class TestBaseUploadedFile
When running `python manage.py test`, the following error is thrown:

```
TypeError: Can't instantiate abstract class TestBaseUploadedFile without an implementation for abstract method 'model_class'
```

danloveg:
Closes #349

The tests now work whether you're using `manage.py`, `pytest`, or the VS Code test explorer.

I removed a package that was only being used when run via `manage.py`. Since the tests are running fine without that package, I've decided to remove it.

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -27,7 +27,7 @@ The debugging configuration has already been set up in the `.vscode/launch.json`
 A GitHub Actions workflow is set up to run Django tests on every pull request to the master branch. All tests must pass before a merge is allowed. The workflow configuration can be found in `.github/workflows/django-tests.yml`.
 
 
-### Re-build JS as Changes are Made
+## Re-build JS as Changes are Made
 
 After you run the dev container (`compose.dev.yml`), you can automatically re-build the JS files any time they change with this command:
 
@@ -41,7 +41,7 @@ podman-compose -f compose.dev.yml exec app npm run watch
 
 This will re-build the bundled JS files any time you save a change to a `.js` file which is useful while you're writing Javascript. This command does not exit until you press CTRL-C, so make sure to run it in a separate terminal.
 
-# Virtual Environment and Poetry Setup
+## Virtual Environment and Poetry Setup
 
 Python dependencies are managed with [Poetry](https://python-poetry.org/). To install Poetry and create a virtual environment with the minimally necessary packages, follow these instructions:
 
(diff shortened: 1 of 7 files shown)


DEVELOPERS.md
```diff
@@ -71,43 +71,62 @@ Finally, install the dependencies with `poetry`:
...
+```
 
 ## Building the Documentation
 
 The documentation for this repository is built with [Sphinx](https://sphinx-doc.org).
 
 Ensure you follow the instructions in the [Virtual Environment and Poetry Setup](#virtual-environment-and-poetry-setup) section before continuing.
 
-Install the `docs` dependencies with Poetry:
+Install the `dev` dependencies with Poetry:
```
yenaing-oo, line 126: Is this not meant to be `docs` here? I'm aware `dev` would work here too, since it includes all of `docs` dependencies.

yenaing-oo (commented):
Everything looks good and works well. I like how you managed to find a way to remove the abstract test class while avoiding code repetition. I just had one question regarding the extras group used for building docs.

DEVELOPERS.md
danloveg: The reason I changed that is because I'm assuming this documentation is aimed at developers. Further up in the document, we recommend running:

```
poetry install -E dev
```

But, if we tell people to run this command at this part of the documentation:

```
poetry install -E docs
```

Then Poetry *uninstalls* the extras that aren't in `docs`. I don't love this behaviour but I guess that was a choice the Poetry devs made. You can try it for yourself, run `poetry install -E dev` and then `poetry install -E docs`; pytest, pytest-django, and selenium are all uninstalled.

yenaing-oo: > Then Poetry _uninstalls_ the extras that aren't in `docs`. I don't love this behaviour but I guess that was a choice the Poetry devs made. You can try it for yourself, run `poetry install -E dev` and then `poetry install -E docs`; pytest, pytest-django, and selenium are all uninstalled.

Ah I see. You're right, they do get uninstalled. I'm fine with the way it is then.


> Edit: And the reason I still kept the `docs` extras in `pyproject.toml` is for building the docs in Read the Docs:
> 
> https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/74df516c63d1c922efc96d87bb72abc4d826d0f6/.readthedocs.yaml#L19

This makes sense, thanks for clarifying.

yenaing-oo (approved):

Issue #383, opened 2025-02-13
Fix stretched country icon when editing CAAIS metadata in admin


danloveg:
The country icon gets weirdly stretched when editing a CAAIS Metadata object from the admin site:

![Image](https://github.com/user-attachments/assets/f56dbf93-c46b-4d0a-9bf0-2e8d8c0ab82b)

To get to this page, go to `/admin/caais/metadata/add/` and scroll down to SOURCES OF MATERIAL.

This should be fixed so that the icon's aspect ratio is not stretched.

PR #397, opened 2025-02-14
Fix stretched country icon when editing CAAIS metadata in admin

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/383

It should look like this now:
image

### DIFF
--- a/app/caais/static/caais/css/base/metadataChangeForm.css
+++ b/app/caais/static/caais/css/base/metadataChangeForm.css
@@ -3,10 +3,12 @@
 .inline-textarea {
     width: 97%;
 }
+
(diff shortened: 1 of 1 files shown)


yenaing-oo:
I wanted to add a left-margin to the flag, but it looks like the margins can't be overrided a they are being set using in-line styles:

image

danloveg:
> I wanted to add a left-margin to the flag, but it looks like the margins can't be overrided as they are being set using in-line styles:

That's fine, I noticed that that field was being rendered with inline styles, too. Fixing the stretched out icon is good enough

danloveg (approved):

Issue #390, opened 2025-02-14
Create Upload Session Expiry Settings


danloveg:
Add and document three new settings:

- `UPLOAD_SESSION_EXPIRE_AFTER_INACTIVE_MINUTES`
  - Default should be 1440 (a day).
  - This feature can be deactivated by setting to -1.
  - Validation: -1 or >= 1
- `UPLOAD_SESSION_EXPIRED_CLEANUP_SCHEDULE`
  - Default should be "0 2 * * *" (every day at 2am).
  - This feature can be deactivated by setting to "".
  - Validation: valid cron string
- `UPLOAD_SESSION_EXPIRING_REMINDER_MINUTES`
  - Default should be 480 (8 hours).
  - This feature can be deactivated by setting to -1.
  - Validation: -1 or < `UPLOAD_SESSION_EXPIRE_AFTER_INACTIVE_MINUTES`

danloveg:
Closed by #399

PR #399, opened 2025-02-18
Create Upload Session Expiry Settings

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/390

- I couldn't find a well-maintained package to validate cron strings. The best I could find was https://github.com/corpusops/croniter but the project was abandoned quite recently. Since the syntax for cron strings are not overly complex, I made my own validation using regex.
- I've included a test class to test the cron string validation method


app/recordtransfer/apps.py
```diff
@@ -238,27 +245,112 @@ def verify_accepted_file_formats() -> None:
...
+    - UPLOAD_SESSION_EXPIRED_CLEANUP_SCHEDULE
+    - UPLOAD_SESSION_EXPIRE_AFTER_INACTIVE_MINUTES
+    - UPLOAD_SESSION_EXPIRING_REMINDER_MINUTES
+
+    Raises:
+        ImproperlyConfigured: If the expiry schedule is not a valid cron string, or if the
+        expiry settings are invalid.
+    """
+    try:
+        _validate_cron(settings.UPLOAD_SESSION_EXPIRED_CLEANUP_SCHEDULE)
```
danloveg, line 307: This setting should be allowed to be empty, since an empty string will disable this feature.

```suggestion
        cron_str = settings.UPLOAD_SESSION_EXPIRED_CLEANUP_SCHEDULE
        if cron_str:
            _validate_cron(cron_str)
```

danloveg (changes requested):
One tiny comment here but otherwise no comment, this looks great.

danloveg (approved):

Issue #392, opened 2025-02-14
Add Expiry Capability to Upload Session model


danloveg:
Add a `last_upload_interaction_time` field to `UploadSession` (with `auto_now_add=True`). Update this field whenever a file is uploaded or finished uploading.

Add an `expires_at()` method to `UploadSession`. This should return a datetime if the state is `CREATED` or `UPLOADING`, or None if not in either of those states. It may also return None if upload sessions cannot be expired.

Add an `upload_session_expires_at()` helper method to `InProgressSubmission`. This should return None if there is no upload session, or the result from `UploadSession.expires_at()`.

Add a *calculated* `expired` field to `UploadSession`. This should compare the last_interaction_time with the `UPLOAD_SESSION_EXPIRE_AFTER_INACTIVE_MINUTES` setting. This should only ever return True if the session is `CREATED` or `UPLOADING`.

PR #402, opened 2025-02-18
Add Expiry Capability to UploadSession Model

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/392


yenaing-oo:
@danloveg 
Would it be a good idea to temporarily include [298-upload-session-expiry](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/tree/298-upload-session-expiry) as a branch for which automated tests should be run?

```
on:
  pull_request:
    branches:
      - master
      - 298-upload-session-expiry
  push:
    branches:
      - master
      - 298-upload-session-expiry
```

app/recordtransfer/tests/unit/test_models.py
```diff
@@ -98,6 +108,72 @@ def test_new_session_creation(self) -> None:
...
+        )
+        self.assertEqual(self.session.expires_at, expected_expires_at)
+
+        # Test UPLOADING state
+        self.session.status = UploadSession.SessionStatus.UPLOADING
+        self.session.last_upload_interaction_time = fixed_now
+        expected_expires_at = self.session.last_upload_interaction_time + timezone.timedelta(
+            minutes=30
+        )
+        self.assertEqual(self.session.expires_at, expected_expires_at)
```
danloveg, line 134: Is this a valid simplification?

```suggestion
        valid_states = [
            UploadSession.SessionStatus.CREATED,
            UploadSession.SessionStatus.UPLOADING,
        ]

        self.session.last_upload_interaction_time = fixed_now
        expected_expires_at = fixed_now + timezone.timedelta(minutes=30)

        for state in valid_states:
            self.session.status = state
            self.assertEqual(self.session.expires_at, expected_expires_at)
```

app/recordtransfer/tests/unit/test_models.py
```diff
@@ -98,6 +108,72 @@ def test_new_session_creation(self) -> None:
...
+        mock_now.return_value = fixed_now
+
+        # Test expired returns True for a session that has expired
+        self.session.status = UploadSession.SessionStatus.CREATED
+        self.session.last_upload_interaction_time = fixed_now - timezone.timedelta(minutes=31)
+        self.assertTrue(self.session.expired)
+
+        # Test expired returns False for a session that has not expired
+        self.session.last_upload_interaction_time = fixed_now - timezone.timedelta(minutes=29)
+        self.assertFalse(self.session.expired)
```
danloveg, line 163: This should be moved to a separate test case

app/recordtransfer/tests/unit/test_models.py
```diff
@@ -98,6 +108,72 @@ def test_new_session_creation(self) -> None:
...
+            UploadSession.SessionStatus.EXPIRED,
+            UploadSession.SessionStatus.DELETED,
+            UploadSession.SessionStatus.COPYING_IN_PROGRESS,
+            UploadSession.SessionStatus.REMOVING_IN_PROGRESS,
+            UploadSession.SessionStatus.STORED,
+            UploadSession.SessionStatus.COPYING_FAILED,
+        ]
+        for state in invalid_states:
+            self.session.status = state
+            self.assertIsNone(self.session.expires_at)
```
danloveg, line 147: These should be moved to a separate test case

app/recordtransfer/tests/unit/test_models.py
```diff
@@ -98,6 +108,72 @@ def test_new_session_creation(self) -> None:
...
+            UploadSession.SessionStatus.EXPIRED,
+            UploadSession.SessionStatus.DELETED,
+            UploadSession.SessionStatus.COPYING_IN_PROGRESS,
+            UploadSession.SessionStatus.REMOVING_IN_PROGRESS,
+            UploadSession.SessionStatus.STORED,
+            UploadSession.SessionStatus.COPYING_FAILED,
+        ]
+        for state in invalid_states:
+            self.session.status = state
+            self.assertFalse(self.session.expired)
```
danloveg, line 176: This should be moved to a separate test case

app/recordtransfer/models.py
```diff
@@ -930,13 +952,23 @@ class InProgressSubmission(models.Model):
...
 
+    def upload_session_expires_at(self) -> Optional[timezone.datetime]:
+        """Get the expiration time of the upload session associated with this submission."""
+        if self.upload_session:
+            return self.upload_session.expires_at
+        return None
+
     def __str__(self):
+        """Return a string representation of this object."""
         return f"Transfer of {self.last_updated} by {self.user}"
```
danloveg, line 974: Adding a docstring here has highlighted that this str method isn't very good. How about this (open to suggestions):

```suggestion
    def __str__(self):
        """Return a string representation of this object."""
        title = self.title or "None"
        session = self.upload_session.token if self.upload_session else "None"
        return f"In-Progress Submission by {self.user} (Title: {title} | Session: {session})"
```

danloveg (changes requested):
I was hoping we could avoid having to link an upload session to an in-progress submission but I don't think that's possible - we'd have to deserialize the pickle every time we wanted that to get at the session's token. All good.

I have a few comments about the tests. It's good practice to generally handle each separate case within a separate test function. Sometimes splitting tests out into separate tests makes the tests *less* clean, but in this case moving the upload session expiry tests out into different functions should help to make the tests a bit more "SRP". For example, `test_expired` should be separated into three separate cases:

1. Test that expired can be True when status ∈ {CREATED, UPLOADING}
2. Test that expired can be False when status ∈ {CREATED, UPLOADING}
3. Test that expired is always False when status ∉ {CREATED, UPLOADING}

An example of where separating tests out into more functions makes the tests less clean is when you're testing all the different `invalid_states` when checking for expiry. Moving each of those out to separate tests would make the tests harder to read since you'd be introducing a bunch of clutter.

yenaing-oo:
> I was hoping we could avoid having to link an upload session to an in-progress submission but I don't think that's possible - we'd have to deserialize the pickle every time we wanted that to get at the session's token. All good.

Yes, deserializing the pickle every time would not be efficient.

> I have a few comments about the tests. It's good practice to generally handle each separate case within a separate test function. Sometimes splitting tests out into separate tests makes the tests _less_ clean, but in this case moving the upload session expiry tests out into different functions should help to make the tests a bit more "SRP". For example, `test_expired` should be separated into three separate cases:
> 
> 1. Test that expired can be True when status ∈ {CREATED, UPLOADING}
> 2. Test that expired can be False when status ∈ {CREATED, UPLOADING}
> 3. Test that expired is always False when status ∉ {CREATED, UPLOADING}
> 
> An example of where separating tests out into more functions makes the tests less clean is when you're testing all the different `invalid_states` when checking for expiry. Moving each of those out to separate tests would make the tests harder to read since you'd be introducing a bunch of clutter.

I see, I prioritized avoiding code repetition. It makes sense to split them up so that it's easier to read and pinpoint what went wrong in failing tests. Thanks for the advice on test structure.

app/recordtransfer/tests/unit/test_models.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/402/commits/e2e3b7537e0117e1829e948061f1dcea8d7c08d3

app/recordtransfer/tests/unit/test_models.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/402/commits/e2e3b7537e0117e1829e948061f1dcea8d7c08d3

app/recordtransfer/tests/unit/test_models.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/402/commits/e2e3b7537e0117e1829e948061f1dcea8d7c08d3

app/recordtransfer/models.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/402/commits/764dd18a4ac8ae5c37554ba3dc147e12ef192528

app/recordtransfer/models.py
```diff
@@ -149,6 +150,24 @@ def file_count(self) -> int:
...
 
+    @property
+    def expires_at(self) -> Optional[timezone.datetime]:
+        """Get the time at which this session will expire. Only sessions in the CREATED or
+        UPLOADING state can have an expiration time. Returns None for sessions in other states.
+        """
+        if self.status in (self.SessionStatus.CREATED, self.SessionStatus.UPLOADING):
+            return self.last_upload_interaction_time + timezone.timedelta(
+                minutes=settings.UPLOAD_SESSION_EXPIRE_AFTER_INACTIVE_MINUTES
+            )
```
danloveg, line 161: Didn't notice this before, but UPLOAD_SESSION_EXPIRE_AFTER_INACTIVE_MINUTES can be disabled by being set to `-1`. Can you handle that case here and add a test to handle that case as well?

danloveg (changes requested):
One last comment here, then looks we're good to go.

app/recordtransfer/models.py
yenaing-oo: That's a good spot! I've address those cases here:

https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/402/commits/960b727d93687cdce60bac4966e2c479d2944f79

yenaing-oo: This does make me think whether we should create a setting similar to `FILE_UPLOAD_ENABLED` to control whether we want to allow file upload sessions to be disabled/enabled.

danloveg: Good point. Let's leave this as-is for this PR and address that in a new issue I'll create

danloveg (approved):
Looks good. The tests look nice and clean now, and the expiry logic is easy to follow.

Issue #393, opened 2025-02-14
Add RQ Cron Job to Clean Expired Upload Sessions


danloveg:
Add an RQ-based cron job to clean expired upload sessions. This should remove all files associated with the session and delete the session object, too. Logs should be written to the `rq` logs output.

This should be a [cron job](https://github.com/rq/rq-scheduler?tab=readme-ov-file#cron-jobs) created when the app first starts up. If this job is started, that fact should be logged to the output. The specific schedule will be set by `UPLOAD_SESSION_EXPIRED_CLEANUP_SCHEDULE`.

This should only be scheduled when `UPLOAD_SESSION_EXPIRE_AFTER_INACTIVE_MINUTES >= 1` and `FILE_UPLOAD_ENABLED=True`.

This change may require including the RQ Scheduler dependency, since I'm not sure whether this is already included in `django-rq`.

danloveg:
Closed by #403

PR #403, opened 2025-02-19
Add RQ Cron Job to Clean Expired Upload Sessions

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/393


danloveg:
> @danloveg Having given it more thought, I think having duplicate logic to find expired upload sessions in the cron job might be the best option.
> 
> Having `expired` and `expires_at` as fields is not ideal, because these fields would have to be updated every time the user interacts with the upload session (addition/removal of a temp uploaded file). That's two additional fields that need to be updated (first one is `last_upload_interaction_time`). This seems like a worse trade-off just to be able to filter using an `expired` field in the cron job.

That makes sense. In that case, we should probably add a comment to each of those places saying that this logic is duplicated, and if it needs to be changed to also change it in the other place.

danloveg:
> > @danloveg Having given it more thought, I think having duplicate logic to find expired upload sessions in the cron job might be the best option.
> > Having `expired` and `expires_at` as fields is not ideal, because these fields would have to be updated every time the user interacts with the upload session (addition/removal of a temp uploaded file). That's two additional fields that need to be updated (first one is `last_upload_interaction_time`). This seems like a worse trade-off just to be able to filter using an `expired` field in the cron job.
> 
> That makes sense. In that case, we should probably add a comment to each of those places saying that this logic is duplicated, and if it needs to be changed to also change it in the other place.

Thinking about this more - the correct place for this logic is probably in a new custom object manager.

```python
# recordtransfer/managers.py
from django.apps import apps

class UploadSessionManager(models.Manager):
    def get_expired(self):
        if settings.UPLOAD_SESSION_EXPIRE_AFTER_INACTIVE_MINUTES == -1:
            return self.none()

        cutoff_time = timezone.now() - timezone.timedelta(
            minutes=settings.UPLOAD_SESSION_EXPIRE_AFTER_INACTIVE_MINUTES
        )

        # Avoiding circular import
        UploadSession = apps.get_model(
            app_label="recordtransfer",
            model_name="UploadSession",
        )

        return self.filter(
            last_upload_interaction_time__lt=cutoff_time,
            status__in=[
            ],
        )

# recordtransfer/models.py
    # ...

    objects = UploadSessionManager

    # ...
```

Then you could get all expired sessions like this: `UploadSession.objects.get_expired().all()`

yenaing-oo:
> > > @danloveg Having given it more thought, I think having duplicate logic to find expired upload sessions in the cron job might be the best option.
> > > Having `expired` and `expires_at` as fields is not ideal, because these fields would have to be updated every time the user interacts with the upload session (addition/removal of a temp uploaded file). That's two additional fields that need to be updated (first one is `last_upload_interaction_time`). This seems like a worse trade-off just to be able to filter using an `expired` field in the cron job.
> > 
> > 
> > That makes sense. In that case, we should probably add a comment to each of those places saying that this logic is duplicated, and if it needs to be changed to also change it in the other place.
> 
> Thinking about this more - the correct place for this logic is probably in a new custom object manager.
> 
> ```python
> # recordtransfer/managers.py
> from django.apps import apps
> 
> class UploadSessionManager(models.Manager):
>     def get_expired(self):
>         if settings.UPLOAD_SESSION_EXPIRE_AFTER_INACTIVE_MINUTES == -1:
>             return self.none()
> 
>         cutoff_time = timezone.now() - timezone.timedelta(
>             minutes=settings.UPLOAD_SESSION_EXPIRE_AFTER_INACTIVE_MINUTES
>         )
> 
>         # Avoiding circular import
>         UploadSession = apps.get_model(
>             app_label="recordtransfer",
>             model_name="UploadSession",
>         )
> 
>         return self.filter(
>             last_upload_interaction_time__lt=cutoff_time,
>             status__in=[
>                 UploadSession.SessionStatus.CREATED,
>                 UploadSession.SessionStatus.UPLOADING,
>             ],
>         )
> 
> # recordtransfer/models.py
>     # ...
> 
>     objects = UploadSessionManager
> 
>     # ...
> ```
> 
> Then you could get all expired sessions like this: `UploadSession.objects.get_expired().all()`

That's a clean approach, thanks for the suggestion.

yenaing-oo:
@danloveg 
Looking at the Containerfile, I see that installation of node dependencies and building static assets are only required for the `app` container. Would it be a good idea to take these steps out of the `base` image and make a separate intermediate image so that the `rq` and `rq-scheduler` containers do not depend on them?

danloveg:
> @danloveg Looking at the Containerfile, I see that installation of node dependencies and building static assets are only required for the `app` container (correct me if I'm wrong). Would it be a good idea to take these steps out of the `base` image and make a separate intermediate image so that the `rq` and `rq-scheduler` containers do not depend on them?
> 
> This would make it so that we don't have to pass `WEBPACK_MODE` to the `rq` and `rq-scheduler` containers.

I don't think that's something we have to worry about. In both the development and production containers, the image gets built *once* and then it's reused for every container that needs it. Yes, in development there are some extraneous dependencies like `npm` and `node`; but what we lose in keeping unused dependencies, we gain the simplicity of using the same image for all three containers that need it. Keep in mind that in the case of the production image, your expectations are already met, since the `node` dependencies are not present in that final image. And the production image is the one we should really care about when it comes to reducing the number of dependencies.

Hope that makes sense. And regarding your concern about having to specify WEBPACK_MODE in the rq and rq-scheduler containers, maybe we can use this approach instead that hides how the container gets built in every container but the one that builds it:

https://stackoverflow.com/a/41597317

Our (dev) compose file would look like this:

```yaml
  app:
    build:
      context: .
      target: dev
      dockerfile: Containerfile
      args:
          WEBPACK_MODE: development
    image: secure-record-transfer-dev

  rq:
    image: secure-record-transfer-dev

  rq-scheduler:
    image: secure-record-transfer-dev
```

Though it looks like this may not work in certain docker versions. Would be worth testing out, though.

danloveg:
> Sorry, just added some commits because I noticed some things that were missing

No problem, I'll review at some point this morning.

yenaing-oo:
> ... the image gets built once ...
Ah I thought `dev` or `prod` was built once for every service that targets it due to the different environment variables that are passed to each service, but it looks like the image is able to change it's behaviour during run-time instead of just being a fixed image.

> Yes, in development there are some extraneous dependencies like `npm` and `node`; but what we lose in keeping unused dependencies, we gain the simplicity of using the same image for all three containers that need it.

I see, this makes a lot of sense.

> Keep in mind that in the case of the production image, your expectations are already met, since the `node` dependencies are not present in that final image. And the production image is the one we should really care about when it comes to reducing the number of dependencies.

That's true, it makes sense care more about the prod image. You're right that the prod image already meets our requirements. The only thing I notice is the `/dist` folder being copied for the non-app containers, but this is acceptable for the benefit of simplicity.

> Hope that makes sense.

Yes, thank you for the detailed explanation.

> And regarding your concern about having to specify WEBPACK_MODE in the rq and rq-scheduler containers, maybe we can use this approach instead that hides how the container gets built in every container but the one that builds it:
> 
> https://stackoverflow.com/a/41597317
> 
> Our (dev) compose file would look like this:
> 
> ```yaml
>   app:
>     build:
>       context: .
>       target: dev
>       dockerfile: Containerfile
>       args:
>           WEBPACK_MODE: development
>     image: secure-record-transfer-dev
> 
>   rq:
>     image: secure-record-transfer-dev
> 
>   rq-scheduler:
>     image: secure-record-transfer-dev
> ```
> 
> Though it looks like this may not work in certain docker versions. Would be worth testing out, though.

I tested it out and it looks like it works fine:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/403/commits/531665aaccebdb10351cb9179de3c33e040f1fd1

app/recordtransfer/admin.py
```diff
@@ -202,9 +202,29 @@ class UploadSessionAdmin(ReadOnlyAdmin):
...
+        "token",
+        linkify("user"),
+        "started_at",
+        "file_count",
+        "upload_size",
+        "status",
+        "last_upload_interaction_time",
+        "expired",
+        "expires_at",
+    ]
```
danloveg, line 227: I recommend re-ordering the fields so the change time appears right after the started at time. I also think renaming this field to "Last upload at" for the list display looks a bit cleaner.

```suggestion
    list_display = [
        "token",
        linkify("user"),
        "started_at",
        "last_upload_at",
        "file_count",
        "upload_size",
        "status",
        "expired",
        "expires_at",
    ]
    
    @admin.display(description="Last upload at")
    def last_upload_at(self, obj: UploadSession):
        return obj.last_upload_interaction_time
```

app/recordtransfer/admin.py
```diff
@@ -202,9 +202,29 @@ class UploadSessionAdmin(ReadOnlyAdmin):
...
```
danloveg, line 215: Grouping the date times together makes this easier to read.

Since this is a readonly admin, I think we're OK to include the shorter name for the last upload time like I suggested in the `list_display`.

```suggestion
    fields = [
        "token",
        linkify("user"),
        "started_at",
        "last_upload_at",
        "file_count",
        "upload_size",
        "status",
        "expired",
        "expires_at",
    ]
```

app/recordtransfer/jobs.py
```diff
@@ -83,3 +83,19 @@ def create_downloadable_bag(submission: Submission, user_triggered: User):
...
+@django_rq.job
+def cleanup_expired_sessions() -> None:
+    """Delete all UploadSession objects that have expired."""
+    LOGGER.info("Deleting expired upload sessions ...")
+    try:
+        expired_sessions = UploadSession.objects.get_expired().all()
+
+        expired_sessions.delete()
+
+        LOGGER.info("Deleted %d expired upload sessions", expired_sessions.count())
```
danloveg, line 97: Can we have a different message here when there are none?

compose.dev.yml
```diff
@@ -66,13 +67,26 @@ services:
...
+      - redis
+
   app:
-    build:
-      context: .
-      target: dev
-      dockerfile: Containerfile
-      args:
-          WEBPACK_MODE: development
+    image: secure-record-transfer-dev
```
danloveg, line 89: I think one of these still needs to have `build:` - otherwise Docker won't have a way to know what the `secure-record-transfer-dev` image is

compose.dev.yml
```diff
@@ -66,13 +67,26 @@ services:
...
+      - path: .dev.env
+        required: false
+    # See manage.py for debugger setup
+    environment:
+      - ENV=dev
+      - IS_RQ=yes
+      - WEBPACK_MODE=development
+      - DJANGO_SETTINGS_MODULE=app.settings.docker_dev
+    depends_on:
+      - redis
```
danloveg, line 86: I suggest changing IS_RQ to no, because `IS_RQ` controls what container does the migrations in the entrypoint. Additionally, when IS_RQ=yes, the message in the container is "Starting RQ worker(s)" (see arrow **1.** in the attached image).

To ensure migrations are done before this container starts, you can make it depend on `rq` instead of `redis`.

```suggestion
    environment:
      - ENV=dev
      - IS_RQ=no
      - WEBPACK_MODE=development
      - DJANGO_SETTINGS_MODULE=app.settings.docker_dev
    depends_on:
      - rq
```

Screenshot 2025-02-20 093302

danloveg (changes requested):

compose.dev.yml
danloveg: Looking at this again, the original log message is still being written. I wonder if there's a way to intercept that message so only our log handler gets used

app/app/settings/docker_dev.py
```diff
@@ -83,5 +83,10 @@
             "level": "INFO",
             "propagate": True,
         },
+        "rq.scheduler": {
+            "handlers": ["console"],
+            "level": "INFO",
+            "propagate": True,
+        },
```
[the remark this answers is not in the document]
danloveg: Looking at this again, the original log message is still being written. I wonder if there's a way to intercept that message so only our log handler gets used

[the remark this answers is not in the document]
yenaing-oo: > Looking at this again, the original log message is still being written. I wonder if there's a way to intercept that message so only our log handler gets used

This is because `rq_scheduler` has another sub-logger called `rq_scheduler.scheduler` which logs messages from within the `scheduler.py`, where the "Registering birth" message is being logged.

Original log call here: https://github.com/rq/rq-scheduler/blob/f7d5787c5f94b5517e209c612ef648f4bfc44f9e/rq_scheduler/scheduler.py#L62

Log handlers setup: 
https://github.com/rq/rq-scheduler/blob/f7d5787c5f94b5517e209c612ef648f4bfc44f9e/rq_scheduler/scripts/rqscheduler.py#L55
and here 
https://github.com/rq/rq-scheduler/blob/f7d5787c5f94b5517e209c612ef648f4bfc44f9e/rq_scheduler/utils.py#L33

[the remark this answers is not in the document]
yenaing-oo: I don't get why this still manages to show when I name the logger `rq.scheduler`:
image

I was under the impression that only our own custom log message would be displayed.

compose.dev.yml
yenaing-oo: I believe the `rq` container already has the `build:` defined, which can be used by the lower two containers `rq-scheduler` and `app`.

compose.dev.yml
yenaing-oo: That makes sense. Having `rq-scheduler` depend on `rq` for migrations is a neat solution.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/403/commits/9710fc3e5764e826d9faaca2ab8f03947197a74f

app/recordtransfer/jobs.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/403/commits/eae0853954ce567ca41369d321d2f69ca6723642

app/recordtransfer/admin.py
yenaing-oo: I agree with everything.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/403/commits/d6e1a7775637863e5d2fae36c914090a98fd7b03

app/recordtransfer/admin.py
yenaing-oo: I agree with everything.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/403/commits/d6e1a7775637863e5d2fae36c914090a98fd7b03

compose.dev.yml
danloveg: Missed that, sorry. You're right

app/app/settings/docker_dev.py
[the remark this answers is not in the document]
danloveg: I'm going to resolve this, I've found a way to get the logs working correctly and I'll start a new review for that

danloveg, line 90: This captures all the logs so it's only our logger who writes them.

```suggestion
        "rq_scheduler": {
            "handlers": ["console"],
            "level": "INFO",
            "propagate": False,
        },
        "rq_scheduler.scheduler": {
            "handlers": ["console"],
            "level": "INFO",
            "propagate": False,
        },
```

app/app/settings/docker_prod.py
```diff
@@ -78,6 +78,11 @@
             "level": "INFO",
             "propagate": True,
         },
+        "rq.scheduler": {
+            "handlers": ["console"],
+            "level": "INFO",
+            "propagate": True,
+        },
```
danloveg, line 85:
```suggestion
        "rq_scheduler": {
            "handlers": ["console"],
            "level": "INFO",
            "propagate": False,
        },
        "rq_scheduler.scheduler": {
            "handlers": ["console"],
            "level": "INFO",
            "propagate": False,
        },
```

app/recordtransfer/management/commands/schedule_jobs.py
```diff
@@ -0,0 +1,39 @@
+import logging
+
+import django_rq
+from django.conf import settings
+from django_rq.management.commands import rqscheduler
+
+from recordtransfer.jobs import cleanup_expired_sessions
+
+scheduler = django_rq.get_scheduler()
+LOGGER = logging.getLogger("rq.scheduler")
```
danloveg, line 10: Don't need a special logger here, the `recordtransfer` logger setup will capture this log.

```suggestion
LOGGER = logging.getLogger(__name__)
```

app/recordtransfer/jobs.py
```diff
@@ -9,7 +9,7 @@
 from django.core.files.base import ContentFile
 from django.utils import timezone
 
-from recordtransfer.models import Job, Submission, User
+from recordtransfer.models import Job, Submission, UploadSession, User
 from recordtransfer.utils import zip_directory
 
 LOGGER = logging.getLogger("rq.worker")
```
danloveg, line 15: A special logger is not needed here either.

```suggestion
LOGGER = logging.getLogger(__name__)
```

danloveg (changes requested):
Just some last changes to fix up the logging and then this should be good to go.

app/recordtransfer/management/commands/schedule_jobs.py
yenaing-oo: Just to make sure I understand this fully - is your solution to let the `recordtransfer` logger capture the logs we made, and let `rq_scheduler` and `rq_scheduler.scheduler` intercept the logs made from the `rq_scheduler` package?

If so:
- could you explain why we would want to intercept the logs made from the package in the first place, if they already show up without intercepting them?
- why are not following the same pattern for different loggers within the app such as in [`emails.py`](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/3f4f4972a752adabee19074f37da76b84abce8ac/app/recordtransfer/emails.py#L21), where the logger used is `rq.worker` and not `__name__` (which would resolve to the `recordtransfer` logger)

yenaing-oo: > could you explain why we would want to intercept the logs made from the package in the first place, if they already show up without intercepting them?
Ah I think this is because we want a consistent format for all log messages, such as the log level and the timestamp.

yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/403/commits/2beb16f2f1e62f77027debbadd860407e2f3b172

app/app/settings/docker_prod.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/403/commits/2beb16f2f1e62f77027debbadd860407e2f3b172

app/app/settings/docker_dev.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/403/commits/2beb16f2f1e62f77027debbadd860407e2f3b172

app/recordtransfer/management/commands/schedule_jobs.py
danloveg: > Just to make sure I understand this fully - is your solution to let the `recordtransfer` logger capture the logs we made, and let `rq_scheduler` and `rq_scheduler.scheduler` intercept the logs made from the `rq_scheduler` package?

Correct, yes.

>  * could you explain why we would want to intercept the logs made from the package in the first place, if they already show up without intercepting them?

This is to provide a uniform log output in the container and provide control over the logs. If we want to write all the logs to a file in the future, we can simply add another handler in our config.

>  * why are we not following the same pattern for different loggers within the app such as in [`emails.py`](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/3f4f4972a752adabee19074f37da76b84abce8ac/app/recordtransfer/emails.py#L21), where the logger used is `rq.worker` and not `__name__` (which would resolve to the `recordtransfer` logger)

This is because of a misunderstanding I made about how logging works when I first wrote this app. All loggers should ideally look like this: `LOGGER = logging.getLogger(__name__)`.

I understand now that each process gets its own logging setup defined by the `dictConfig`. My mistake came from thinking that I would have to log to `rq.worker` to get the logs to show up in the `rq` container. I know now that there's a separate `recordtransfer` logger that runs in each `app`, `rq`, and now `rq-scheduler`. So each of those apps can write logs to `recordtransfer` and they'll get written to the logs of their respective containers.

Similarly with the emails, I thought those would have to get logged to `rq.worker`, since the emails are asynchronous jobs. No need to fix all the instances of bad logging setup right now, but we may as well fix the one I identified right now because that logger gets called in your new `cleanup_expired_sessions` function.

yenaing-oo: Things are much clearer now - thank you very much for the detailed explanation!

danloveg: For sure, thanks for asking!

danloveg (approved):
Looks great!

Issue #394, opened 2025-02-14
Display Expiring Sessions to Users on Profile Page


danloveg:
Users should be aware of when a session might expire. To give the user feedback about a session expiring, add an "Expires at" column to the in-progress submission table on the profile page. Show the date+time the session will expire. If it's within `UPLOAD_SESSION_EXPIRING_REMINDER_MINUTES` minutes of expiring, make the text red. If it's expired, make the text red and ~~struck-through~~. If upload session expiry is disabled, file uploads are disabled, or the in-progress submission has no upload session, populate a `-` in each row of this column.

If a user tries to load an expired session, send an error message saying "this session has expired" or something similar. An expired session should only be able to be deleted.

When the user saves an in-progress submission with an upload session, in the "Transfer Saved Successfully" message, mention that this submission will expire at a certain time.

yenaing-oo:
TODO:

Make tests to test that:
- IP submissions with no upload sessions have "-" under their expiry date
- Expiry date is shown for IP submissions with upload sessions
- The expiry datetime is accurate based on UPLOAD_SESSION_EXPIRING_REMINDER_MINUTES
- Close to or expired IP submissions have their expiry dates shown in red
- Expired IP submissions have their expiry dates shown struck-through

- Saving of an IP submission with an upload session displays a success message containing the expiry date
- Saving of an IP submission without an upload session displays just the generic saved success message

danloveg:
Closed by #405

PR #405, opened 2025-02-20
Display Expiring Sessions on Profile Page

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/394

Change notes:
- `UploadSession`'s `last_upload_interaction_time` is now only added automatically on creation, and not updated upon every save. I think it makes more sense to update this manually only when a temp file gets added, rather than if we change any of UploadSession's fields and call save()
- Attempt to access/resume an expired in-progress submission will result in a redirect to a "In-progress submission has expired" page
- I moved the initialization of forms to be used in the `TransferFormWizard` from `urls.py` to `transfer.py`, and chose the right forms to use based on whether `settings.FILE_UPLOAD_ENABLED` was set to True. It was necessary for `TransferFormWizard` to have the class variable `form_list` initialized, for the test I wrote for it to run.
- Fixed the issue of the `UploadSession` not being associated with an `InProgressSubmission` upon saving on the Upload Files step due to the `session_token` being looked for only in the `past_data` and not in the `current_data` as well.

Tests for the following are included:
- Attempt to access  expired in-progress submission redirects to "In-progress submission has expired" page
- IP submissions with no upload sessions have "-" under their expiry date
- Expiry date is shown for IP submissions with upload sessions
- Close to or expired IP submissions have their expiry dates shown in red
- Expired IP submissions have their expiry dates shown struck-through
- Test for `testuntil_minutes` util function


yenaing-oo:
TODO:

Make tests to test that:

IP submissions with no upload sessions have "-" under their expiry date

Expiry date is shown for IP submissions with upload sessions

The expiry datetime is accurate based on UPLOAD_SESSION_EXPIRING_REMINDER_MINUTES

Close to or expired IP submissions have their expiry dates shown in red

Expired IP submissions have their expiry dates shown struck-through

Saving of an IP submission with an upload session displays a success message containing the expiry date

Saving of an IP submission without an upload session displays just the generic saved success message

yenaing-oo:
@danloveg 
I struggled to write tests for these:
- Saving of an IP submission with an upload session displays a success message containing the expiry date
- Saving of an IP submission without an upload session displays just the generic saved success message

as it involved creating a valid complex request object that the `TransferFormWizard` view would accept. Perhaps an E2E test would be more appropriate, however, the E2E test [test_review_step()](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/3f4f4972a752adabee19074f37da76b84abce8ac/app/recordtransfer/tests/e2e/test_e2e.py#L366) is still broken, so until that gets fixed, I can't think of another way to test the displayed messages.

danloveg:
> @danloveg I struggled to write tests for these:
> 
>  * Saving of an IP submission with an upload session displays a success message containing the expiry date
> 
>  * Saving of an IP submission without an upload session displays just the generic saved success message
> 
> as it involved creating a valid complex request object that the `TransferFormWizard` view would accept. Perhaps an E2E test would be more appropriate, however, the E2E test [test_review_step()](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/3f4f4972a752adabee19074f37da76b84abce8ac/app/recordtransfer/tests/e2e/test_e2e.py#L366) is still broken, so until that gets fixed, I can't think of another way to test the displayed messages.

No problem. Figuring out how to test the TransferFormWizard is something we can figure out in a later issue, I kind of knew that that'd be a bit difficult to test.

I'm going to review this pull request now so keep an eye out for my feedback

app/recordtransfer/views/profile.py
```diff
@@ -52,6 +54,17 @@ def get_context_data(self, **kwargs) -> dict[str, Any]:
...
+        for submission in in_progress_submissions:
+            if submission.upload_session and submission.upload_session.expires_at:
+                minutes_until_expiry = timeuntil_minutes(submission.upload_session.expires_at)
+                # Dynamically add expiring_soon attribute to object as context for template
+                submission.expiring_soon = (  # type: ignore
+                    minutes_until_expiry <= settings.UPLOAD_SESSION_EXPIRING_REMINDER_MINUTES
+                )
+            else:
+                submission.expiring_soon = False  # type: ignore
+
```
danloveg, line 67: I'd recommend moving this into the upload session model, "expiring soon" is a useful function that should be implemented on the object itself.

```python
    # ...

    @property
    def expires_soon(self) -> bool:
        """Determine if this session will expire within the set expiration reminder time."""
        threshold = settings.UPLOAD_SESSION_EXPIRING_REMINDER_MINUTES
        if threshold == -1:
            threshold = 480
        return self.expires_within(minutes=threshold)

    def expires_within(self, minutes: int) -> bool:
        """Determine if this session will expire within the given number of minutes.

        Args:
            minutes: The number of minutes in the future to check for expiration.

        Returns:
            True if the session will expire before the given number of minutes from now, False
            otherwise.
        """
        expiry_time = self.expires_at
        if expiry_time is None:
            return False
        return expiry_time < timezone.now() + timezone.timedelta(minutes=minutes)

    # ...
```

If you create these methods, you don't need to attach any new attributes to the in progress submission.

Also, I should mention that [annotate()](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#django.db.models.query.QuerySet.annotate) would be the typically correct way to do this kind of operation, but since `expires_at` is not known in the database annotation would be messy and probably *not* the right call for this operation. This is just so you're aware of the annotate method.

app/recordtransfer/views/transfer.py
```diff
@@ -169,6 +212,13 @@ def dispatch(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
         if not self.in_progress_submission:
             return redirect("recordtransfer:transfer")
 
+        # Check if associated upload session is expired or not
+        if (
+            self.in_progress_submission.upload_session
+            and self.in_progress_submission.upload_session.expired
+        ):
+            return redirect("recordtransfer:in_progress_submission_expired")
```
danloveg, line 220: This logical statement can be moved to the In-progress submission model. I'm also including an `upload_session_expires_soon` function here because it can be used in the template.

```python
class InProgressSubmission(models.Model):
    # ...

    def upload_session_expired(self) -> bool:
        """Determine if the associated upload session has expired or not."""
        return self.upload_session is not None and self.upload_session.expired

    def upload_session_expires_soon(self) -> bool:
        """Determine if the associated upload session is expiring soon or not."""
        return self.upload_session is not None and self.upload_session.expires_soon

    # ...
```

```suggestion
        # Check if associated upload session is expired or not
        if self.in_progress_submission.upload_session_expired():
            return redirect("recordtransfer:in_progress_submission_expired")
```

app/recordtransfer/utils.py
```diff
@@ -358,3 +360,20 @@ def accept_session(filename: str, filesize: Union[str, int], session: "UploadSes
...
+
+    Returns:
+        The number of minutes until the datetime value, to the nearest minute. If the datetime
+        value is in the past, returns 0.
+    """
+    now = timezone.now()
+    if value < now:
+        return 0
+    delta = value - now
+    return int(delta.total_seconds() / 60)
```
danloveg, line 379: Won't be needed if the `expires_within` method is created on `UploadSession`.

app/recordtransfer/templates/includes/in_progress_submission_table.html
```diff
@@ -3,17 +3,29 @@
     <table class="full-width-table standard-table" cellspacing="0">
         <tr>
             <th>{% trans "Last Updated" %}</th>
+            <th>{% trans "Expires At" %}</th>
             <th>{% trans "Transfer Title" %}</th>
             <th>{% trans "Resume Transfer" %}</th>
             <th>{% trans "Delete" %}</th>
         </tr>
         {% for submission in in_progress_page_obj %}
```
danloveg, line 11: Can you change the name submission to `in_progress`? `submission` makes me think it's an actual `Submission` model.

app/recordtransfer/templates/includes/in_progress_submission_table.html
```diff
@@ -3,17 +3,29 @@
...
-                <td>{{ submission.title }}</td>
                 <td>
-                    <a href="{% url 'recordtransfer:transfer' %}?transfer_uuid={{ submission.uuid }}">Resume</a>
+                    {% if submission.upload_session.expires_at %}
+                        <span class="{% if submission.expiring_soon %}red-text{% endif %} {% if submission.upload_session.expired %}strikethrough{% endif %}">
+                            {{ submission.upload_session.expires_at|date:"D M j, Y" }} @ {{ submission.upload_session.expires_at|time:"H:i" }}
+                        </span>
+                    {% else %}
+                        -
+                    {% endif %}
```
danloveg, line 21:
```suggestion
                    {% if in_progress.upload_session_expires_at %}
                        <span class="
                            {% if in_progress.upload_session_expires_soon %}red-text{% endif %}
                            {% if in_progress.upload_session_expired %}strikethrough{% endif %}
                            ">
                            {{ in_progress.upload_session_expires_at|date:"D M j, Y" }} @
                            {{ in_progress.upload_session_expires_at|time:"H:i" }}
                        </span>
                    {% else %}
                        -
                    {% endif %}
```

app/recordtransfer/views/transfer.py
```diff
@@ -195,7 +245,20 @@ def post(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
...
+                and self.in_progress_submission.upload_session.expires_at
+            ):
+                expires_at_formatted = (
+                    self.in_progress_submission.upload_session.expires_at.strftime(
+                        "%a %b %d, %Y @ %H:%M"
+                    )
+                )
+                message += f" Submission will expire on {expires_at_formatted}"
+
+            messages.success(request, gettext(message))
```
danloveg, line 261: A few things here:

1. `gettext` works best if you give it a non-dynamic string. See [Django's documentation on translations](https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#internationalization-in-python-code):

> *"The caveat with using variables or computed values, as in the previous two examples, is that Django’s translation-string-detecting utility, [django-admin makemessages](https://docs.djangoproject.com/en/4.2/ref/django-admin/#django-admin-makemessages), won’t be able to find these strings"*

2. The if block can be cleaned up a little bit with an [assignment expression](https://peps.python.org/pep-0572/).
3. Similar to point 1, you should use a placeholder for the date which will make `gettext` work better.
4. I added more context to the expiry message so the user knows why they're being notified of that.

```suggestion
            message = gettext("Transfer saved successfully.")
            if (
                expires_at := self.in_progress_submission.upload_session_expires_at()
                if self.in_progress_submission
                else None
            ):
                expiry_message = gettext(
                    "This submission will expire on %(date)s. Please be sure to complete your "
                    "submission before then."
                ) % {"date": expires_at.strftime(r"%a %b %d, %Y @ %H:%M")}
                message = message + " " + expiry_message

            messages.success(request, message)
```

app/recordtransfer/templates/includes/in_progress_submission_table.html
```diff
@@ -3,17 +3,29 @@
...
                 </td>
+                <td>{{ submission.title }}</td>
+                <td>
+                    {% if not submission.upload_session.expired %}
+                        <a href="{% url 'recordtransfer:transfer' %}?transfer_uuid={{ submission.uuid }}">Resume</a>
+                    {% endif %}
+                    </td>
```
danloveg, line 28: A dash when the upload session has expired would be suitable here. Also, Resume needs to be translatable.

```suggestion
                <td>
                    {% if not submission.upload_session_expired %}
                        <a href="{% url 'recordtransfer:transfer' %}?transfer_uuid={{ submission.uuid }}">
                            {% trans "Resume" %}
                        </a>
                    {% else %}
                        -
                    {% endif %}
                </td>
```

danloveg (changes requested):
The profile page looks great, looks pretty much how I expected it to look. I also like how you moved the form list out of `urls.py` into the transfer form wizard itself, that makes a lot more sense to me.

I have several comments about the organization of the code. A lot of the expiry checking code can be moved into the `InProgressSubmission` and `UploadSession` models. I hope this doesn't mess with your tests too much.

app/recordtransfer/views/profile.py
yenaing-oo: I see, I've seen similar functionalities in other ORMs. It's good to know `annotate()` is the Django-equivalent.

I've extracted the mention functionalities into methods under `UploadSession`. 
Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/405/commits/a7d7bb9363851f4c318c0381f1d9e39493478c1d

app/recordtransfer/utils.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/405/commits/41b0830a9c660a20cf9dfb50490ccba35afd3dc0

app/recordtransfer/templates/includes/in_progress_submission_table.html
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/405/commits/a7d7bb9363851f4c318c0381f1d9e39493478c1d

app/recordtransfer/views/transfer.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/405/commits/53091b69f40a3dd335881735638193d02fa0af0a

Note: I accidentally included an email template meant for a different branch in this commit. I've removed that file here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/405/commits/40c3a4e3a744111c46ba22ac462ae387c1dc4b75

app/recordtransfer/templates/includes/in_progress_submission_table.html
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/405/commits/b58c4b748e549fd4e15577086879151716a4d992

app/recordtransfer/templates/includes/in_progress_submission_table.html
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/405/commits/1c464838ef3d0dabedb9a71e75a25b1c1a88fc57

app/recordtransfer/utils.py
danloveg, line 1: There are unused imports in this file that were being used for the `timeuntil_minutes` function - those can be removed

app/recordtransfer/views/transfer.py
```diff
@@ -195,7 +242,19 @@ def post(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
 
         try:
             self.save_form_data(request)
-            messages.success(request, gettext("Transfer saved successfully."))
+            message = gettext("Transfer saved successfully.")
+            if (
+                expires_at := self.in_progress_submission.upload_session_expires_at
+                if self.in_progress_submission.upload_session
```
danloveg, line 248: `upload_session_expires_at` already checks if the upload session exists

```suggestion
                if self.in_progress_submission
```

danloveg (changes requested):
Two last things and then we're good to go!

app/recordtransfer/utils.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/405/commits/0fbd537ccbec4515b6cd146554892bab2def366d

danloveg (approved):

Issue #395, opened 2025-02-14
Remind Users of Expiring In-Progress Submissions


danloveg:
When a user saves an expire-able upload session, schedule an email to be sent to the user `UPLOAD_SESSION_EXPIRING_REMINDER_MINUTES` *before* the session expires. If they already submitted or if they deleted the in-progress submission, do not send the email when this job runs. It's not necessary to remove this email from the queue if either of those conditions are met - just skip sending the email at the time it's scheduled to run in that case.

danloveg:
With the new rq-scheduler dependency, it should be straightforward to [queue a job to run at a specific](https://github.com/rq/rq-scheduler?tab=readme-ov-file#scheduling-a-job) time now.

danloveg:
Closed by #409

PR #409, opened 2025-02-24
Remind Users of Expiring In-Progress Submissions

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/395

Change notes:
- Registered `group_uuid` and `transfer_uuid` as parameters on the `/transfer` endpoint
- Added a method `get_resume_url` on `InProgressSubmssion` to conveniently retrieve a URL to access and resume an in-progress submission


yenaing-oo:
@danloveg
Whenever you have the chance, could you take a quick look at my changes for any glaring issues before I write tests for the scheduler and job? I'm not sure how I'll be doing that just yet, but I'll figure it out.

danloveg (commented):
This makes sense so far, feel free to go ahead with writing the tests.

danloveg:
@yenaing-oo I have a question before I submit my review. Do you think we should do anything when the session's last interaction time gets changed but the session doesn't explicitly get saved? The way this works, the email is only scheduled when the user clicks the save link. But theoretically someone could do this:

- Save the form
- Load the form, delete a file
- Go back to their profile *without* saving

`touch()` would have been called on the session, but the session wasn't explicitly saved. It *did* change though, because `touch()` would have been called when the file was deleted. This would make it so the email never gets sent since 'expires soon' will return False.

yenaing-oo:
> @yenaing-oo I have a question before I submit my review. Do you think we should do anything when the session's last interaction time gets changed but the session doesn't explicitly get saved? The way this works, the email is only scheduled when the user clicks the save link. But theoretically someone could do this:
> 
> * Save the form
> * Load the form, delete a file
> * Go back to their profile _without_ saving
> 
> `touch()` would have been called on the session, but the session wasn't explicitly saved. It _did_ change though, because `touch()` would have been called when the file was deleted. This would make it so the email never gets sent since 'expires soon' will return False.

This is an interesting edge case that I hadn't thought of. From the user's perspective, I would expect no change in expiry date of the in-progress submission if I left the form without saving the second time. Would it work if we called `touch()` exclusively when an `InProgressSubmission` gets saved (which we're already doing), and not when a file gets added/deleted to an `UploadSession`?

This way, when a user adds/deletes files and leaves the form without saving, no change will be made to the last upload interaction time and the expiry date remains the same, which means the user will end up receiving the initial email from the first save.

yenaing-oo:
> > > @yenaing-oo I have a question before I submit my review. Do you think we should do anything when the session's last interaction time gets changed but the session doesn't explicitly get saved? The way this works, the email is only scheduled when the user clicks the save link. But theoretically someone could do this:
> > > 
> > > * Save the form
> > > * Load the form, delete a file
> > > * Go back to their profile _without_ saving
> > > 
> > > `touch()` would have been called on the session, but the session wasn't explicitly saved. It _did_ change though, because `touch()` would have been called when the file was deleted. This would make it so the email never gets sent since 'expires soon' will return False.
> > 
> > 
> > This is an interesting edge case that I hadn't thought of. From the user's perspective, I would expect no change in expiry date of the in-progress submission if I left the form without saving the second time. Would it work if we called `touch()` exclusively when an `InProgressSubmission` gets saved (which we're already doing), and not when a file gets added/deleted to an `UploadSession`?
> > This way, when a user adds/deletes files and leaves the form without saving, no change will be made to the last upload interaction time and the expiry date remains the same, which means the user will end up receiving the initial email from the first save.
> 
> Yes I think that makes the most sense. It should be the saving of the form that updates that time, and probably not any other operation.
> 
> Want to make that change on this branch?

Yes, I can do that

danloveg:
> > Yes I think that makes the most sense. It should be the saving of the form that updates that time, and probably not any other operation.
> > Want to make that change on this branch?
> 
> Yes, I can do that

I just had a thought, if we're *not* uploading the last upload interaction time when a file gets uploaded, that's kind of defeating the purpose of that field.

Do you think it might make more sense to have an email *watcher* that watches for sessions expiring soon, rather than scheduling an email to be sent? I think our approach to scheduling the email is a bit too simplistic, and having a watcher might make this more robust in handling weird edge cases.

yenaing-oo:
> > > Yes I think that makes the most sense. It should be the saving of the form that updates that time, and probably not any other operation.
> > > Want to make that change on this branch?
> > 
> > 
> > Yes, I can do that
> 
> I just had a thought, if we're _not_ updating the last upload interaction time when a file gets uploaded, that's kind of defeating the purpose of that field.
> 
> Do you think it might make more sense to have an email _watcher_ that watches for sessions expiring soon, rather than scheduling an email to be sent? I think our approach to scheduling the email is a bit too simplistic, and having a watcher might make this more robust in handling weird edge cases.

Would it be like a cron job that periodically looks through all `InProgressSubmission`s to see if their associated `UploadSession` is expiring soon?

danloveg:
> > > > Yes I think that makes the most sense. It should be the saving of the form that updates that time, and probably not any other operation.
> > > > Want to make that change on this branch?
> > > 
> > > Yes, I can do that
> > 
> > I just had a thought, if we're _not_ updating the last upload interaction time when a file gets uploaded, that's kind of defeating the purpose of that field.
> > Do you think it might make more sense to have an email _watcher_ that watches for sessions expiring soon, rather than scheduling an email to be sent? I think our approach to scheduling the email is a bit too simplistic, and having a watcher might make this more robust in handling weird edge cases.
> 
> Would it be like a cron job that periodically looks through all `InProgressSubmission`s to see if their associated `UploadSession` is expiring soon?

Yes that's right. The way it would work is:

Every *t* time interval, get all in-progress submissions with an "expiring soon" upload session, and send an email to those users. Nice and simple.

> > I just had a thought, if we're _not_ updating the last upload interaction time when a file gets uploaded, that's kind of defeating the purpose of that field.
> 
> EDIT: The field `last_upload_interaction_time` would still be used to calculate `expires_at` and `expires_soon`. Perhaps we could rename it to something more fitting, like `last_interaction_time` instead.

Let's leave it as it is. I think the current logic with how last interaction time is updated is correct.

yenaing-oo:
@danloveg 
This PR is ready for review now, whenever you have the chance.

app/recordtransfer/tests/unit/test_managers.py
```diff
@@ -0,0 +1,157 @@
...
+    def setUp(self) -> None:
+        """Set up the test case."""
+        self.user = User.objects.create(username="testuser", password="password")
+        self.upload_session = UploadSession.new_session(user=self.user)
+
+    @patch("django.utils.timezone.now")
+    def test_get_expired(self, mock_now: MagicMock) -> None:
+        """Test when there are expired upload sessions."""
+        # Setup mock settings
+        settings.UPLOAD_SESSION_EXPIRE_AFTER_INACTIVE_MINUTES = 30
```
danloveg, line 23: You should `@patch()` this setting like in the other test modules

app/recordtransfer/tests/unit/test_managers.py
```diff
@@ -0,0 +1,157 @@
...
+            upload_session=self.upload_session,
+            current_step=TransferStep.UPLOAD_FILES.value,
+        )
+
+    @patch("django.utils.timezone.now")
+    def test_one_expiring(self, mock_now: MagicMock) -> None:
+        """Test when there is an expiring submission."""
+        # Setup mock settings
+        settings.UPLOAD_SESSION_EXPIRE_AFTER_INACTIVE_MINUTES = 30
+        settings.UPLOAD_SESSION_EXPIRING_REMINDER_MINUTES = 10
```
danloveg, line 92: Similar to my other suggestion, the settings should be `@patch()`-ed in this class

app/app/settings/base.py
```diff
@@ -278,3 +278,9 @@
     default="0 2 * * *",
     cast=str,
 )
+
+IN_PROGRESS_SUBMISSION_EXPIRING_EMAIL_SCHEDULE = config(
+    "IN_PROGRESS_SUBMISSION_EXPIRING_EMAIL_SCHEDULE",
+    default="0 * * * *",
+    cast=str,
+)
```
danloveg, line 286: This is a good addition. Can you also document it in the settings docs?

app/recordtransfer/management/commands/schedule_jobs.py
```diff
@@ -23,12 +23,21 @@ def register_scheduled_jobs() -> None:
...
         LOGGER.info(
             "Scheduling cleanup job (schedule: %s)",
-            schedule,
+            cleanup_schedule,
+        )
+        scheduler.cron(cleanup_schedule, func=cleanup_expired_sessions, queue_name="default")
+
+        email_schedule = settings.IN_PROGRESS_SUBMISSION_EXPIRING_EMAIL_SCHEDULE
+        LOGGER.info(
+            "Scheduling expiring email job (schedule: %s)",
```
danloveg, line 35: I think this is a bit clearer:

```suggestion
            "Scheduling periodic job to email reminders about expiring upload sessions (schedule: %s)",
```

danloveg (changes requested):
The logic is simple to follow now, it's clear that there's a job that runs every so often to check if people need to be reminded about expiring sessions.

Had a few very minor comments but other than that the functionality looks great.

app/recordtransfer/management/commands/schedule_jobs.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/409/commits/705310eb08ac8fc8cf0358e722024c5dd3bdea4f

app/app/settings/base.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/409/commits/705310eb08ac8fc8cf0358e722024c5dd3bdea4f

app/recordtransfer/tests/unit/test_managers.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/409/commits/9fa0eeec2e1cd6d1409ce6f550a3455d49a19a1e

app/recordtransfer/tests/unit/test_managers.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/409/commits/9fa0eeec2e1cd6d1409ce6f550a3455d49a19a1e

docs/settings/index.rst
```diff
@@ -479,6 +479,36 @@ UPLOAD_SESSION_EXPIRED_CLEANUP_SCHEDULE
...
+    
+    *Cron schedule expression for sending reminder emails for in-progress submissions with expiring upload sessions*
+
+    .. table::
+
+        ============  =========
+        Default       Type
+        ============  =========
+        "0 2 * * *"   string
+        ============  =========
```
danloveg, line 496: Sphinx doesn't like the asterisks because it thinks they're the start of an emphasis (italic or bold). I'd suggest escaping them like this:

```suggestion
        ===============  =========
        Default          Type
        ===============  =========
        "0 \* \* \* \*"   string
        ===============  =========
```

docs/settings/index.rst
```diff
@@ -479,6 +479,36 @@ UPLOAD_SESSION_EXPIRED_CLEANUP_SCHEDULE
...
+
+    Sets the cron schedule expression for sending reminder emails for in-progress submissions with expiring upload sessions. Defaults to "0 * * * *" (runs every hour).
```
danloveg, line 498: Same goes here

```suggestion
    Sets the cron schedule expression for sending reminder emails for in-progress submissions with expiring upload sessions. Defaults to "0 \* \* \* \*" (runs every hour at minute zero).
```

danloveg (changes requested):

danloveg (approved):

Issue #396, opened 2025-02-14
Add Upload Session Creation Protection


danloveg:
Add protection to only allow the creation of new upload sessions when the user gets to the upload files step.

This could likely be done using a decorator on the view that checks the user's current step of the form.

danloveg:
Closed by #421

PR #421, opened 2025-02-25
Add Protection for File Upload Related Endpoints

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/396

### DIFF
--- a/app/recordtransfer/decorators.py
+++ b/app/recordtransfer/decorators.py
@@ -1,0 +1,31 @@
+import functools
+from typing import Callable
+
+from django.http import HttpRequest, JsonResponse
+from django.utils.translation import gettext
+
+from recordtransfer.enums import TransferStep
+
+
+def require_upload_step(view_func: Callable) -> Callable:
+    """Restricts access to views based on the current wizard step. Only allows access if the
+    request originates from the UPLOAD_FILES step of TransferFormWizard.
+    """
+
+    @functools.wraps(view_func)
+    def _wrapped_view(request: HttpRequest, *args, **kwargs):
+        # Check if request comes from the upload step
+        wizard_data = request.session.get("wizard_transfer_form_wizard", {})
+        current_step = wizard_data.get("step")
+
+        # Allow access only if we're on the upload_files step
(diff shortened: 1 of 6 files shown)


yenaing-oo:
@danloveg 
I added the mentioned route protection in the issue to not just the endpoint for creating a new upload session, but also for the other file upload related endpoints. I think this makes sense (at the moment), since we hitting the file-upload endpoints from the client side from only the Upload Files step page.

The uploaded files information that is presented on the Review step is served from the server side, so the added protection won't affect that.

yenaing-oo:
Also, do you think the decorator `require_upload_step` belongs in `media.py`? I'm not sure where else to put it.

danloveg:
> Also, do you think the decorator `require_upload_step` belongs in `media.py`? I'm not sure where else to put it.

Django puts theirs in a [decorators.py file](https://github.com/django/django/blob/main/django/contrib/auth/decorators.py), we could follow that example.

yenaing-oo:
> > Also, do you think the decorator `require_upload_step` belongs in `media.py`? I'm not sure where else to put it.
> 
> Django puts theirs in a [decorators.py file](https://github.com/django/django/blob/main/django/contrib/auth/decorators.py), we could follow that example.

Ok, that sounds good. I'll make that change.

app/recordtransfer/decorators.py
```diff
@@ -0,0 +1,31 @@
+import functools
```
danloveg, line 1: Can you add a docstring here, and register this module in the code section of the docs?

danloveg (changes requested):
Looks great, my only suggestion is that we add this file to the documentation.

app/recordtransfer/decorators.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/421/commits/d78f937ea1930a928cd0fd879b28a7174b566b16

danloveg (approved):

PR #423, opened 2025-02-25
Update Version to 2025.02.28

danloveg:
Update version strings for end-of-month February milestone completion.


yenaing-oo (approved):
Everything looks good here.

Issue #407, opened 2025-02-21
Fix failing E2E test for Review Step


yenaing-oo:
The [test_review_step()](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/3f4f4972a752adabee19074f37da76b84abce8ac/app/recordtransfer/tests/e2e/test_e2e.py#L366) test case is currently broken. It appears to pass when run on its own, but when run as part of its test class, it fails.

danloveg:
This makes me think there could be some operations in the previous tests that might be changing the starting state of the `test_review_step` test. Are the previous two tests for sure cleaning up after themselves?

PR #424, opened 2025-02-26
Fix Failing E2E Test for Review Step

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/407

### DIFF
--- a/app/recordtransfer/tests/e2e/test_e2e.py
+++ b/app/recordtransfer/tests/e2e/test_e2e.py
@@ -50,7 +50,7 @@ class TransferFormWizardTest(StaticLiveServerTestCase):
         TransferStep.RECORD_DESCRIPTION: {
             "section_title": get_section_title(TransferStep.RECORD_DESCRIPTION),
             "accession_title": "Test Accession Title",
-            "date_of_materials": "2021-01-01 - 2021-12-31",
+            "date_of_materials": "[ca. 2021-01-01 - 2021-01-31]",
             "language": "English",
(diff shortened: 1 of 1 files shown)


yenaing-oo:
I'm not sure why this happens, but when the test class is run as a whole, during the execution of `complete_record_rights_step()` in `test_review_step`, the different default right types are missing. See output of debugging prints [here](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/62d0fc6a097025ed7ca4e0eab789024f288b96f2/app/recordtransfer/tests/e2e/test_e2e.py#L297-L300), removing [these lines](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/62d0fc6a097025ed7ca4e0eab789024f288b96f2/app/recordtransfer/tests/e2e/test_e2e.py#L99C9-L101C55) in the `setUp()` before running.

The output should look like this:
image

I managed to fix the test by explicitly adding the "Copyright" `RightType` to the database in the `setUp`: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/62d0fc6a097025ed7ca4e0eab789024f288b96f2/app/recordtransfer/tests/e2e/test_e2e.py#L99C9-L101C55

danloveg (approved):
These changes line up with our conversation earlier today, glad that we figured out what was going wrong. I tested this locally and it's working on my end

Issue #352, opened 2025-02-04
Enable date range when USE_DATE_WIDGETS=False


danloveg:
When `USE_DATE_WIDGETS` is `False` (i.e., date widgets are disabled), only a single date can be added to the `date_of_materials`. This should be fixed so that a date *range* can be specified whether date widgets are enabled or disabled.

PR #429, opened 2025-02-26
Enable date range when USE_DATE_WIDGETS=False

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/352


danloveg (approved):
Works great!

PR #431, opened 2025-02-26
Upload Session Expiry Feature Final Changes

yenaing-oo:


yenaing-oo:
@danloveg 
I've discovered a bug in our logic for "expiring" an in-progress submission. After the cron job `cleanup_expired_sessions()` and deletes an expired upload session, the associated in-progress submission will no longer have an `upload_session`, which means this method in `InProgressSubmission` will return `False`:

```python
# in recordtransfer.models.InProgressSubmission
@property
    def upload_session_expired(self) -> bool:
        """Determine if the associated upload session has expired or not."""
        return self.upload_session is not None and self.upload_session.expired
```

This means that an already "expired" in-progress submission will become resumable (no longer expired) after its upload session gets deleted.

I'm thinking that to address this, we can move the expiry logic from the `UploadSession` to `InProgressSubmission`. `InProgressSubmission`s will expire instead of `UploadSession`s. The existing cron job can be modified to delete any `UploadSession`s that do not have an associated `InProgressSubmission` or `Submission` instead.

Or alternatively, we can set a new field, perhaps called `expired` on `InProgressSubmission` which is set to `False` by default. The cron job can be modified to look for active `InProgressSubmission`s with an expired upload session. If their upload session is expired, we set `expired` to `True` and we use this flag instead on the Profile page and Transfer Form. 

Let me know what you think about either approach, or if you think of a better one.

yenaing-oo:
> > @danloveg I've discovered a bug in our logic for "expiring" an in-progress submission. After the cron job `cleanup_expired_sessions()` and deletes an expired upload session, the associated in-progress submission will no longer have an `upload_session`, which means this method in `InProgressSubmission` will return `False`:
> > ```python
> > # in recordtransfer.models.InProgressSubmission
> > @property
> >     def upload_session_expired(self) -> bool:
> >         """Determine if the associated upload session has expired or not."""
> >         return self.upload_session is not None and self.upload_session.expired
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This means that an already "expired" in-progress submission will become resumable (no longer expired) after its upload session gets deleted.
> > I'm thinking that to address this, we can move the expiry logic from the `UploadSession` to `InProgressSubmission`. `InProgressSubmission`s will expire instead of `UploadSession`s. The existing cron job can be modified to delete any `UploadSession`s that do not have an associated `InProgressSubmission` or `Submission` instead.
> > Or alternatively, we can set a new field, perhaps called `expired` on `InProgressSubmission` which is set to `False` by default. The cron job can be modified to look for active `InProgressSubmission`s with an expired upload session. If their upload session is expired, we set `expired` to `True` and we use this flag instead on the Profile page and Transfer Form.
> > Let me know what you think about either approach, or if you think of a better one.
> 
> Another solution to consider is that we don't actually delete `UploadSession` objects, we just set their state to EXPIRED and delete the temporary files.
> 
> If we went that way, we should be checking for expiry like this:
> 
> ```python
>     # ...
> 
>     def expired(self) -> property:
>         return (
>             self.status == self.SessionStatus.EXPIRED
>             or (self.expires_at is not None and self.expires_at < timezone.now())
>         )
> 
>     # ...
> ```

Would that not eventually cause a large build-up of `UploadSession`s in the database, with no form of deleting them?

danloveg:
> > > @danloveg I've discovered a bug in our logic for "expiring" an in-progress submission. After the cron job `cleanup_expired_sessions()` and deletes an expired upload session, the associated in-progress submission will no longer have an `upload_session`, which means this method in `InProgressSubmission` will return `False`:
> > > ```python
> > > # in recordtransfer.models.InProgressSubmission
> > > @property
> > >     def upload_session_expired(self) -> bool:
> > >         """Determine if the associated upload session has expired or not."""
> > >         return self.upload_session is not None and self.upload_session.expired
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This means that an already "expired" in-progress submission will become resumable (no longer expired) after its upload session gets deleted.
> > > I'm thinking that to address this, we can move the expiry logic from the `UploadSession` to `InProgressSubmission`. `InProgressSubmission`s will expire instead of `UploadSession`s. The existing cron job can be modified to delete any `UploadSession`s that do not have an associated `InProgressSubmission` or `Submission` instead.
> > > Or alternatively, we can set a new field, perhaps called `expired` on `InProgressSubmission` which is set to `False` by default. The cron job can be modified to look for active `InProgressSubmission`s with an expired upload session. If their upload session is expired, we set `expired` to `True` and we use this flag instead on the Profile page and Transfer Form.
> > > Let me know what you think about either approach, or if you think of a better one.
> > 
> > 
> > Another solution to consider is that we don't actually delete `UploadSession` objects, we just set their state to EXPIRED and delete the temporary files.
> > If we went that way, we should be checking for expiry like this:
> > ```python
> > class UploadSession(models.Model):
> >     # ...
> > 
> >     def expired(self) -> property:
> >         return (
> >             self.status == self.SessionStatus.EXPIRED
> >             or (self.expires_at is not None and self.expires_at < timezone.now())
> >         )
> > 
> >     # ...
> > ```
> 
> Would that not eventually cause a large build-up of `UploadSession`s in the database, with no form of deleting them?
> 
> EDIT: I think this can be addressed by having a cron job delete any `UploadSession`s that are not associated with either a `Submission` or an `InProgressSubmission`.

A database can hold millions of records in a table and still work just fine, I think it's OK if there's a bit of a buildup.

For your edit, maybe the cron job could do this:

```
# (This is pseudo code)
for expired_session in all_expired_sessions:
    if expired_session is not attached to an inprogress submission:
        delete it
    else:
        delete all temp files
        set to EXPIRED
```

yenaing-oo:
Notable changes:
- [Rename expired method to is_expired, add new method to expire UploadSe…](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/431/commits/628b11668933e4e5258fdf2f47960eb5dc5b4232): I renamed this method so that there is less confusion with the new method `expire()`
- [Remove SessionStatus.DELETED](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/431/commits/796766c5f7870163570665f8159f2eda79789fa6): Since we are now actually deleting `UploadSession`s, I think this status is no longer needed
- [Rename get_expired to get_expirable](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/431/commits/9c2da3dfac2f21f1b25c524ad94143d1bbbabc5d): I believe this function name is more accurate, since we the upload sessions we are fetching don't actually have a status of EXPIRED yet
- [Modify job to keep upload sessions if associated with an InProgressSu…](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/431/commits/f6d77e099ee26440e04074ab9352e04e7e740242): This is the main change that retains an UploadSession if it is associated with an InProgressSubmission, instead of deleting it
- [Return 0 for file_count and upload_size of expired UploadSessions](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/431/commits/4b1a3b899afd4063536583642f07dcdce654d795) and [Fix display of file count and upload size in admin](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/431/commits/ed002d44e13e28f12a3d9dd920f4bfcd4100a859): I noticed that the admin for the UploadSession model raises an error if an UploadSession with a status of anything other than UPLOADING or CREATED exists. This originates from the `file_count` and `upload_size` methods, which only caters to the mentioned two statuses. I fixed this by returning 0 for sessions with EXPIRED status, since we know we remove their files before setting it to EXPIRED. Sessions with COPYING_IN_PROGRESS or REMOVING_IN_PROGRESS status will have `file_count` and `upload_size` show up as "n/a" in admin now.
- [Allow EXPIRED upload sessions to have an expires_at time](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/431/commits/4f55899b5618056f66444a76f2b3934b46d1dcaf): Since we still need to be able to display the expiry datetime of an `InProgressSubmission` with an upload session that has a status of EXPIRED on the Profile page, I've allowed `expires_at` to calculate this expiry time for sessions in the EXPIRED status as well.

app/recordtransfer/jobs.py
```diff
@@ -88,18 +89,35 @@ def create_downloadable_bag(submission: Submission, user_triggered: User):
...
+    with any InProgressSubmission objects are deleted, while those that are associated with
+    InProgressSubmission objects have their uploads removed and are expired.
+    """
+    LOGGER.info("Cleaning up upload sessions ...")
     try:
-        expired_sessions = UploadSession.objects.get_expired().all()
-
-        if expired_sessions.count() == 0:
-            LOGGER.info("No expired upload sessions to delete")
+        expirable_sessions: QuerySet[UploadSession] = UploadSession.objects.get_expirable().all()
```
danloveg, line 98: This will never return sessions with an EXPIRED status, but I think it should. I'm imagining this series of actions:

1. A session related to an InProgressSubmission expires.
2. That session's status is set to EXPIRED
3. Later, a user deletes the InProgressSubmission

The next time this job runs, it won't return this session, since it was connected to an InProgressSubmission at one point.

Since we're deleting upload sessions not linked to any upload session, I think that this hypothetical upload session should be returned at this time as a deletable object.

Maybe `UploadSessionManager` needs two different methods:

```python
def get_deletable(self):
    """Get expired upload sessions that can be deleted."""

def get_expirable(self):
    """Get expired upload sessions that can be set to EXPIRED."""
```

app/recordtransfer/models.py
```diff
@@ -82,7 +82,6 @@ class SessionStatus(models.TextChoices):
         STORED = "SD", _("All Files in Permanent Storage")
         COPYING_FAILED = "FD", _("Copying Failed")
         REMOVING_IN_PROGRESS = "RP", _("File Removal in Progress")
```
danloveg, line 84: The flowchart in the docstring is now out of date

danloveg (changes requested):

app/recordtransfer/jobs.py
yenaing-oo: > This will never return sessions with an EXPIRED status, but I think it should. I'm imagining this series of actions:
> 
> 1. A session related to an InProgressSubmission expires.
> 2. That session's status is set to EXPIRED
> 3. Later, a user deletes the InProgressSubmission
> 
> The next time this job runs, it won't return this session, since it was connected to an InProgressSubmission at one point.

That's a good catch.

> Maybe `UploadSessionManager` needs two different methods:
> 
> ```python
> def get_deletable(self):
>     """Get expired upload sessions that can be deleted."""
> 
> def get_expirable(self):
>     """Get expired upload sessions that can be set to EXPIRED."""
> ```

What do you think about modifying it to perform the functionality of both, like this?:

```python
def get_cleanable(self) -> query.QuerySet:
        """Get expired upload sessions that can be set to EXPIRED or deleted."""
        if settings.UPLOAD_SESSION_EXPIRE_AFTER_INACTIVE_MINUTES == -1:
            return self.none()

        cutoff_time = timezone.now() - timezone.timedelta(
            minutes=settings.UPLOAD_SESSION_EXPIRE_AFTER_INACTIVE_MINUTES
        )

        # Avoiding circular import
        UploadSession = apps.get_model(
            app_label="recordtransfer",
            model_name="UploadSession",
        )

        return self.filter(
            last_upload_interaction_time__lt=cutoff_time,
            status__in=[
                UploadSession.SessionStatus.EXPIRED,
            ],
        )
```
The proposed two separate methods would have a lot of shared code otherwise.

app/recordtransfer/models.py
yenaing-oo: Since I removed the `DELETED` status, I'm trying to think of how to best handle the UploadSession after `remove_uploads` is called on it. It makes sense to me to set the status to `CREATED` if it was `UPLOADING` before. But if it was `STORED`, would it make sense to set it to `EXPIRED`?

```python
def remove_uploads(self) -> None:
        """Remove all uploaded files associated with this session."""
        if self.status == self.SessionStatus.REMOVING_IN_PROGRESS:
            LOGGER.warning("File removal is already in progress for session %s", self.token)
            return
        elif self.status == self.SessionStatus.CREATED:
            LOGGER.warning("There are no uploaded files in the session %s to remove", self.token)
            return
        elif self.status == self.SessionStatus.EXPIRED:
            raise ValueError(
                f"Cannot remove uploaded files from session {self.token} because the session has "
                "expired."
            )
        elif self.status == self.SessionStatus.COPYING_IN_PROGRESS:
            raise ValueError(
                f"Cannot remove uploaded files from session {self.token} while copying files is "
                "in progress"
            )

        initial_status = self.status
        self.status = self.SessionStatus.REMOVING_IN_PROGRESS
        self.save()

        for f in chain(self.permuploadedfile_set.all(), self.tempuploadedfile_set.all()):
            f.remove()

        if initial_status == self.SessionStatus.UPLOADING:
            self.status = self.SessionStatus.CREATED
        elif initial_status == self.SessionStatus.STORED:
            self.status = self.SessionStatus.EXPIRED
        self.save()
```

image

danloveg: > Since I removed the `DELETED` status, I'm trying to think of how to best handle the UploadSession after `remove_uploads` is called on it. It makes sense to me to set the status to `CREATED` if it was `UPLOADING` before. But if it was `STORED`, would it make sense to set it to `EXPIRED`?

The simplest solution is to raise an exception in that method if the state is STORED. Those files are supposed to be "permanent." Let's worry about how to handle deleting permanent files in a different issue.

Maybe this method should be `remove_temp_uploads`

yenaing-oo: Addressed by:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/431/commits/d249af696f251b4ad8250dd8d46030bf3f526deb

app/recordtransfer/jobs.py
yenaing-oo: I think it would be cleaner and easier to test by having two separate methods, like you proposed. I've addressed that here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/431/commits/e928ab0860210066784ef32dc5874bf6f0fc3ba6 and https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/431/commits/0b726a3f887a4d5cf103ea4e308faaaed7f4ce68

app/recordtransfer/managers.py
```diff
@@ -76,6 +76,26 @@ def get_expired(self) -> query.QuerySet:
...
+        UploadSession = apps.get_model(
+            app_label="recordtransfer",
+            model_name="UploadSession",
+        )
+
+        return self.filter(
+            status__in=[
+            UploadSession.SessionStatus.EXPIRED,
+            ],
+            inprogresssubmission__isnull=True,
```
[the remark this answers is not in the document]
danloveg: Interesting, okay. I think this might be clearer if InProgressSubmission had a `OneToOneField` to the upload session, and an explicit related_name is set. What do you think about this?

```python
class InProgressSubmission(models.Model):
    # ...

    upload_session = models.OneToOneField(
        UploadSession, null=True, on_delete=models.SET_NULL, related_name="in_progress_submission"
    )

    # ...

class UploadSessionManager(models.Manager):
    # ...

    def get_deletable(self) -> query.QuerySet:
        # ...
        return self.filter(
            status__in=[
                UploadSession.SessionStatus.EXPIRED,
            ],
            in_progress_submission__isnull=True,
        )
```

I don't love all the magic Django does, and it can make things hard to read. At least in this case, we explicitly defined the related_name. I'm open to suggestions.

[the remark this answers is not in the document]
yenaing-oo: I don't mind this at all. Giving it a related name does make it a bit clearer on how it's performing the operation.

I've addressed that here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/431/commits/09d4d70f296a4b20a5bf4e0c372e9d6cf7423ac8

In my commit, I changed it so that I'm passing the status itself instead of a list containing it.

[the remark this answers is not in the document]
danloveg: Great, thanks. I do have one more comment on this PR, I'll put it in a real review in a second

app/recordtransfer/managers.py
```diff
@@ -53,8 +53,8 @@ def export_csv(
 class UploadSessionManager(models.Manager):
     """Custom manager for UploadSession model."""
 
-    def get_expired(self) -> query.QuerySet:
-        """Return all upload sessions that have expired."""
+    def get_expirable(self) -> query.QuerySet:
+        """Return all expired upload sessions that can be set to EXPIRED."""
```
danloveg, line 57: This function has some overlap with `get_deletable`. This one should *only* return upload sessions that can be expired. The way it works right now, it can return some that are expirable, and some that are deletable.

I've made a suggestion on the doc string for how this should work:

```suggestion
    def get_expirable(self) -> query.QuerySet:
        """Return all expired upload sessions that can be set to EXPIRED.

        A session can be expired if it has not been interacted with in at least
        UPLOAD_SESSION_EXPIRE_AFTER_INACTIVE_MINUTES minutes. Additionally, the session must be
        linked to an existing InProgressSubmission.

        If a session is *not* linked to an InProgressSubmission, and matches the same last
        interaction criteria above, then it will be returned by get_deletable instead.
        """
```

app/recordtransfer/managers.py
```diff
@@ -76,6 +76,24 @@ def get_expired(self) -> query.QuerySet:
             ],
         )
 
+    def get_deletable(self) -> query.QuerySet:
+        """Return all upload sessions that can be deleted. A session is deletable if it is EXPIRED
+        and not associated with any in-progress submissions.
+        """
```
danloveg, line 82: This function can be expanded to grab the deletable sessions that are being returned by get_expirable.

I've made a suggestion on the doc string for how this should work:

```suggestion
    def get_deletable(self) -> query.QuerySet:
        """Return all upload sessions that can be safely deleted.

        An upload session that can be safely deleted matches these criteria:

        - It has expired (either by checking the last_upload_interaction_time, or the status)
        - It is not linked to an in-progress submission

        If a session *is* linked to an InProgressSubmission, and matches the same last interaction
        criteria above, then it will be returned by get_expirable instead.
        """
```

The query is a bit more complex, since we would check whether something is expired, as in `get_expirable`, but we're also checking whether the status is EXPIRED. Here is how the query might look (I haven't tested this):

```python
return self.filter(
    (
        Q(last_upload_interaction_time__lt=cutoff_time)
        & Q(
            status__in=[
            ]
        )
    )
    | Q(status=UploadSession.SessionStatus.EXPIRED),
    in_progress_submission__isnull=True,
)
```

app/recordtransfer/jobs.py
```diff
@@ -88,18 +89,40 @@ def create_downloadable_bag(submission: Submission, user_triggered: User):
...
+                session.remove_temp_uploads()
+                session.expire()
+                expired += 1
+            else:
+                # Session is in limbo, delete it.
+                session.delete()
+                deleted += 1
+        for session in deletable_sessions:
+            session.delete()
+            deleted += 1
```
danloveg, line 118: If `get_deletable` and `get_expirable` both return completely separate upload sessions, then we can simplify this part of the job. I'm also recommending we move `remove_temp_uploads()` inside `expire()`.

```suggestion
        deleted = deletable_sessions.count()
        expired = expirable_sessions.count()
 
        for session in expirable_sessions:
            session.expire()

        for session in deletable_sessions:
            session.delete()
```

app/recordtransfer/models.py
```diff
@@ -202,6 +195,23 @@ def expires_within(self, minutes: int) -> bool:
...
+        """
+        if self.status not in (self.SessionStatus.CREATED, self.SessionStatus.UPLOADING):
+            raise ValueError(
+                f"Cannot mark session {self.token} as expired because the session status "
+                f"is {self.status} and not {self.SessionStatus.CREATED} or "
+                f"{self.SessionStatus.UPLOADING}"
+            )
+
+        self.status = self.SessionStatus.EXPIRED
+        self.save()
```
danloveg, line 213: We may as well move calling this function inside `expire` so that we never forget to remove temp uploads when expiring an upload session:

```suggestion
        self.remove_temp_uploads()
        self.status = self.SessionStatus.EXPIRED
        self.save()
```

danloveg (changes requested):
If you imagine a venn diagram with the upload sessions returned by `get_expirable` in one circle, and the upload sessions returned by `get_deletable` in the other circle, then there is going to be some overlap (with the way it's written right now).

My suggested changes here make it so that the sets of upload sessions returned by these methods are *disjoint*. This simplifies the cleanup job a lot: just expire everything returned by `get_expirable`, and just delete everything returned by `get_deletable`.

yenaing-oo:
> If you imagine a venn diagram with the upload sessions returned by `get_expirable` in one circle, and the upload sessions returned by `get_deletable` in the other circle, then there is going to be some overlap (with the way it's written right now).
> 
> My suggested changes here make it so that the sets of upload sessions returned by these methods are _disjoint_. This simplifies the cleanup job a lot: just expire everything returned by `get_expirable`, and just delete everything returned by `get_deletable`.

I agree the `get_expirable` method is also returning deletable sessions, which adds complexity to the processing in the cleanup job. Just to clarify, do you mean that there is an overlap between `get_expirable` and `get_deletable` concept-wise (they may both contain deletable sessions), or a literal overlap in the upload sessions that are being returned from both methods?

From what I understand, the latter case is not possible, since `get_expirable` is looking for sessions with the statuses CREATED or UPLOADING, whereas `get_deletable` looks for sessions with the status EXPIRED.

danloveg:
> > If you imagine a venn diagram with the upload sessions returned by `get_expirable` in one circle, and the upload sessions returned by `get_deletable` in the other circle, then there is going to be some overlap (with the way it's written right now).
> > My suggested changes here make it so that the sets of upload sessions returned by these methods are _disjoint_. This simplifies the cleanup job a lot: just expire everything returned by `get_expirable`, and just delete everything returned by `get_deletable`.
> 
> I agree the `get_expirable` method is also returning deletable sessions, which adds complexity to the processing in the cleanup job. Just to clarify, do you mean that there is an overlap between `get_expirable` and `get_deletable` concept-wise (they may both contain deletable sessions), or a literal overlap in the upload sessions that are being returned from both methods?
> 
> From what I understand, the latter case is not possible, since `get_expirable` is looking for sessions with the statuses CREATED or UPLOADING, whereas `get_deletable` looks for sessions with the status EXPIRED.

Good point, I meant concept-wise, not literally. The point I'm getting at is that there's overlap between what's expirable, and what's deletable. My suggestion is that the deletable sessions that are being returned by `get_expirable` should actually be returned by `get_deletable` instead.

app/recordtransfer/managers.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/431/commits/1a375334156ecf11c05f45b76004183d4f64a2fb

app/recordtransfer/managers.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/431/commits/1a375334156ecf11c05f45b76004183d4f64a2fb

app/recordtransfer/jobs.py
yenaing-oo: Addressed by: [Simplify session processing for cleanup](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/431/commits/7b9f429218a5d3b3dfa44c59d6599834d401a7c3)

yenaing-oo:
> @yenaing-oo this is good to go, but there's a failing test on the profile page (the same one that's failing on the other pull requests). Looks to be an issue with the date number being zero-padded; it's expecting "March 05" but is finding "March 5". Have you found a fix for that yet?

I did not notice this test was failing, thanks for pointing that out.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/431/commits/ce50d8d98f8c66bb993ce0f919a776782f67b8b9

Issue #426, opened 2025-02-26
Fix Outstanding ESLint Issues


danloveg:
All outstanding ESLint warnings and errors should be addressed and committed.

danloveg:
Closed by #435

PR #435, opened 2025-02-28
Fix Outstanding ESLint Issues

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/426


danloveg (approved):
I don't love how `currentFormStep` is a global but it's not a big deal right now.

Everything else looks great.

Issue #417, opened 2025-02-25
Remove the `submission` field from the Job model


danloveg:
Remove the `submission` field from the Job model.

This is currently only used in `SubmissionAdmin.change_form_view`. That method in the admin should be completely removed. This will make it so that there is no state kept that will track whether a BagIt bag has already been generated or not for a given Submission, but this is OK. I am hoping to create an overhaul for how BagIt bags are generated in the future, but it's dependent on these Job changes.

**This change requires a migration.**

Issue #416, opened 2025-02-25
Add a `message_log` field to the Job model


danloveg:
Add a new text field to the Job model for logs to be written to.

When a Job is started, an extra log handler should be created programmatically and attached to the LOGGER that is configured to append messages to the Job's `message_log`. In the admin, this message log should be rendered as a box with a black background and a monospace white font to make it look like output from a terminal. Before the job function exits, the extra logging handler should be removed from the logger.

**This issue requires a migration.**

danloveg:
Closed by #437

PR #438, opened 2025-03-03
Remove submission field from Job model

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/417


danloveg (changes requested):
The `submit_buttons_bottom` block in the `submission_change_form.html` should be updated to reflect the fact that there is no `has_generated_bag` context variable now.

yenaing-oo:
> The `submit_buttons_bottom` block in the `submission_change_form.html` should be updated to reflect the fact that there is no `has_generated_bag` context variable now.

I've addressed that here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/438/commits/d1e52e272ea9ffa22321820f5c6586ad6e61e4fc

I also renamed the migration for removing the `submission` field, since migration `045` was used in the previous sub-issue https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/416.

I believe both will get squashed into one at the end.

app/recordtransfer/migrations/0046_remove_job_submission.py
danloveg, line 1: Sorry about the confusion with the migrations, but you'll need to merge 415-job-overhaul into this branch and then rename this migration one more time.

That should be done *after* #439 gets merged into 415-job-overhaul, since that issue also has new migrations.

danloveg (changes requested):
One more change, but this should be done after #439 is merged into the target branch. The migrations are all dependent on each other which makes the ordering a bit tedious to manage.

danloveg:
> I believe both will get squashed into one at the end.

Yes, we'll squash these once all the migrations are made.

app/recordtransfer/migrations/0046_remove_job_submission.py
yenaing-oo: No worries at all, this all makes sense to me.

yenaing-oo: I've fixed the migrations here, so that the latest one is now `049`: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/438/commits/ea29d36586f460664b3a5757ca2cb238ed3692e3

danloveg: Perfect

danloveg (approved):
This must be a record for most closed pull requests in an hour. Looks good!

PR #443, opened 2025-03-05
Enforce ESLint Rules in Pull Requests

### LINKED ISSUE #425: Enforce ESLint Rules in Pull Requests
We should look to start locking down the Javascript style since we're at a place where most JS is in the proper style now. The simplest way to enforce JS style will be to disallow merging pull requests with a GitHub action unless ESLint notices no warnings or errors.

yenaing-oo:
Closes #425


.github/workflows/pr-lint-check.yml
```diff
@@ -0,0 +1,30 @@
+name: PR Lint Check
+
+on:
+  pull_request:
+    branches: [ 425-enforce-eslint-rules-in-pull-requests ]
```
danloveg, line 5: We can remove this branch from here now that we know it works

danloveg (changes requested):

.github/workflows/pr-lint-check.yml
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/443/commits/73f2c7b59285c503a900be3c83a63b56ade5ab56

PR #444, opened 2025-03-05
Overhaul Job Model

danloveg:
Re-work the Job model and make it more useful and generic.

- Makes Jobs searchable in the admin
- Adds a message log to jobs so that logs are exposed to the admin
- Removes the coupling to the Submission model
- Changes the attached file location for jobs to something more generic (`attachments`)
- Fixes a bug with generating large zip files in memory


danloveg:
@yenaing-oo I know you wrote most of this code, but it looks good from my end. If there's anything to fix feel free to flag those

yenaing-oo (approved):
Everything looks good to me.

PR #445, opened 2025-03-05
Don't zip files in memory

### LINKED ISSUE #420: Don't zip files in memory
The way `create_downloadable_bag` job currently works is that it writes the zipped file into memory using `BytesIO`. This should be changed so that the zip file gets written to the `TEMP_STORAGE_FOLDER` instead. The reason being, that a `MemoryError` may be raised if the size of the zip file exceeds the machine's memory.

yenaing-oo:
Closes #420

### DIFF
--- a/app/recordtransfer/jobs.py
+++ b/app/recordtransfer/jobs.py
@@ -3,11 +3,10 @@ import os.path
 import shutil
 import tempfile
 import zipfile
-from io import BytesIO
 
 import django_rq
 from django.conf import settings
-from django.core.files.base import ContentFile
+from django.core.files.base import File
 from django.db.models.query import QuerySet
 from django.utils import timezone
 
@@ -61,30 +60,31 @@ def create_downloadable_bag(submission: Submission, user_triggered: User) -> Non
                 new_job.save()
                 return
 
-        temp_zip_path = None
         try:
             os.makedirs(settings.TEMP_STORAGE_FOLDER, exist_ok=True)
-            temp_fd, temp_zip_path = tempfile.mkstemp(
+            with tempfile.TemporaryFile(
                 suffix=".zip", dir=settings.TEMP_STORAGE_FOLDER
-            )
-            os.close(temp_fd)
+            ) as temp_zip_file:
+                LOGGER.info(
+                    "Creating temporary zip file on disk at %s ...",
+                    f"{settings.TEMP_STORAGE_FOLDER}/{temp_zip_file.name}.zip",
+                )
+                zip_directory(
+                    submission.location,
+                    zipfile.ZipFile(temp_zip_file, "w", zipfile.ZIP_DEFLATED, False),
+                )
+                LOGGER.info("Zipped directory successfully")
 
-            LOGGER.info("Creating zip file on disk at %s ...", temp_zip_path)
-            with zipfile.ZipFile(temp_zip_path, "w", zipfile.ZIP_DEFLATED, False) as zipped_bag:
-                zip_directory(submission.location, zipped_bag)
-            LOGGER.info("Zipped directory successfully")
+                file_name = f"{user_triggered.username}-{submission.bag_name}.zip"
+                LOGGER.info("Saving zip file as %s ...", file_name)
+                new_job.attached_file.save(file_name, File(temp_zip_file), save=True)
+                LOGGER.info("Saved file successfully")
 
-            file_name = f"{user_triggered.username}-{submission.bag_name}.zip"
-            LOGGER.info("Saving zip file as %s ...", file_name)
-            with open(temp_zip_path, "rb") as f:
-                new_job.attached_file.save(file_name, ContentFile(f.read()), save=True)
-            LOGGER.info("Saved file successfully")
+                new_job.job_status = Job.JobStatus.COMPLETE
+                new_job.end_time = timezone.now()
+                new_job.save()
 
-            new_job.job_status = Job.JobStatus.COMPLETE
-            new_job.end_time = timezone.now()
-            new_job.save()
-
-            LOGGER.info("Downloadable bag created successfully")
+                LOGGER.info("Downloadable bag created successfully")
         except Exception as exc:
             new_job.job_status = Job.JobStatus.FAILED
             new_job.save()
@@ -94,10 +94,6 @@ def create_downloadable_bag(submission: Submission, user_triggered: User) -> Non
                 str(exc),
             )
         finally:
-            if temp_zip_path and os.path.exists(temp_zip_path):
-                os.unlink(temp_zip_path)
-                LOGGER.info("Removed temporary zip file from disk")
-
             if os.path.exists(submission.location):
                 LOGGER.info("Removing bag from disk after zip generation.")
                 shutil.rmtree(submission.location)
--- a/app/recordtransfer/jobs.py
+++ b/app/recordtransfer/jobs.py
@@ -1,6 +1,7 @@
 import logging
 import os.path
 import shutil
+import tempfile
 import zipfile
 from io import BytesIO
 
@@ -36,7 +37,6 @@ def create_downloadable_bag(submission: Submission, user_triggered: User) -> Non
         start_time=timezone.now(),
         user_triggered=user_triggered,
         job_status=Job.JobStatus.IN_PROGRESS,
-        submission=submission,
     )
     new_job.save()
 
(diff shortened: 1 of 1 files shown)


app/recordtransfer/jobs.py
```diff
@@ -61,18 +61,23 @@ def create_downloadable_bag(submission: Submission, user_triggered: User) -> Non
                 new_job.save()
                 return
 
-        zipf = None
+        temp_zip_path = None
```
danloveg, line 64: I'm going to suggest we leave the file open when calling `mkstemp`. This will require having to have a handle to the file in the `finally:` block.

```suggestion
        temp_zip_path = None
        temp_zip_fd = None
```

app/recordtransfer/jobs.py
```diff
@@ -61,18 +61,23 @@ def create_downloadable_bag(submission: Submission, user_triggered: User) -> Non
...
-            LOGGER.info("Zipping directory to an in-memory file ...")
-            zipf = BytesIO()
-            zipped_bag = zipfile.ZipFile(zipf, "w", zipfile.ZIP_DEFLATED, False)
-            zip_directory(submission.location, zipped_bag)
-            zipped_bag.close()
+            os.makedirs(settings.TEMP_STORAGE_FOLDER, exist_ok=True)
+            temp_fd, temp_zip_path = tempfile.mkstemp(
+                suffix=".zip", dir=settings.TEMP_STORAGE_FOLDER
+            )
+            os.close(temp_fd)
```
danloveg, line 70: You're closing the file and then re-opening it later, you may as well just keep it open and reuse that file handle later:

```suggestion
            temp_zip_fd, temp_zip_path = tempfile.mkstemp(
                suffix=".zip", dir=settings.TEMP_STORAGE_FOLDER
            )
```

app/recordtransfer/jobs.py
```diff
@@ -61,18 +61,23 @@ def create_downloadable_bag(submission: Submission, user_triggered: User) -> Non
...
+
+            LOGGER.info("Creating zip file on disk at %s ...", temp_zip_path)
+            with zipfile.ZipFile(temp_zip_path, "w", zipfile.ZIP_DEFLATED, False) as zipped_bag:
+                zip_directory(submission.location, zipped_bag)
```
danloveg, line 74:
```suggestion
            zip_directory(submission.location, temp_zip_fd)
```

app/recordtransfer/jobs.py
```diff
@@ -89,8 +94,10 @@ def create_downloadable_bag(submission: Submission, user_triggered: User) -> Non
                 str(exc),
             )
         finally:
-            if zipf is not None:
-                zipf.close()
+            if temp_zip_path and os.path.exists(temp_zip_path):
+                os.unlink(temp_zip_path)
+                LOGGER.info("Removed temporary zip file from disk")
```
danloveg, line 99:
```suggestion
            if temp_zip_fd is not None:
                temp_zip_fd.close()

            if temp_zip_path and os.path.exists(temp_zip_path):
                os.unlink(temp_zip_path)
                LOGGER.info("Removed temporary zip file from disk")
```

app/recordtransfer/jobs.py
```diff
@@ -61,18 +61,23 @@ def create_downloadable_bag(submission: Submission, user_triggered: User) -> Non
...
+            LOGGER.info("Creating zip file on disk at %s ...", temp_zip_path)
+            with zipfile.ZipFile(temp_zip_path, "w", zipfile.ZIP_DEFLATED, False) as zipped_bag:
+                zip_directory(submission.location, zipped_bag)
             LOGGER.info("Zipped directory successfully")
 
             file_name = f"{user_triggered.username}-{submission.bag_name}.zip"
             LOGGER.info("Saving zip file as %s ...", file_name)
-            new_job.attached_file.save(file_name, ContentFile(zipf.getvalue()), save=True)
+            with open(temp_zip_path, "rb") as f:
+                new_job.attached_file.save(file_name, ContentFile(f.read()), save=True)
```
danloveg, line 80: We still have the same memory issue here. The entirety of the file is read into memory with `read()`. Another issue is that you're opening the file again, whereas this is not needed.

Two things:

- `File` is the proper class to use here ([Django docs reference](https://docs.djangoproject.com/en/4.2/ref/files/file/#the-file-class))
- You can reuse `temp_zip_fd`, you just need to `seek()` back to the start of the file

```suggestion
            temp_zip_fd.seek(0)
            new_job.attached_file.save(file_name, File(temp_zip_fd), save=True)
```

danloveg (changes requested):
A few comments here. You're opening the temp zip file and closing it three separate times, whereas you can just leave the file open and keep reusing the same file handle

app/recordtransfer/jobs.py
yenaing-oo: Unfortunately, the `ZipFile` class cannot be instantiated with a file descriptor: https://docs.python.org/3/library/zipfile.html#zipfile-objects
I think we might have to pass the path here instead.

danloveg: A "file-like" object means it should accept a file handle as input. Does it throw an exception when you try to pass it the file handle?

yenaing-oo: I found an alternative class `TemporaryFile` within the `tempfile` module which returns a file-like object:
https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile

I'm passing this object directly into `ZipFile` instead here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/445/commits/c22094c163fd4a821d29877ca087d0c0385bc4da

It also handles deletion upon closing, which should be taken care of when being used as a context manager with `with`.

> Unlike [TemporaryFile()](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile), the user of [mkstemp()](https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp) is responsible for deleting the temporary file when done with it.
https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp

app/recordtransfer/jobs.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/445/commits/c22094c163fd4a821d29877ca087d0c0385bc4da

app/recordtransfer/jobs.py
yenaing-oo: This may not be necessary since my latest change uses a context manager that handles closing and deletion of the temp file.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/445/commits/c22094c163fd4a821d29877ca087d0c0385bc4da

app/recordtransfer/jobs.py
yenaing-oo: > We still have the same memory issue here. The entirety of the file is read into memory with `read()`. Another issue is that you're opening the file again, whereas this is not needed.
> 
> Two things:
> 
> * `File` is the proper class to use here ([Django docs reference](https://docs.djangoproject.com/en/4.2/ref/files/file/#the-file-class))
> * You can reuse `temp_zip_fd`, you just need to `seek()` back to the start of the file

That makes sense. I wonder whether `File` streams the contents of the file rather than read the entire file like I did. It's not mentioned anywhere in the docs explicitly, but I assume it does the former since I see it has a `chunks()` method that reads the file in chunks.

Since I'm using `tempfile.TemporaryFile`, which is a file-like object, I think I don't need to seek back to the start of the file before passing it to `File()`.

danloveg: > > We still have the same memory issue here. The entirety of the file is read into memory with `read()`. Another issue is that you're opening the file again, whereas this is not needed.
> > Two things:
> > 
> > * `File` is the proper class to use here ([Django docs reference](https://docs.djangoproject.com/en/4.2/ref/files/file/#the-file-class))
> > * You can reuse `temp_zip_fd`, you just need to `seek()` back to the start of the file
> 
> That makes sense. I wonder whether `File` streams the contents of the file rather than read the entire file like I did. It's not mentioned anywhere in the docs explicitly, but I assume it does the former since I see it has a `chunks()` method that reads the file in chunks.
> 
> Since I'm using `tempfile.TemporaryFile`, which is a file-like object, I think I don't need to seek back to the start of the file before passing it to `File()`.
> 
> Addressed by: [c22094c](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/commit/c22094c163fd4a821d29877ca087d0c0385bc4da)

I went down a bit of a rabbit hole, but you're absolutely right about both your last points.

- In [File](https://github.com/django/django/blob/8f942f1c1dbf4222c8ca48253f7959366ed1bb60/django/core/files/base.py#L75), you can see that it calls `chunks()` when the file is iterated over.

- In [FieldFile](https://github.com/django/django/blob/8f942f1c1dbf4222c8ca48253f7959366ed1bb60/django/db/models/fields/files.py#L96), the file is passed through to the storage in `save`.

- In [FileSystemStorage](https://github.com/django/django/blob/8f942f1c1dbf4222c8ca48253f7959366ed1bb60/django/core/files/storage/filesystem.py#L121), you can see it reads the file chunk-by-chunk in `_save`.

- And back in [File](https://github.com/django/django/blob/8f942f1c1dbf4222c8ca48253f7959366ed1bb60/django/core/files/base.py#L55), you can see when `chunks()` is called, a `seek(0)` is always called every time the file is read

yenaing-oo: That's perfect, thanks for looking into this!

danloveg (approved):
Using TemporaryFile instead of mkstemp makes this a lot easier to follow. Looks good!

PR #447, opened 2025-03-06
Constrain `date_of_materials` when editing Metadata object in admin

### LINKED ISSUE #354: Constrain `date_of_materials` when editing Metadata object in admin
The `date_of_materials` should be constrained to YYYY-MM-DD - YYYY-MM-DD the same way it is when added in the transfer form.

jQuery masking can be used for this since jQuery is included in Django's admin.

yenaing-oo:
Closes #354 

Change notes:
- Applies a date input mask on the `date_of_materials` field in the admin change form for the Metadata model
(the input mask for the admin is not as advanced as the mask used in the Record Description step of the transfer form, as [jQuery.mask](https://igorescobar.github.io/jQuery-Mask-Plugin/docs.html) does not support ranged patterns like [imask.js](https://imask.js.org/) does. For e.g. in imask.js, you can define a `M` pattern, and restrict it from 1 to 12. Whereas in jQuery.mask, each pattern character can only be used to restrict a single character)

- Applied same validation used for the `date_of_materials` field in the `RecordDescriptionForm`  in `MetadataForm`
- Included `date_is_approximate` field in both `MetadataForm` and `Metadata` model
- Formatting of approximate date is no longer done in `RecordDescriptionForm`, but rather in the `create_flat_representation` method of the `Metadata` model
- Modified the front-end HTML render of the `date_of_materials` and `date_is_approximate` fields for the admin `MetadataForm` to display what approximate date format to expect when `date_is_approximate` is set to True.
- Added tests for `caais.forms.MetadataForm`, removed date formatting tests for `recordtransfer.forms.RecordDescription`, added date formatting test for `create_flat_representation` method of `Metadata` model


yenaing-oo:
@danloveg 
1. In addition to masking, I think it would make sense to apply the same form validation for the `date_of_materials` field that is being  used in `recordtransfer.transfer_forms.RecordDescriptionForm` for `caais.forms.Metadata`. Since we want `caais` to be independent of `recordtransfer`, how do you suggest I handle this? Is code duplication acceptable in this case?
2. `recordtransfer.transfer_forms.RecordDescriptionForm` has a `date_is_approximate` field that is not reflected in the value of `date_of_materials` through formatting (i.e. if `date_is_approximate` is True, we don't see that `date_of_materials` is formatted as [ca. {date}]). Is this intentional in `Metadata`?

danloveg:
> 1. In addition to masking, I think it would make sense to apply the same form validation for the `date_of_materials` field that is being  used in `recordtransfer.transfer_forms.RecordDescriptionForm` for `caais.forms.Metadata`. Since we want `caais` to be independent of `recordtransfer`, how do you suggest we handle this? Is code duplication acceptable in this case?

Sorry for the late reply. Yes, code duplication is OK in this instance.

danloveg:
I noticed a bug I had written with how the accession identifier is being saved, so I fixed that. I'll review this now

app/caais/static/caais/js/admin/dateOfMaterials.js
```diff
@@ -0,0 +1,90 @@
...
+            options: {
+                clearIfNotMatch: false,
+                placeholder: "YYYY-MM-DD",
+                translation: {
+                    Y: {pattern: /[1-9]/},
+                }
+            }
+        },
+        dateRange: {
+            pattern: "Y000-00-00 - Y000-00-00",
```
danloveg, line 28: I have to hit space before I start typing the next date, is that intended?

danloveg (changes requested):

app/caais/static/caais/js/admin/dateOfMaterials.js
yenaing-oo: Yes, that is intentional. Do you think it would be better for the UX if they type in the next date directly instead of hitting space first? To me, a space is indicative of the user wanting to enter another date.

danloveg: It wasn't clear to me that I had to press space. Maybe that behaviour can be mentioned in the help text for that field?

yenaing-oo: That makes sense, I'll add some help text.

app/caais/forms.py
```diff
@@ -87,47 +91,153 @@ def __init__(self, *args, **kwargs):
...
+            )
+
+        def custom_dia_render(name, value, attrs=None, renderer=None):
+            original_html = original_dia_render(name, value, attrs, renderer)
+            return (
+                f'<div class="date-is-approximate-wrapper help">{original_html}</div>'
+            )
+
+        self.fields["date_of_materials"].widget.render = custom_dom_render
+        self.fields["date_is_approximate"].widget.render = custom_dia_render
```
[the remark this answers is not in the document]
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/447/commits/f716cb39651fcab33a7c02bf017f0247d51b1cc0

app/caais/static/caais/js/admin/dateOfMaterials.js
danloveg: Sure, that makes sense to me

app/caais/static/caais/css/base/metadataChangeForm.css
danloveg, line 1: The label floats over to the right when the screen's width is reduced:

Screenshot 2025-03-10 134439

The label should be kept snug to the right of the checkbox.

I believe this file is responsible for the styling of the date is approximate checkbox (I didn't do any digging to check).

I also noticed the flag gets unusually large in the source of materials section, but that's not related to this issue

danloveg (changes requested):
This looks a lot cleaner with the custom widgets. Just had one final issue I noticed then we should be good to merge this into the feature branch

app/caais/static/caais/css/base/metadataChangeForm.css
yenaing-oo: > The date is approximate label floats over to the right when the screen's width is reduced:

> The label should be kept snug to the right of the checkbox.
> 
> I believe this file is responsible for the styling of the date is approximate checkbox (I didn't do any digging to check).
> 

That's a good catch. I made the margin and padding on the div containing the label and checkbox responsive - I couldn't find the width size of the media query used for the Django admin site in Django's docs, but I think it 767px.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/447/commits/6020b1037b965a539840003c4df30f449d2edea1


> I also noticed the flag gets unusally large in the source of materials section, but that's not related to this issue
I'll open up an issue for this.

yenaing-oo: > I couldn't find the width size of the media query used for the Django admin site in Django's docs, but I think it 767px.

I found it in Django's source code:
https://github.com/django/django/blob/e44e8327d3d88d86895735c0e427102063ff5b55/django/contrib/admin/static/admin/css/responsive.css#L385C1-L387C28

danloveg: > > I couldn't find the width size of the media query used for the Django admin site in Django's docs, but I think it 767px.
> 
> I found it in Django's source code: https://github.com/django/django/blob/e44e8327d3d88d86895735c0e427102063ff5b55/django/contrib/admin/static/admin/css/responsive.css#L385C1-L387C28

Perfect!

danloveg (approved):

PR #448, opened 2025-03-06
Refactor date parsing into Metadata model

### LINKED ISSUE #355: Refactor date parsing into `Metadata` model
A new function should be added on the `caais.models.Metadata` model with this function signature:

```python

def parse_event_date_for_atom(self) -> tuple[str, datetime.date, datetime.date]:
    """Parses this metadata's date of materials into a three-tuple containing an event date for AtoM.

    If the date cannot be parsed, returns a three tuple containing:

    - CAAIS_UNKNOWN_DATE_TEXT
    - A date object representing CAAIS_UNKNOWN_START_DATE
    - A date object representing CAAIS_UNKNOWN_END_DATE

    Returns:
        A three-tuple containing the text representation of the date, the earliest date in the
        range, and the latest date in the range.
    """
```

This function should be used instead of the `EventDateParser` class.

yenaing-oo:
Closes #355

### DIFF
--- a/app/caais/test/unit/test_models.py
+++ b/app/caais/test/unit/test_models.py
@@ -607,7 +607,7 @@ class TestMetadata(TestCase):
         metadata = Metadata(
             repository="Repository",
             accession_title="Title",
-            date_of_materials="September 2023",
+            date_of_materials="2023-09-01 - 2023-09-30",
             rules_or_conventions="CAAIS v1.0",
             language_of_accession_record="en",
         )
@@ -622,7 +622,7 @@ class TestMetadata(TestCase):
         self.assertEqual(flat["title"], "Title")
         self.assertEqual(flat["acquisitionType"], "")
         self.assertEqual(flat["eventTypes"], "Creation")
-        self.assertEqual(flat["eventDates"], "September 2023")
+        self.assertEqual(flat["eventDates"], "2023-09-01 - 2023-09-30")
         self.assertEqual(flat["eventStartDates"], "2023-09-01")
         self.assertEqual(flat["eventEndDates"], "2023-09-30")
 
@@ -632,7 +632,7 @@ class TestMetadata(TestCase):
         metadata = Metadata(
             repository="Repository",
             accession_title="Title",
-            date_of_materials="June 2023",
+            date_of_materials="2023-06-01 - 2023-06-30",
             rules_or_conventions="CAAIS v1.0",
             language_of_accession_record="en",
         )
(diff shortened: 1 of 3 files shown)


yenaing-oo:
Blocked by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/447

yenaing-oo:
TODO: Write tests for flat metadata for versions of AtoM that support creation/event dates (not Atom 2.1). Look into existing flat metadata tests to see if they are comprehensive, for e.g. does `test_flatten_metadata_no_related_atom_2_1` check for all expected fields?

app/caais/test/unit/test_models.py
```diff
@@ -697,4 +781,129 @@ def test_flatten_metadata_no_related_atom_2_1(self):
...
+    def test_parse_event_date_for_atom_empty_date(self) -> None:
+        """Test parsing an empty date."""
+        metadata = Metadata(date_of_materials="")
+        metadata.save()
+
+        date_text, start_date, end_date = metadata.parse_event_date_for_atom()
+
+        self.assertEqual(date_text, "Unknown")
+        self.assertEqual(start_date, datetime.strptime("1900-01-01", "%Y-%m-%d").date())
+        self.assertEqual(end_date, datetime.strptime("1900-12-31", "%Y-%m-%d").date())
```
danloveg, line 796: You can construct dates using the `date` constructor:

```suggestion
        self.assertEqual(start_date, date(1900, 1, 1))
        self.assertEqual(end_date, date(1900, 12, 31))
```

You'll need `from datetime import date` at the top of the file.

This applies to this test and the rest of the tests in this class

app/caais/test/unit/test_models.py
```diff
@@ -308,7 +316,7 @@ def test_flatten_metadata_full_section_3_related_caais_1_0(self):
         for key in flat.keys():
             self.assertIn(key, ExportVersion.CAAIS_1_0.fieldnames)
 
-        self.assertEqual(flat["dateOfMaterials"], "Circa 2018")
+        self.assertEqual(flat["dateOfMaterials"], "March 2018")
```
danloveg, line 319: I think this should say "Circa March 2018"

danloveg (changes requested):

app/caais/test/unit/test_models.py
yenaing-oo: It shouldn't format the date since `date_is_approximate` is set to False on [line 272](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/c499615b5617f307f3e273820679eafcd4dbfe62/app/caais/test/unit/test_models.py#L274):

```
metadata = Metadata(
            date_of_materials="March 2018",
            date_is_approximate=False,
        )
```

app/caais/test/unit/test_models.py
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/448/commits/c499615b5617f307f3e273820679eafcd4dbfe62

app/caais/test/unit/test_models.py
danloveg: I missed that, yes, you're right

danloveg (approved):

Issue #450, opened 2025-03-07
Broken validation check for SITE_ID setting


yenaing-oo:
Upon attempting to reset the development database, I ran into this error in the logs of the `rq_workers` container:

```shell
2025-03-06 14:46:42 The above exception was the direct cause of the following exception:
2025-03-06 14:46:40 >> Starting RQ worker(s)
2025-03-06 14:46:41 Attaching debugger to port 8010
2025-03-06 14:46:42 
2025-03-06 14:46:42 Traceback (most recent call last):
2025-03-06 14:46:42   File "/opt/secure-record-transfer/app/manage.py", line 71, in <module>
2025-03-06 14:46:42     main()
2025-03-06 14:46:42   File "/opt/secure-record-transfer/app/manage.py", line 67, in main
2025-03-06 14:46:42     execute_from_command_line(sys.argv)
2025-03-06 14:46:42   File "/opt/secure-record-transfer/.venv/lib/python3.10/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
2025-03-06 14:46:42     utility.execute()
2025-03-06 14:46:42   File "/opt/secure-record-transfer/.venv/lib/python3.10/site-packages/django/core/management/__init__.py", line 416, in execute
2025-03-06 14:46:42     django.setup()
2025-03-06 14:46:42   File "/opt/secure-record-transfer/.venv/lib/python3.10/site-packages/django/__init__.py", line 24, in setup
2025-03-06 14:46:42     apps.populate(settings.INSTALLED_APPS)
2025-03-06 14:46:42   File "/opt/secure-record-transfer/.venv/lib/python3.10/site-packages/django/apps/registry.py", line 124, in populate
2025-03-06 14:46:42     app_config.ready()
2025-03-06 14:46:42   File "/opt/secure-record-transfer/app/recordtransfer/apps.py", line 376, in ready
2025-03-06 14:46:42     verify_site_id()
2025-03-06 14:46:42   File "/opt/secure-record-transfer/app/recordtransfer/apps.py", line 266, in verify_site_id
2025-03-06 14:46:42     if not Site.objects.filter(pk=site_id).exists():
2025-03-06 14:46:42   File "/opt/secure-record-transfer/.venv/lib/python3.10/site-packages/django/db/models/query.py", line 1241, in exists
2025-03-06 14:46:42     return self.query.has_results(using=self.db)
[...]

PR #451, opened 2025-03-07
Verify settings that require database initialization after migrations

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/450

Calls a command to verify settings that require database access **after** migrations are made in the entrypoint.

EDIT: To make things more consistent, all setting verification is moved to `verify_settings.py` and is called with this new command.


app/recordtransfer/management/commands/verify_settings.py
```diff
@@ -0,0 +1,28 @@
+import logging
+
+from django.conf import settings
+from django.contrib.sites.models import Site
+from django.core.exceptions import ImproperlyConfigured
+from django.core.management.base import BaseCommand
+
+LOGGER = logging.getLogger("recordtransfer")
```
danloveg, line 8:
```suggestion
LOGGER = logging.getLogger(__name__)
```

app/recordtransfer/management/commands/verify_settings.py
danloveg, line 1: At first glance, the name of this file makes me think it verifies *all* settings. I was going to suggest renaming the file to be more specific, but then I thought maybe we could just move all the setting verification into this file.

What do you think about making that change?

danloveg (changes requested):
This makes sense. The Site model needs to exist in the database before this validation function is run.

I'm wondering if it's worth moving all setting verification into this file. Then the code is all in one place.

app/recordtransfer/management/commands/verify_settings.py
yenaing-oo: I agree with this change, it makes things more consistent so there's less confusion on where to verify settings.

yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/451/commits/afc596d166a34c700abe060c0eee9f1e466c3d29

danloveg (approved):
Great, I like that the validation only gets called once now. Previously I think the verification was being run twice per container, now it just gets run once when the RQ container starts.

app/recordtransfer/management/commands/verify_settings.py
```diff
@@ -0,0 +1,390 @@
...
+from django.conf import settings
+from django.contrib.sites.models import Site
+from django.core.exceptions import ImproperlyConfigured
+from django.core.management.base import BaseCommand
+
+LOGGER = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """Verify application settings defined in settings/docker_dev.py."""
```
danloveg, line 15: Sorry, one last thing:

```suggestion
    """Verify application settings defined in the DJANGO_SETTINGS_MODULE."""
```

danloveg (changes requested):

app/recordtransfer/management/commands/verify_settings.py
yenaing-oo: This makes sense.

danloveg (approved):

PR #453, opened 2025-03-07
Upgrade Django to Version 4.2.20

danloveg:
Ran "poetry update" to update dependencies.


yenaing-oo (approved):
Builds without error, all tests pass

PR #455, opened 2025-03-10
Fix collectstatic being run twice

### LINKED ISSUE #454: collectstatic is Run Twice in Production
The `collectstatic` command is run twice when the production container is built, once in `rq-scheduler` and once in `app`. The entrypoint script should be updated so that this only happens in `app`.

We will likely need another environment variable like `IS_RQ_SCHEDULER` or something to skip running collectstatic if that environment variable is set. I wish we could use different entrypoints but I don't think that's possible using the same Containerfile for all three containers.

yenaing-oo:
Closes #454

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -9,14 +9,14 @@ if [ "$ENV" != 'dev' ]; then
 fi
 
 # Run database migrations in RQ container
-if [ "$IS_RQ" = 'yes' ]; then
+if [ "$SERVICE_NAME" = 'rq' ]; then
   echo ">> Running database migrations."
   python manage.py migrate --no-input
   echo ">> Verifying settings."
   python manage.py verify_settings
   echo ">> Starting RQ worker(s)"
 
-else
+elif [ "$SERVICE_NAME" = 'app' ]; then
   if [ "$ENV" != 'dev' ]; then
     echo ">> Collecting static files."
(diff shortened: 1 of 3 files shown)


docker/entrypoint.sh
```diff
@@ -9,14 +9,14 @@ if [ "$ENV" != 'dev' ]; then
...
-if [ "$IS_RQ" = 'yes' ]; then
+if [ "$SERVICE_NAME" = 'rq' ]; then
   echo ">> Running database migrations."
   python manage.py migrate --no-input
   echo ">> Verifying settings."
   python manage.py verify_settings
   echo ">> Starting RQ worker(s)"
 
-else
+elif [ "$SERVICE_NAME" = 'app' ]; then
```
danloveg, line 19: We may as well add a custom message for when the scheduler is starting up

```suggestion
elif [ "$SERVICE_NAME" = 'rq-scheduler' ]; then
  echo ">> Starting RQ scheduler"

elif [ "$SERVICE_NAME" = 'app' ]; then
```

danloveg (changes requested):
Using a service name as the environment variable is a *much* better solution than using a boolean environment variable, great suggestion!

docker/entrypoint.sh
yenaing-oo: That's a good idea.

danloveg (approved):

PR #457, opened 2025-03-10
Fix large flag in CAAIS Metadata admin change form on small screen size

### LINKED ISSUE #456: Large flag in CAAIS Metadata admin change form on small screen size
The country flag in the CAAIS Metadata change form becomes unproportionately large when the screen size is reduced:

Image

yenaing-oo:
Closes #456 

Wraps the original rendered HTML for the `CountrySelectField` in a div, so that they are displayed side-by-side. The issue before was that the flag was a direct child of a flexbox containing the select field and the flag image, which meant that on a small screen, the flag image gets wrapped onto a separate row, causing it to fill up the available width of the entire flexbox.

### DIFF
--- a/app/caais/forms.py
+++ b/app/caais/forms.py
@@ -3,7 +3,6 @@ from django.core.exceptions import ValidationError
 from django.utils.translation import gettext
 
 from django_countries.fields import CountryField
-from django_countries.widgets import CountrySelectWidget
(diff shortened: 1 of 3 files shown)


danloveg (approved):
That fixed it, nice use of the custom widget!

PR #460, opened 2025-03-11
Add hash to static files in production to allow cache busting

### LINKED ISSUE #458: Use contenthash in Static File Names in Production
To avoid browsers caching old static files in production, a [contenthash](https://webpack.js.org/guides/caching/#output-filenames) should be applied to static file names so that browsers do not cache a static file that has changed.

Take this scenario for example:

- A user loads the app, and `transferform.js` is cached in their browser
- The application gets re-deployed, with some major changes in the transfer form JS
- The user loads the app again, but the old `transferform.js` is cached, so it's loaded instead of fetched from the server

The most common way to address this issue is to add a hash to static file names. That way, the static file names change whenever their contents change.

Refer to this tutorial for details on implementation: https://saashammer.com/blog/load-webpack-hash-bundle-django/

yenaing-oo:
Closes #458

### DIFF
--- a/app/app/settings/base.py
+++ b/app/app/settings/base.py
@@ -148,6 +148,13 @@ STATICFILES_DIRS = [
     os.path.join(os.path.dirname(BASE_DIR), "dist"),
 ]
 
+# Storage settings
+# Default STORAGES from Django documentation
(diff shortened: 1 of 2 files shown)


danloveg (approved):
This is much simpler than I thought it would be, good find with the `ManifestStaticFilesStorage` class

PR #466, opened 2025-03-12
Freeze time in TestUserProfileVIew

### LINKED ISSUE #459: Make test_in_progress_submission_expires_at in TestUserProfileView deterministic
The mentioned test is flaky as it uses the actual system time rather than a mocked time.

yenaing-oo:
Closes #459


poetry.lock
danloveg, line 1: Django and a bunch of other dependencies were updated in this file. Any chance you could install only freezegun and not update the rest of the dependencies? If not, we can go forward with reviewing pull #453 to upgrade Django prior to merging this in. A lot of the changes I'm seeing are included in that PR already

danloveg (changes requested):

poetry.lock
yenaing-oo: > Any chance you could install only freezegun and not update the rest of the dependencies?

I tried to do this by calling `poetry add freezegun==^1.5.1` instead of `poetry lock`. Could you check if this works as expected? 

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/466/commits/a1556a8962853a8bdea3934613cf7287914ad0cd

danloveg: Looks like it worked yeah. freezegun was already installed as a dependency as some other library and the version matches up with 1.5.1 so all good

PR #480, opened 2025-03-13
Redirect to homepage when already logged in

### LINKED ISSUE #479: Redirect user to Home Page if Accessing Create Account and Log In page while Logged In
Currently, the user is able to access the Create Account and Log In pages while already logged in. The user should be redirected to the Home Page if they attempt to do so.

yenaing-oo:
Closes #479 

Redirects the user to the homepage if they attempt to access the Create Account or Login page while already logged in. This can be tested by accessing the following URLs in development before and after logging in:
- http://localhost:8000/accounts/login/
- http://localhost:8000/account/


yenaing-oo:
I see that the default paths provided by Django for authentication begin with the plural noun `accounts`:

https://docs.djangoproject.com/en/4.2/topics/auth/default/#using-the-views
> accounts/login/ [name='login']
> accounts/logout/ [name='logout']
> accounts/password_change/ [name='password_change']
> accounts/password_change/done/ [name='password_change_done']
> accounts/password_reset/ [name='password_reset']
> accounts/password_reset/done/ [name='password_reset_done']
> accounts/reset/<uidb64>/<token>/ [name='password_reset_confirm']
> accounts/reset/done/ [name='password_reset_complete']

This is inconsistent with the existing singular noun `account` paths defined in [`recordtransfer.urls.py`](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/e19fb30bbe568f87d7ae95fb02d117e4b0e7f643/app/recordtransfer/urls.py#L94C13-L121C15).

This inconsistency should be addressed in a separate issue.

danloveg (approved):

PR #488, opened 2025-03-14
Update admin setup

### LINKED ISSUE #469: Update section "Admin Set-up with Django Admin"
Users should be linked to the "Updating the Domain of Your Site" page instead of being told how to manually update the site's domain.

danloveg:
Simplifies the admin setup with all of the simplifications we've recently made to the startup process. Also adds information on how the NGINX configuration works.

Closes #469


docs/running/index.rst
```diff
@@ -19,30 +19,22 @@ Development Application
...
-From the root of the repository, run the following commands before running the application for the
-first time.
-
-.. code-block:: bash
-
-    cp example.dev.env .dev.env
-
-
-To start up the application, run this command:
+To start the application, run this command:
```
yenaing-oo, line 22: I think it would be good to still mention that this command should be run from the root of the repository:

```suggestion
To start the application, run this command from the root of the repository:
```

docs/running/index.rst
```diff
@@ -19,30 +19,22 @@ Development Application
...
 
 .. code-block:: bash
 
     podman-compose -f compose.dev.yml down
 
 
-Visit http://localhost:8000 to see the application running.
-
 If you are using `Podman Desktop <https://podman-desktop.io/>`_, you should now see the application
 and all of its services running there, too.
```
yenaing-oo, line 39: I think this should be moved below where "Visit http://localhost:8000 to see the application running." is.

docs/running/index.rst
```diff
@@ -19,30 +19,22 @@ Development Application
...
     podman-compose -f compose.dev.yml up -d
 
 
+Visit http://localhost:8000 to see the application running.
+
 To stop the application, run the :code:`down` command:
 
 .. code-block:: bash
 
     podman-compose -f compose.dev.yml down
```
yenaing-oo, line 35: Do you think it would be a good idea to move this part to the bottom of this section, or perhaps elsewhere? This part seems out of place to me, since everything else is about getting the app running.

yenaing-oo (commented):
Just had some minor suggestions in a couple of places. 

I see that instructions for stopping the application are mentioned in "Development Application". Would it make sense to do the same for "Production Application" as well?

EDIT: Opted to remove instructions for stopping app entirely.

docs/running/index.rst
danloveg: Agreed, resolved in 39846ba896cc450dfead9e7bf5f2c58ad60b19b3

docs/running/index.rst
danloveg: You're right, it interrupts the flow of this part of the documentation. I removed it in this commit: 39846ba896cc450dfead9e7bf5f2c58ad60b19b3

docs/running/index.rst
danloveg: Resolved in 39846ba896cc450dfead9e7bf5f2c58ad60b19b3

yenaing-oo (approved):
Looks good!

PR #500, opened 2025-03-21
Update version to 2025.03.21

danloveg:
Version update to coincide with the [March 2025](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/milestone/4) milestone.


yenaing-oo (approved):
Works as expected

Issue #482, opened 2025-03-13
Improve REST-ful Design of urls.py


danloveg:
There are some outstanding URLs that require some attention to make the application more REST-ful. See sub-issues for more information.

yenaing-oo:
Upon familiarising myself with REST-ful API practices, I've realised that the majority of the routes in recordtransfer.urls do not need to adhere to REST. Most of the routes are "front-end routes", meaning they return HTML rather than a standard data format such as JSON. Some of them are not stateless too, such as "submission/", which stores form data through the user's session.

The intention of these routes are to provide pages to the user rather than to perform actions on entities (such as queries or mutations). As such, I believe the naming of their paths do not have to adhere to REST-ful standards, and should aim to be intuitive to the user through abstraction. For instance, a path like "submission/new" would be fine even though REST-ful API practices dictate that verbs should be avoided in URIs to represent operations, since this path is not intended to be an API.

We do have some routes which are "eligible" to be REST-ful, such as a GET to "user/<int:user_id>/submission-group/" to fetch a list of a user's submission groups, although this is not REST-ful as its path name should be changed to describe the entity concisely where possible: "submission-groups/". The issue with making this change is that we have an existing path "submission-group/" that returns a page to create a new submission group when a GET request is received, and create a new submission group from form data when a POST is received. Since we still need a way to render HTML templates and display a UI page to create new groups, we cannot simply choose to remove the frontend route in favour of the API route. 

The solution would be to create a separate root path "api/" in order to make calls to API paths such as "api/submission-groups". 
This way, we could have frontend and API paths which do not conflict, and the two types of paths could be clearly distinguished. We could apply the same concept to other entities in paths within `urls.py` to have a fully REST-ful API. Having this would be nice, as it opens the possibility of non-browser apps and third-party clients using our API in the future. But since we have no intention to do any of that, and are simply focused on catering the routes to the needs of this web application, I believe creating REST-ful APIs for the desired entities in the database is unnecessary at this time.

yenaing-oo:
There are two places where the paths in could be cleaned up though, as agreed in our meeting:

- Change `accounts/` to `account/` in app.urls for consistency with existing `account/` paths in recordtrasfer.urls
- Change `DeleteInProgressSubmission` in pre_submmision.py to use the DELETE HTTP method to delete an in-progress submission rather than using POST

danloveg:
[...]

PR #502, opened 2025-04-03
Improve REST-ful Design of urls.py

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/482

Change notes:
- Refactored accounts/ to account/ in app.urls for consistency with existing account/ paths in recordtrasfer.urls
- Set the `LOGIN_URL` setting in base.py to set the URL to which unauthenticated users should be redirected to if attempting to access a page that requires login
- Changed DeleteInProgressSubmission in pre_submmision.py to accept the DELETE HTTP method to delete an in-progress submission rather than POST


webpack.config.js
```diff
@@ -91,6 +91,11 @@ module.exports = {
                 "recordtransfer/js/submission_group/*.js")
                 .map(file => "./" + path.relative(__dirname, file)),
         ],
+        in_progress_submission: [
+            ...glob.sync("./app/recordtransfer/static/" +
+                "recordtransfer/js/in_progress_submission/*.js")
+                .map(file => "./" + path.relative(__dirname, file)),
+        ],
```
danloveg, line 98: Is it possible to bundle this with the `profile` entry point? I'd like it if we can avoid making new Webpack entry points because they're adding complexity to the webpack configuration

danloveg (changes requested):

webpack.config.js
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/502/commits/e59cadf0329b7ddbb33f3b60ff472167dca3cc76

This makes sense to me. 

I would suggest bundling the `submission_detail` entry point into the `profile` entry point as well, but since I know that the Submission Detail report page is available to admin users as well, I think it's a good idea to leave keep the contents of `profile` separate from `submission_detail`.

danloveg:
Can you try and run the app in production and let me know if it works? I tried testing this with the production configuration and I'm getting the error

`Conflicting migrations detected; multiple leaf nodes in the migration graph: (0044_uploadsession_user, 0047_job_update in recordtransfer).`

Screenshot 2025-04-07 090505

This is causing a 500 error on the profile page

yenaing-oo:
> Can you try and run the app in production and let me know if it works? I tried testing this with the production configuration and I'm getting the error
> 
> `Conflicting migrations detected; multiple leaf nodes in the migration graph: (0044_uploadsession_user, 0047_job_update in recordtransfer).`
> 
> This is causing a 500 error on the profile page

I rebuilt and ran the app in production and didn't get any errors:
image

Perhaps the database needs to be reset?

danloveg (approved):
Everything looks good here. I especially like the test `test_delete_error` that you created, we should try to write more tests like that in the future since views can of course throw exceptions that we need to be sure are handled gracefully.

By the way, the issues I had were resolved by just doing a `down -v` and bringing the app up again. 100% unrelated to these changes.

PR #516, opened 2025-04-10
Update poetry installation info

### LINKED ISSUE #508: Update Poetry Installation Info
Two things:

- We should recommend using the Poetry official installer
- We should specify that Poetry version 1.8.5 needs to be installed

Poetry 2 is the newest version, but it changes the format of the poetry.lock file a great deal so we should stick with 1.8.5 for the time being.

danloveg:
Closes #508 

Clean up Poetry installation docs


yenaing-oo (approved):
I see that you removed the instructions for activating a virtual environment on Windows in favor of `source env/bin/activate` alone. This makes sense, if we recommend that development on Windows is carried out via WSL.

I also like how you cleaned up the dependency installation section so that the "dev" extras is included by default.

PR #530, opened 2025-04-28
Dynamically Render Table Pagination Using HTMX

### LINKED ISSUE #529: Dynamically Render Table Pagination on Profile Page
Implement dynamic pagination for tables on the profile page to avoid full page reloads when navigating between pages.

yenaing-oo:
Closes #529

### DIFF
--- a/app/recordtransfer/templates/includes/submission_table.html
+++ b/app/recordtransfer/templates/includes/submission_table.html
@@ -23,7 +23,7 @@
         </tr>
     {% empty %}
     <tr>
-        <td colspan="5" class="empty-table-text">{% trans "You have not sent any records to the NCTR." %}</td>
+        <td colspan="6" class="empty-table-text">{% trans "You have not sent any records to the NCTR." %}</td>
     </tr>
     {% endfor %}
 </table>
--- a/app/recordtransfer/templates/includes/submission_table.html
+++ b/app/recordtransfer/templates/includes/submission_table.html
@@ -1,27 +1,32 @@
 {% load i18n %}
-<div class="flex-full-width-item">
-    <table class="full-width-table standard-table" cellspacing="0">
+<table class="full-width-table standard-table margin-bottom-25px" cellspacing="0">
+    <tr>
+        <th class="uuid-column">{% trans "Submission ID" %}</th>
+        <th>{% trans "Date Submitted" %}</th>
+        <th>{% trans "Submission Title" %}</th>
+        <th>{% trans "Files Submitted" %}</th>
+        <th>{% trans "View Report" %}</th>
+        <th>{% trans "Download CSV" %}</th>
+    </tr>
+    {% for submission in page %}
         <tr>
-            <th class="uuid-column">{% trans "Submission ID" %}</th>
-            <th>{% trans "Date Submitted" %}</th>
-            <th>{% trans "Submission Title" %}</th>
-            <th>{% trans "Files Submitted" %}</th>
-            <th>{% trans "View Report" %}</th>
-            <th>{% trans "Download CSV" %}</th>
+            <td class="uuid-column">{{ submission.uuid }}</td>
+            <td>{{ submission.submission_date }}</td>
+            <td>{{ submission.metadata.accession_title }}</td>
+            <td>{{ submission.extent_statements }}</td>
+            <td>
+                <a href="{% url "recordtransfer:submission_detail" uuid=submission.uuid %}" target="_blank">{% trans "Click to view report" %}</a>
+            </td>
+            <td>
+                <a href="{% url "recordtransfer:submission_csv" uuid=submission.uuid %}">{% trans "Click to download" %}</a>
+            </td>
         </tr>
-        {% for submission in submissions %}
-            <tr>
-                <td class="uuid-column">{{ submission.uuid }}</td>
-                <td>{{ submission.submission_date }}</td>
-                <td>{{ submission.metadata.accession_title }}</td>
-                <td>{{ submission.extent_statements }}</td>
-                <td>
-                    <a href="{% url "recordtransfer:submission_detail" uuid=submission.uuid %}" target="_blank">{% trans "Click to view report" %}</a>
-                </td>
-                <td>
-                    <a href="{% url "recordtransfer:submission_csv" uuid=submission.uuid %}">{% trans "Click to download" %}</a>
-                </td>
-            </tr>
-        {% endfor %}
-    </table>
-</div>
+    {% empty %}
+    <tr>
+        <td colspan="5" class="empty-table-text">{% trans "You have not sent any records to the NCTR." %}</td>
+    </tr>
+    {% endfor %}
+</table>
+{% if page %}
+    {% include "includes/pagination.html" with page=page target_id=target_id paginate_url=paginate_url %}
+{% endif %}
--- a/app/recordtransfer/templates/recordtransfer/base.html
+++ b/app/recordtransfer/templates/recordtransfer/base.html
@@ -1,5 +1,6 @@
 {% load static %}
 {% load i18n %}
+{% load django_htmx %}
 <!DOCTYPE html>
 <html lang="en">
     <head>
@@ -7,6 +8,7 @@
         <!-- JS -->
         {% block javascript %}
             <script src="{% static 'base.bundle.js' %}"></script>
+            {% htmx_script %}
         {% endblock javascript %}
         {# Fonts #}
         <link rel="preconnect" href="https://fonts.googleapis.com">
@@ -27,7 +29,7 @@
             {% block title %}{% endblock title %}
         </title>
     </head>
-    <body>
+    <body hx-headers='{"x-csrftoken": "{{ csrf_token }}"}'>
         <div class="flex-wrapper">
             {% block banner %}
                 {% include "recordtransfer/banner.html" %}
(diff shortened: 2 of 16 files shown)


yenaing-oo:
@danloveg
I've established a pattern we can follow to implement pagination using HTMX. For now, I've applied it only to the submission groups table on the Profile page. Could you let me know what you think before I extend it to the other tables?

Here’s a quick summary of how it works:

- On page load, HTMX sends a GET request to the server to fetch the rendered table dynamically, filling the initially empty table container div.
- The view serving this request also passes the ID of the container holding the table to the table template via the context.
- The pagination template, included in the table template, uses HTMX to send similar requests that include the desired page number. Because it has access to the container ID, it can target and replace both the table and the pagination controls dynamically.

app/recordtransfer/templates/recordtransfer/base.html
```diff
@@ -1,12 +1,14 @@
 {% load static %}
 {% load i18n %}
+{% load django_htmx %}
```
danloveg, line 3: I was wondering what kind of interesting setup happened here but it's super basic, it just includes the HTMX source code on this page:

https://github.com/adamchainz/django-htmx/blob/main/src/django_htmx/jinja.py

yenaing-oo: Ah yes. The first one loads the tag and the second one includes the HTML.

yenaing-oo:
@danloveg 
What do you think about the path names that I have created (`submission-table/`, `in-progress-submission-table/`, `submission-group-table/`) for returned the rendered HTML for the tables? I don't think the naming of these are highly important, since they are not meant to be accessed directly by the user, but rather called via HTMX. This means the user will not get to see these paths.

But perhaps they can be changed to look more consistent (not necessarily RESTful) with the other paths, such as `submission/table`, `submission-group/table` and `submission/inprogress/table`. What do you think?

danloveg:
> @danloveg What do you think about the path names that I have created (`submission-table/`, `in-progress-submission-table/`, `submission-group-table/`) for returning rendered HTML for the tables? I don't think the naming of these are highly important, since they are not meant to be accessed directly by the user, but rather called via HTMX. This means the user will not get to see these paths.
> 
> But perhaps they can be changed to look more consistent (not necessarily RESTful) with the existing paths - perhaps `submission/table`, `submission-group/table` and `submission/in-progress/table`. What do you think?

I like how they're named right now, I don't think we need to change them from how you have them.

> EDIT: There is also the question of whether the view for an existing path like `submission-group/` should be shared to serve HTMX requests as well, by using [this](https://django-htmx.readthedocs.io/en/latest/middleware.html#django_htmx.middleware.HtmxDetails.__bool__) to differentiate between regular HTTP requests and HTMX requests. The view class would be less readable, but we end up with less URL paths.

Can you remind me where this might need to happen? Does the submission-group/ view get rendered in multiple places?

---

I'm going to start my review now so keep an eye out for that

yenaing-oo:
> > EDIT: There is also the question of whether the view for an existing path like submission-group/ should be shared to serve HTMX requests as well, by using [this](https://django-htmx.readthedocs.io/en/latest/middleware.html#django_htmx.middleware.HtmxDetails.__bool__) to differentiate between regular HTTP requests and HTMX requests. The view class would be less readable, but we end up with less URL paths.
> 
> Can you remind me where this might need to happen? Does the submission-group/ view get rendered in multiple places?

For sure. Right now, the path `submission-group/` is handled by [this view](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/b1c810562011a069cfcf93fd74eb445d0b8abbda/app/recordtransfer/views/post_submission.py#L155). My question is: should we combine that view with the new view function [`submission_group_table()`](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/b1c810562011a069cfcf93fd74eb445d0b8abbda/app/recordtransfer/views/profile.py#L133) so that they share the same path `submission-group/`?

danloveg (changes requested):

danloveg:
> > > EDIT: There is also the question of whether the view for an existing path like submission-group/ should be shared to serve HTMX requests as well, by using [this](https://django-htmx.readthedocs.io/en/latest/middleware.html#django_htmx.middleware.HtmxDetails.__bool__) to differentiate between regular HTTP requests and HTMX requests. The view class would be less readable, but we end up with less URL paths.
> > 
> > 
> > Can you remind me where this might need to happen? Does the submission-group/ view get rendered in multiple places?
> 
> For sure. Right now, the path `submission-group/` is handled by [this view](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/b1c810562011a069cfcf93fd74eb445d0b8abbda/app/recordtransfer/views/post_submission.py#L155). My question is: should we combine that view with the new view function [`submission_group_table()`](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/b1c810562011a069cfcf93fd74eb445d0b8abbda/app/recordtransfer/views/profile.py#L133) so that they share the same path `submission-group/`?

I don't think so, no. "submission-group" is singular, so it's intuitive that one group is handled by that view. That view might get really messy if you wanted to add a queryset to it, so it's best to keep the group table separate from the code that grabs the details of a single group in my opinion.

app/recordtransfer/templates/includes/submission_table.html
```diff
@@ -1,27 +1,32 @@
...
-                <td>
-                    <a href="{% url "recordtransfer:submission_csv" uuid=submission.uuid %}">{% trans "Click to download" %}</a>
-                </td>
-            </tr>
-        {% endfor %}
-    </table>
-</div>
+    {% empty %}
+    <tr>
+        <td colspan="5" class="empty-table-text">{% trans "You have not sent any records to the NCTR." %}</td>
```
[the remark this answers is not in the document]
yenaing-oo: Good catch, I missed that.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/530/commits/6d2777027b501bcaf6ca8be12e40d67c7a64742a

app/recordtransfer/views/profile.py
```diff
@@ -126,3 +102,65 @@ def form_invalid(self, form: BaseModelForm) -> HttpResponse:
...
+    queryset: QuerySet,
+    template_name: str,
+    target_id: str,
+    paginate_url: str,
+) -> HttpResponse:
+    """Define a generic function to render paginated tables. Request must be made by HTMX, or else
+    a 400 Error is returned.
+    """
+    if not request.htmx:
+        return HttpResponse(status=400)
```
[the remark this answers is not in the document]
yenaing-oo: I added a test for this, as well as testing that pagination works the way it should for the In-Progress Submission and Submission Group tables. I did not include tests for the Submission table as I haven't yet figured out a way to create test Submission objects.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/530/commits/5eac304cb763065994e46890aa251460ca18f5f0

app/recordtransfer/tests/unit/views/test_profile.py
```diff
@@ -319,15 +376,117 @@ def test_in_progress_submission_expiring_soon_shown_in_red(self) -> None:
...
         self.assertEqual(response.status_code, 200)
         content = response.content.decode()
         self.assertIn("red-text", content)
         self.assertIn("strikethrough", content)
+
+    @patch("recordtransfer.views.profile.PAGINATE_BY", 3)
+    def test_in_progress_submission_table_display(self) -> None:
+        """Test that the in-progress submission table displays in-progress submissions correctly."""
+        # Clear any existing in-progress submissions
+        InProgressSubmission.objects.filter(user=self.user).delete()
```
danloveg, line 396: You should use the [tearDown](https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDown) method instead of doing cleanup in test cases.

`tearDown` gets run after each test runs, so you can delete all in-progress submissions and groups in that function

app/recordtransfer/views/profile.py
```diff
@@ -84,10 +61,9 @@ def get_context_data(self, **kwargs) -> dict[str, Any]:
                     "ID_NEW_PASSWORD": ID_NEW_PASSWORD,
                     "ID_CONFIRM_NEW_PASSWORD": ID_CONFIRM_NEW_PASSWORD,
                 },
-                # Pagination
-                "IN_PROGRESS_PAGE": IN_PROGRESS_PAGE,
-                "SUBMISSIONS_PAGE": SUBMISSIONS_PAGE,
-                "GROUPS_PAGE": GROUPS_PAGE,
+                # Table container IDs
+                "ID_SUBMISSION_GROUP_TABLE": ID_SUBMISSION_GROUP_TABLE,
+                "ID_IN_PROGRESS_SUBMISSION_TABLE": ID_IN_PROGRESS_SUBMISSION_TABLE,
```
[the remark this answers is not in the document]
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/530/commits/058fcc2851261c2fcefc8f1dc809cbc34fbce430

danloveg (changes requested):
One last thing and then we're good to go

app/recordtransfer/tests/unit/views/test_profile.py
yenaing-oo: There were other test methods which relied on having an initialised `self.upload_session` and `self.in_progress_submission`, which is why I opted to clear any existing in-progress submissions at the start of my new test methods. 

I've refactored the test methods to be able to utilse the `tearDown()` method correctly now.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/530/commits/106c7a5100446dca575185e016bd2690e639ebf6

danloveg (approved):
Looks good!

Issue #532, opened 2025-04-30
Re-instate E2E tests on Pull Request


danloveg:
We had previously created a GitHub Action that runs end-to-end tests when making a pull request to the default branch, we should look to bring that action back so we have some automated test coverage on the Javascript in our application.

@yenaing-oo you had written that action, it's possible it's still in the commit history somewhere to be resurrected. I'm worried it might not be though because I typically squash commits and delete old branches, so if it's still around it might be in your local git repository.

PR #541, opened 2025-05-01
Re-instate E2E tests on Pull Request

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/532

### DIFF
--- a/.github/workflows/django-tests.yml
+++ b/.github/workflows/django-tests.yml
@@ -32,3 +32,8 @@ jobs:
 
       - name: Run unit tests
         run: poetry run pytest -k "unit"
+
+      - uses: browser-actions/setup-chrome@v1
+      - run: chrome --version
+      - name: Run E2E tests
+        run: poetry run pytest -m e2e
--- a/app/recordtransfer/tests/e2e/test_e2e.py
+++ b/app/recordtransfer/tests/e2e/test_e2e.py
@@ -85,7 +85,7 @@ class SubmissionFormWizardTest(StaticLiveServerTestCase):
         chrome_options = webdriver.ChromeOptions()
         chrome_options.add_argument("--disable-autofill")
         chrome_options.add_argument("--disable-save-password-bubble")
-        # chrome_options.add_argument("--headless")
+        chrome_options.add_argument("--headless")
         chrome_options.add_argument("--no-sandbox")
         chrome_options.add_argument("--disable-dev-shm-usage")
         prefs = {"autofill.profile_enabled": False}
--- a/app/recordtransfer/tests/e2e/test_e2e.py
+++ b/app/recordtransfer/tests/e2e/test_e2e.py
@@ -90,6 +90,7 @@ class SubmissionFormWizardTest(StaticLiveServerTestCase):
             chrome_options.add_argument("--headless")
         chrome_options.add_argument("--no-sandbox")
         chrome_options.add_argument("--disable-dev-shm-usage")
+        chrome_options.add_argument("--guest")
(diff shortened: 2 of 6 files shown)


yenaing-oo:
I changed the settings for the Chrome driver to run in headless mode, which means the browser window will not be visible. This setting required to run our E2E tests with solely a Chrome driver, although this also means we would have to unset during local development if we wish to view the tests being executed on our browser.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/541/commits/01b49bf8bef17acc3690bb0adc503bed39739aaf

app/recordtransfer/tests/e2e/test_e2e.py
```diff
@@ -85,7 +85,7 @@ def setUp(self) -> None:
         chrome_options = webdriver.ChromeOptions()
         chrome_options.add_argument("--disable-autofill")
         chrome_options.add_argument("--disable-save-password-bubble")
-        # chrome_options.add_argument("--headless")
+        chrome_options.add_argument("--headless")
```
danloveg, line 88: Is there a way to specify this argument conditionally? I like seeing the browser when these tests run locally but having to comment this line out when I want to do that doesn't sound great.

If there's no way to conditionally include this argument with `pytest`, maybe it's possible to conditionally include this argument when some environment variable is set?

```python
if os.getenv("SELENIUM_TESTS_HEADLESS_MODE"):
    chrome_options.add_argument("--headless")
```

Or something like that. Thoughts?

danloveg (changes requested):

.github/workflows/django-tests.yml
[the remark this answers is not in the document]
yenaing-oo: Even though the current way is more efficient, I agree that it is best to keep the two concerns separate. I'll make a separate action for that.

app/recordtransfer/tests/e2e/test_e2e.py
yenaing-oo: I think using an environment variable is the cleanest approach, rather than adding a custom flag and parsing it from the pytest command. Thanks for the suggestion.

yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/541/commits/cd000ad881dbc161e344d23db53adb2ea7d6938f

.github/workflows/django-tests.yml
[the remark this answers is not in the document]
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/541/commits/6f3e392ea939186828a47e92bccf0f3233078e70

yenaing-oo:
@danloveg 
Could you double-check that the E2E tests aren't flaky? I re-ran that workflow a couple of times to make sure.

danloveg:
> @danloveg Could you double-check that the E2E tests aren't flaky? I re-ran that workflow a couple of times to make sure.

They don't look to be flaky in the workflow, but they are flaky when I run them locally. For some reason, the file upload takes forever because it's waiting for the server to respond.

weirdly long wait

I can look more into why this happens. Uploading the file seems to complete really quick but then it stalls for a good 15 seconds before the file upload looks complete (i.e., the file upload progress indicator gets stuck at 100%)

Also, I got a popup that the password was included in a data breach and the second test got stuck at that popup. Can you create a more secure password for the test user so that popup doesn't appear?

danloveg (approved):

Issue #538, opened 2025-04-30
Navigate Between Submission Form Steps Without Page Reload


yenaing-oo:
Similar to [Issue #529](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/529), use HTMX to dynamically render different steps of the Submission Form, avoiding a full page reload.

This will require modifying each step’s template to include only the form content, rather than extending the base template as is currently done.

PR #550, opened 2025-05-07
Navigate Between Submission Form Steps Without Page Reload

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/538

Change notes:
- Installed HTMX's [head-support](https://htmx.org/extensions/head-support/) extension to support replacement/update of head elements
- Dynamically replace main contents of the Submission Form, as well as corresponding context for each form step using HTMX

Optional changes:
- Reorganized handling of JS form context to call right setup function based on context element ID in `submission_form/index.js`
- Refactored extracting of file upload settings, made it so that it is directly accessible from context in `submission_form/uppyForm.js`


yenaing-oo:
Note to self: Fix submission group form not showing due to missing context. Continue exploring [head-support](https://htmx.org/extensions/head-support/) extension to update head elements.

yenaing-oo:
I see that a Django template formatter will be added to the project soon (https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/553). I don't see any conflicting changes in that PR, so merging either one into `master` first should be OK.

app/recordtransfer/static/recordtransfer/js/submission_form/index.js
```diff
@@ -15,25 +15,61 @@ import {
...
-document.addEventListener("DOMContentLoaded", function () {
+const _setupWithContext = () => {
+    const contextElement = document.querySelector("[id^=\"js_context_\"]");
+    if (!contextElement) {
+        return;
+    }
+    const context = JSON.parse(contextElement.textContent);
+    
+    const idParts = contextElement.id.split("js_context_");
+    const contextFor = idParts.length > 1 ? idParts[1] : "";
```
danloveg, line 26: This is a little bit hard to read. How about this?

```suggestion
    const contextFor = contextElement.id.replace(/^js_context_/, "");
```

app/recordtransfer/static/recordtransfer/js/submission_form/uppyForm.js
```diff
@@ -18,17 +17,14 @@ import {
...
-    const settings = getFileUploadSettings();
-    // Don't render Uppy at all if settings are not available
-    if (!settings) {return;}
-
+export async function setupUppy(context) {
     const reviewButton = document.getElementById("form-review-button");
     const submissionForm = document.getElementById("submission-form");
     const issueFileIds = [];
-
+
```
danloveg, line 27: There is some extra whitespace here. Can you remove it and also add this setting to `settings.json`?

```json
"files.trimTrailingWhitespace": true,
```

app/recordtransfer/templates/recordtransfer/submission_form_base.html
```diff
@@ -34,7 +34,14 @@
...
     {% endblock form_explanation %}
-    <form id="submission-form" action="" method="post">
+    <form
+        id="submission-form"
+        hx-post="{{ request.path }}"
+        hx-target="#main-container"
+        hx-select="#main-container"
+        hx-swap="outerHTML"
+        method="POST"
+    >
```
danloveg, line 44: Can you explain what's happening here? And why choose `#main-container` to swap instead of the `<form>` itself?

I thought that you'd need to modify the template more than this, since this template still inherits from `base.html`. And just to be clear, I think this is a *good* thing that you didn't have to create a smaller template with just the form, I'm just surprised because it doesn't match with how I thought HTMX worked.

danloveg (changes requested):
It works great! I had a couple very minor comments, and one of them was more of a question of how this works just so I understand what's happening here.

app/recordtransfer/templates/recordtransfer/submission_form_base.html
yenaing-oo: Here's an summary of what happens:
- When one of the buttons within the form is clicked, it fires a submit event that bubbles up to the form, which is handled by HTMX
- It sends a POST request to the same URL as the current page (which is good because we want to maintain any query parameters that were present, such as `resume` or `group`
- The target element to replace is `#main-container`, with `hx-swap="outerHTML"` being used to ensure that the response HTML replaces `#main-container`, instead of it being _fitted_ inside `#main-container` (which is the default behaviour, set using `innerHTML`.  
- I found out that I could choose which part of the response HTML I want to use for replacement. `hx-select` allows me to do this by choosing just that element to use from the response. This means I don't have to modify any of the individual templates for each step to include just the relevant elements. 

I chose `#main-container` instead of the `<form>` element because there are other elements aside from the form element that need replacing, such as:
- The heading title of the form step
- Info message
- Messages (success/error)
- Form explanation text

In addition, only the Group Submission step involves the use of a modal. Replacing just the form element would cause the modal to be missing on that step. Including the modal under the `content` block in `submission_form_base.html` is not ideal either, since I only want to include it for that step.

app/recordtransfer/static/recordtransfer/js/submission_form/index.js
yenaing-oo: That looks better. I've changed that here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/550/commits/0d3e8cd7643c730276fdd2cc6985bf8610c366eb

app/recordtransfer/static/recordtransfer/js/submission_form/uppyForm.js
yenaing-oo: I've added that setting here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/550/commits/08a5ca2c065e2357b406665edf31fb905ca5ca43
And removed the mentioned whitespace here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/550/commits/810515889296f3ae5f2628f40d3a9271827ea423

app/recordtransfer/templates/recordtransfer/submission_form_base.html
yenaing-oo: I was surprised by HTMX's `hx-select` feature too. It does save me time from creating snippets of HTML to use for replacement. Although, I don't think making a snippet template like I did in https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/529 would be feasible here because the individual step templates _extend_ from the base template and use {% block block_name %} to override the base template. I would also have to find a way to fetch the new context for the step.

danloveg: That all makes sense, thanks for the explanation.

I do have one more concern - we've got a bit of a weird setup with including the main HTMX code with `{% htmx_script %}` and then we're packaging `htmx-ext-head-support` via Webpack. Do you think it makes sense to add HTMX to the `package.json` and manage it in the same way we're managing `htmx-ext-head-support` now? Say for example there's a security vulnerability in HTMX - we'd have better luck updating to the latest version via npm than we would waiting for the maintainers of the `django-htmx` Python package to pull the latest version of HTMX and make it available in a new release of the `django-htmx` package. Open to suggestions here; I think it makes sense to manage all our static assets with Webpack.

yenaing-oo: That's a good point, I agree that how HTMX is set up right now is a bit weird. I'll see if dropping `django-htmx` will have any other consequences aside from losing the middleware that allows us to use shortcuts to access htmx headers, e.g. `request.htmx`. For that, we would have to compare the header values ourselves.

danloveg: I'm thinking we keep `django-htmx` for its middleware, but we don't use its bundled version of HTMX, we bundle HTMX ourselves. Does that make sense?

yenaing-oo: I had to make a few changes to include HTMX as a dependency directly via NPM. If we find that the features provided by `django-htmx` are worthwhile, we can always switch back to it. Currently, the only feature we were using was the shortcut to checking a request for HTMX headers. It looks like everything works as before, please verify this is the case.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/550/commits/092ba13c744831f1e39a2607c8236e2751c6ce06

yenaing-oo: > I'm thinking we keep `django-htmx` for its middleware, but we don't use its bundled version of HTMX, we bundle HTMX ourselves. Does that make sense?

Sorry, I only saw this comment after refreshing my page.

yenaing-oo: I see what you mean now. We can still use `django-htmx` for it's middleware without using the HTMX bundle that comes with it.

danloveg: That looks good, it's clearer where `htmx` is coming from now.

danloveg (approved):
No issues, everything is working well. I like that we're bundling all the Javascript ourselves now, loading a static `.js` file from some other Django app didn't really feel right to me so this looks great.

PR #552, opened 2025-05-08
540 Remove optional logger arguments

Eashana0104:
Remov option logger arguments and keep the module's logger

### DIFF
--- a/app/recordtransfer/models.py
+++ b/app/recordtransfer/models.py
@@ -384,13 +384,13 @@ class UploadSession(models.Model):
             if save:
                 self.save()
 
-    def make_uploads_permanent(self, logger: Optional[logging.Logger] = None) -> None:
+    def make_uploads_permanent(self) -> None:
         """Make all temporary uploaded files associated with this session permanent.
 
         Args:
             logger: Optional logger instance to use for logging operations
         """
-        logger = logger or LOGGER
+        logger = LOGGER
 
         if self.status == self.SessionStatus.STORED:
             logger.info(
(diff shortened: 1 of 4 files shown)


app/recordtransfer/models.py
```diff
@@ -384,13 +384,13 @@ def remove_temp_uploads(self, save: bool = True) -> None:
...
 
-    def make_uploads_permanent(self, logger: Optional[logging.Logger] = None) -> None:
+    def make_uploads_permanent(self) -> None:
         """Make all temporary uploaded files associated with this session permanent.
 
         Args:
             logger: Optional logger instance to use for logging operations
         """
-        logger = logger or LOGGER
+        logger = LOGGER
```
danloveg, line 393: Assigning `LOGGER` to `logger` is redundant, can you refactor the rest of the functions to use `LOGGER` instead of re-assigning to `logger`?

For example, instead of:

```python
logger = LOGGER

# ... some other code ...

logger.info(
    "Moving %d temporary uploaded files from the session %s to permanent storage",
    len(files),
    self.token,
)
```

You can do this:

```python
LOGGER.info(
    "Moving %d temporary uploaded files from the session %s to permanent storage",
    len(files),
    self.token,
)
```

danloveg (changes requested):
It's nice to see these logger arguments finally removed. I do recommend though that you use the `LOGGER` defined at the top of the module instead of re-assigning LOGGER to a new logger variable.

The way scoping works in Python is that variables defined outside a function are available to be used in that function:

```python

def function():
    LOGGER.info("Using the LOGGER is valid here")
```

app/recordtransfer/models.py
Eashana0104: I will do that right now

danloveg: Looks good. When it's ready for review you can re-request my review

danloveg (approved):
Looks good

Issue #547, opened 2025-05-06
Standardize Podman/Docker Mentions in Sphinx Docs


yenaing-oo:
In the "Running the App" section (`docs/running/index.rst`), the only sub-section with Docker-equivalent commands specified is "Resetting the Database". The other sub-sections include commands for Podman only.

To be consistent, remove individual Docker-equivalent commands entirely since the alternative would be to provide equivalents everywhere else, and instead mention that `docker compose` can be used in place of `podman-compose` at the start of the page.

PR #554, opened 2025-05-08
Standardize Podman/Docker Mentions in Sphinx Docs 

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/547

Change notes:
- Mentioned `docker compose` can be used in place of `podman-compose` in Sphinx docs
- Removed individual Docker-equivalent commands from Sphinx docs
- Added text wrapping setting for RST files, since line breaks within a single paragraph are handled by the rendering application and explicitly including them do not make a difference

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -28,6 +28,7 @@
     "[restructuredtext]": {
         "editor.rulers": [
             99
-        ]
+        ],
+        "editor.wordWrap": "on"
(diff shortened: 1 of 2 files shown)


danloveg (approved):
Looks good! I also like the new setting, since it matches the way we're writing Markdown files.

PR #556, opened 2025-05-09
Add Django Template Formatter

danloveg:
Adds `djlint` for formatting Django templates. Also ran it against two templates just to check that it's working.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,10 +10,6 @@
     "ruff.nativeServer": "on",
     "python.analysis.typeCheckingMode": "standard",
     "editor.formatOnSave": true,
-    // For indentation to be set as 4 spaces
-    "editor.tabSize": 4,
-    "editor.insertSpaces": true,
-    "editor.detectIndentation": false,
     // Python Testing
     "python.testing.pytestEnabled": true,
     "python.analysis.extraPaths": [
@@ -59,5 +55,5 @@
         "H030",
         "H031",
         "D018"
-    ],
+    ]
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -30,5 +30,28 @@
             99
         ],
         "editor.wordWrap": "on"
-    }
+    },
+    "[django-html]": {
+        "editor.formatOnSave": true,
+        "editor.defaultFormatter": "monosans.djlint"
+    },
+    "files.associations": {
+        "**/templates/**/*.html": "django-html"
+    },
+    "djlint.ignore": [
+        "T002",
+        "T003",
+        "H005",
+        "H006",
+        "H008",
+        "H013",
+        "H016",
+        "H020",
+        "H021",
+        "H023",
+        "H025",
+        "H030",
+        "H031",
+        "D018"
+    ]
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,6 +10,10 @@
     "ruff.nativeServer": "on",
     "python.analysis.typeCheckingMode": "standard",
     "editor.formatOnSave": true,
+    // For indentation to be set as 4 spaces
+    "editor.tabSize": 4,
+    "editor.insertSpaces": true,
+    "editor.detectIndentation": false,
     // Python Testing
     "python.testing.pytestEnabled": true,
     "python.analysis.extraPaths": [
@@ -55,5 +59,5 @@
         "H030",
         "H031",
         "D018"
-    ]
+    ],
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -38,6 +38,8 @@
     "files.associations": {
         "**/templates/**/*.html": "django-html"
     },
+    // Find the rules here:
+    // https://www.djlint.com/docs/linter/
     "djlint.ignore": [
         "T002",
         "T003",
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,6 +10,10 @@
     "ruff.nativeServer": "on",
     "python.analysis.typeCheckingMode": "standard",
     "editor.formatOnSave": true,
+    // For indentation to be set as 4 spaces
+    "editor.tabSize": 4,
+    "editor.insertSpaces": true,
+    "editor.detectIndentation": false,
     // Python Testing
     "python.testing.pytestEnabled": true,
     "python.analysis.extraPaths": [
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -41,23 +41,5 @@
     },
     "files.associations": {
         "**/templates/**/*.html": "django-html"
-    },
-    // Find the rules here:
-    // https://www.djlint.com/docs/linter/
-    "djlint.ignore": [
-        "T002",
-        "T003",
-        "H005",
-        "H006",
-        "H008",
-        "H013",
-        "H016",
-        "H020",
-        "H021",
-        "H023",
-        "H025",
-        "H030",
-        "H031",
-        "D018"
-    ]
+    }
 }
--- a/app/recordtransfer/templates/recordtransfer/about.html
+++ b/app/recordtransfer/templates/recordtransfer/about.html
@@ -1,8 +1,8 @@
-{% extends 'recordtransfer/base.html' %}
+{% extends "recordtransfer/base.html" %}
 {% load i18n %}
 {% block title %}
     {% trans "About" %}
-{% endblock %}
+{% endblock title %}
 {% block content %}
     <div class="title-text">{% trans "About" %}</div>
     <p>
@@ -39,4 +39,4 @@
             {% endif %}
         {% endfor %}
     {% endif %}
-{% endblock %}
+{% endblock content %}
--- a/app/recordtransfer/templates/recordtransfer/about.html
+++ b/app/recordtransfer/templates/recordtransfer/about.html
@@ -1,47 +1,42 @@
 {% extends 'recordtransfer/base.html' %}
 {% load i18n %}
-
-{% block title %}{% trans "About" %}{% endblock %}
-
+{% block title %}
+    {% trans "About" %}
+{% endblock %}
 {% block content %}
-<div class="title-text">{% trans "About" %}</div>
-<p>
-    {% blocktrans %}
-    The NCTR Record Transfer Portal allows people to send records to the National Centre for Truth
-    and Reconciliation (NCTR) Archives securely over the internet. For small record submissions, this
-    is the preferred method of sending records to the NCTR Archives.
-    {% endblocktrans %}
-</p>
-
-{% if FILE_UPLOAD_ENABLED %}
-<div class="title-text margin-top-25px">{% trans "Maximum Submission Size" %}</div>
-<p>
-    {% blocktrans %}
-    Submissions made through the transfer portal are subject to the maximum submission size.
-    You may upload a <b>maximum of {{ MAX_TOTAL_UPLOAD_COUNT }} files</b>, and <b>at most
-    {{ MAX_TOTAL_UPLOAD_SIZE_MB }} MB of files</b>. Any <b>one file may not exceed
-    {{ MAX_SINGLE_UPLOAD_SIZE_MB }} MB</b>. If you would like to submit <i>more</i> than this
-    amount, please split your submission into multiple parts, and use the group form to group similar
-    submissions together so our archivists know your submissions are related.
-    {% endblocktrans %}
-</p>
-<div class="title-text margin-top-25px">{% trans "Accepted File Types" %}</div>
-<p>
-    {% blocktrans %}
-    The following are the types of files that donors can submit using this application.
-    {% endblocktrans %}
-</p>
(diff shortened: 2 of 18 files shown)


danloveg:
@Eashana0104 can you add a comment to `settings.json` above the djlint ignore rules linking to the page where you can find those rules?

danloveg:
@yenaing-oo I worked on this with Ash, if you could take a look and make sure everything's working right that'd be appreciated, thanks

yenaing-oo:
> @yenaing-oo I worked on this with Ash, if you could take a look and make sure everything's working right that'd be appreciated, thanks

The indentation seems off. For e.g., when I press save on `profile.html`, I get this, where the indents are 2 spaces instead of the default 4:

image

Apparently, you can configure ignore rules as well as other settings using a Djlint config file: https://www.djlint.com/docs/configuration/

But setting the `indent` rule  using a config file did not seem to fix the issue. Might be an issue on my end? The indentation on the sample files in the PR look correct with 4 spaces.

I made sure to install the extension and run `uv sync --extra dev`.

danloveg:
> > @yenaing-oo I worked on this with Ash, if you could take a look and make sure everything's working right that'd be appreciated, thanks
> 
> The indentation seems off for me. For e.g., when I press save on `profile.html`, I get this, where the indents are 2 spaces instead of the default 4:
> 
> image
> 
> Apparently, you can configure ignore rules as well as other settings using a Djlint config file: https://www.djlint.com/docs/configuration/
> 
> I tried setting the `indent` rule using a config file but that did not fix the issue. Might be an issue on my end? The indentation on the sample files in the PR look correct with 4 spaces.
> 
> I made sure to install the extension and run `uv sync --extra dev`.

I had to set "Spaces: 4" in VS Code, I think it obeys whatever spacing is set by VS Code. But I agree it'd be better if there were some way to enforce that rule

Screenshot 2025-05-09 114307

Eashana0104:
> @Eashana0104 can you add a comment to `settings.json` above the djlint ignore rules linking to the page where you can find those rules?

I added a link where all these rules could be found

yenaing-oo:
> I had to set "Spaces: 4" in VS Code, I think it obeys whatever spacing is set by VS Code. But I agree it'd be better if there were some way to enforce that rule

Ah yes. I forgot VSCode has its own setting for indentation. I'll let @Eashana0104 look into enforcing consistent spacing.

Eashana0104:
I think i added to settings.json to set the default indentation to 4 spaces, lemme know if it works

danloveg:
> > I think i added to settings.json to set the default indentation to 4 spaces, lemme know if it works
> 
> This works great, I think it should be fine to set the indentation to 4 spaces globally for all file types like this, since this is consistent with the formatting settings for JS and Python. What do you think @danloveg ?

This works for me too, and is a good solution to ensure four spaces are used (almost) everywhere. The only files I can think of that don't have four spaces that are affected by this change are `package.json` and `package-lock.json` but I'm OK with changing those to four spaces next time anyone touches either of those files.

Eashana0104:
Yeah i hit cntr save and it automatically indents to 4 spaces , these two files. I can commit that now

.vscode/settings.json
```diff
@@ -30,5 +34,30 @@
...
+        "H013",
+        "H016",
+        "H020",
+        "H021",
+        "H023",
+        "H025",
+        "H030",
+        "H031",
+        "D018"
+    ],
```
yenaing-oo, line 62: I think it would be better if these settings were moved to a config file. The VSCode djlint extension is able to read this config file and lint accordingly. I already tested this out by adding `djlint.toml` in the root directory:

```
ignore = """
    T002,
    T003,
    H005,
    H006,
    H008,
    H013,
    H016,
    H020,
    H021,
    H023,
    H025,
    H030,
    H031,
    D018
"""
```
You might have to restart your VSCode to see these changes take place, since this extension does not provide a way to restart like Ruff does.

danloveg:
> Yeah i hit cntr save and it automatically indents to 4 spaces , these two files. I can commit that now

Can you revert that change? `git revert d10e694bff50270836eaa327105c9dcc8f5086fd` and then `git push`. We don't need to update those files right now, we can just update their indentation the next time we modify those files. Kind of like the templates - we're not re-formatting all of them right now, we'll format them in a gradual manner whenever we're working on any of those files.

.vscode/settings.json
yenaing-oo: Also, could you change the link to the "Rules" section specifically like this?:
https://www.djlint.com/docs/linter/#rules

danloveg: I like that more, then the ignore settings are in a similar place to where `ruff.toml` is

yenaing-oo: Upon further testing, it seems that djlint does not like multi-line strings. You'll have to include all the rules as a single line string in `djlint.toml`:
`ignore = "T002,T003,H005,H006,H008,H013,H016,H020,H021,H023,H025,H030,H031,D018"`

yenaing-oo: I'd like to discuss some of the rules that have been selected to ignore.

Ignored rules that we should keep (in my opinion):
- H016
- H021
- H025
- D018

Special-case ignored rules:
- H020: I'm aware of the empty <label> tags we currently have in our templates, used for spacing. Perhaps we can ignore this rule for now until we use proper styling and remove these tags
- H030, H031: I think we can add these back in once we start caring about SEO.

danloveg: Looking at them now, I think the ignore rules warrant some more discussion, too.

Rules I think we should keep (some overlap with Kelvin's list):

- H006. Because you can set an image size in CSS.
- H020. We need this for empty labels like Kelvin mentioned. After re-designing the site we won't need this rule.
- H021. Sometimes you need inline styles, like in emails.
- H031. Not sure we need these yet, I agree with Kelvin on this one.
- D018. I don't think this is fully necessary especially for static links.

Rules I'm not so sure about:

- H016. We have a <title> tag in our base template so I don't know if ignoring this one has any merit.
- H023. I don't really know what this rule is talking about.
- H025. I don't really know what this rule is talking about.

Rules I think we should remove from the list:

- T002. This is good practice.
- T003. This is also good practice, consider when there are multiple endblocks one after another, explicitly detailing which block is ending there will help make the template more readable.
- H005. I don't see why we can't add a lang attribute to html.
- H008. This is standard when writing HTML attributes.
- H013. This is required for websites to be accessible, we should look to add alt text to images going forward.
- H030. Having a meta description is standard, we should add this.

Eashana0104: yeah this makes sense, so maybe whenever someone is changing these files, they can fix all these errors that vs code is showing. I changed the rules to only include the necessary rules in the djlint.toml file.

danloveg: Great, thanks. @yenaing-oo are you happy with the new list of rules? I'd like us all to come to a consensus before we finalize these rules.

@Eashana0104 do you have any opinions on the specific rules I chose to ignore or the ones I wasn't sure about? I wasn't certain what H023 or H025 were. Let's discuss what rules we should be using first here before we change the list again.

yenaing-oo: You can disregard my list. It'll be easier to reference Daniel's list from now on. I agree with everything, although I'd like to discuss a few of the rules:

> D018. I don't think this is fully necessary especially for static links.

I think this rule only applies to _internal_ links (i.e. ones that begin with `/`), so static links like the social links won't be affected. So, I think it should be fine to exclude this rule from the ignore list.

> H023. I don't really know what this rule is talking about.

I believe this is referring to [symbol entities](https://www.w3schools.com/charsets/ref_html_symbols.asp). I think we should remove this rule from the ignore list since I don't see why we shouldn't use them.

> H025. I don't really know what this rule is talking about.

I'm not sure on this either. It should be able to tell whether a tag is an orphan or not, but the wording of the rule with "seems to be an orphan" is throwing me off

danloveg: > > D018. I don't think this is fully necessary especially for static links.
> 
> I think this rule only applies to _internal_ links (i.e. ones that begin with `/`), so static links like the social links won't be affected. So, I think it should be fine to exclude this rule from the ignore list.

It looks like this link triggers an error with that rule:

https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/c64ed75685359700ffe0f8c9910087f21d536043/app/recordtransfer/templates/admin/base_site.html#L88

I do think it'd be good to change that to a `{% url %}` link, you don't really need to hardcode internal links in Django. So yeah, let's remove rule D018 too.

> > H023. I don't really know what this rule is talking about.
> 
> I believe this is referring to [symbol entities](https://www.w3schools.com/charsets/ref_html_symbols.asp). I think we should remove this rule from the ignore list since I don't see why we shouldn't use them.

Ah, I see. Yes, I don't know why that'd be against the rules, it's fine to leave that rule out.

> > H025. I don't really know what this rule is talking about.
> 
> I'm not sure on this either. It should be able to tell whether a tag is an orphan or not, but the wording of the rule with "seems to be an orphan" is throwing me off

Let's leave it out then, if we get an error that's flagging that rule we'll look at it then.

---

So far, then, these are the rules we want to ignore (i.e., these rules should be in the list):

- H006
- H020
- H021
- H031

Eashana0104: and I changed this too

Eashana0104: So i think i agree with what the final list of rules look like, I just added the previous rules just because they were giving me an error at that point and I didn't want to create error while fixing an issue. But yeah those are definitely good practices.

 I think D006 should be ignored as we can even dynamically size images in CSS, we don't have to give it a fixed height and width ,

 H020 could be kept in the list for now as I noticed some empty label tags, Once we add something there which would be a good practice , we could remove it off the list .

For H021, even I was confused given that its always a good practice to create a separate CSS file for the purpose of reusability and maintainability , but if that code is not a lot and not getting repeated in different places, we could ignore that rule as well. 

And for H031, honestly I didn't see meta tag get used almost anywhere , that makes me think its not important anymore, even when I googled about it , It was used for SEO optimization but google doesn't even use it anymore.

 For H025 , regarding the orphan tags, I think this is bad, weshould not ignore this given its bad to have unclosed tags.

danloveg: Alright, sounds like we're in agreement!

I had to push another commit to this branch to revert some changes made to the templates but everything looks good to me

app/recordtransfer/templates/recordtransfer/about.html
```diff
@@ -1,47 +1,42 @@
 {% extends 'recordtransfer/base.html' %}
 {% load i18n %}
-
-{% block title %}{% trans "About" %}{% endblock %}
-
+{% block title %}
+    {% trans "About" %}
+{% endblock %}
```
yenaing-oo, line 5:
```suggestion
{% endblock title %}
```

app/recordtransfer/templates/recordtransfer/about.html
```diff
@@ -1,47 +1,42 @@
 {% extends 'recordtransfer/base.html' %}
```
yenaing-oo, line 1:
```suggestion
{% extends "recordtransfer/base.html" %}
```

app/recordtransfer/templates/recordtransfer/about.html
```diff
@@ -1,47 +1,42 @@
...
+        {% for group, extensions in ACCEPTED_FILE_FORMATS.items %}
+            {% if extensions %}
+                <div class="medium-text">{{ group }} Files</div>
+                <ul>
+                    {% for extension in extensions %}<li>{{ extension }}</li>{% endfor %}
+                </ul>
+            {% endif %}
+        {% endfor %}
+    {% endif %}
+{% endblock %}
```
yenaing-oo, line 42:
```suggestion
{% endblock content %}
```

yenaing-oo (changes requested):
Since we're enforcing these new rules, we might as well fully comply with them for the two template files we're making changes to in this PR. I found several places that could use fixing.

Aside from that, everything looks good to me!

app/recordtransfer/templates/recordtransfer/about.html
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/556/commits/72334f2999189a3f1e03750e1616988ed62a7f3e

app/recordtransfer/templates/recordtransfer/about.html
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/556/commits/72334f2999189a3f1e03750e1616988ed62a7f3e

yenaing-oo (approved):
Everything looks good to me. Could you take a final look when you have the chance, @danloveg?

danloveg:
> Everything looks good to me. Could you take a final look when you have the chance, @danloveg?

Looks good to me too. @yenaing-oo I'm going to merge this in but you'll probably get some merge conflicts in #557 because both pull requests touched the `pyproject.toml` and `uv.lock` file.

Issue #543, opened 2025-05-02
Run Ruff Linting on Pull Request


danloveg:
To ensure a consistent quality of Python files, we should look to run linting with Ruff against Python files when a pull request is made.

Because there are going to be a *lot* of linting errors at first, I'm proposing we only target a small part of the codebase with this linter just to make sure it's working as we expect. I am going to suggest we target `app/recordtransfer/views/*` because there are only 43 errors in that module, versus 1344 errors app-wide (at the time of writing). We can expand the scope later on.

This issue should include four things:

1. `ruff` should be added to the *dev* dependencies in `pyproject.toml`
  - Add it with: `poetry add ruff --extras dev`
2. `.ruff_cache/` should be added to `.gitignore`
3. A new GitHub action that runs when a pull request is made (see other actions for how to specify this trigger)
  - The command should be: `poetry run ruff check app/recordtransfer/views/`
  - We'll increase the linting scope later on, just focus on the views for now.
4. The linting errors in `views` should be fixed
  - You can find and fix the issues locally in VSCode or by running the same command mentioned above, `poetry run ruff check app/recordtransfer/views/`

If there are any rules that seem too strict, open an issue in the [Python Development Guide](https://github.com/NationalCentreTruthReconciliation/Python-Development-Guide).

danloveg:
The poetry commands above should be replaced with their `uv` equivalent now that #545 is merged.

- `uv add ruff --optional dev`
- `uv run ruff check app/recordtransfer/views/`

PR #557, opened 2025-05-09
Run Ruff Linting on Pull Request

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/543

Change notes:
- Fixed all Ruff linting errors in `app/recordtransfer/views`
- Created a GitHub action to lint check Python code in `app/recordtransfer/views` using Ruff

### DIFF
--- a/.github/workflows/lint-py.yml
+++ b/.github/workflows/lint-py.yml
@@ -18,4 +18,4 @@ jobs:
         with:
           src: "./app/recordtransfer/views"
           version-file: "./pyproject.toml"
-          args: "--config ./ruff.toml"
+          args: "--config ./ruff.toml check"
--- a/.github/workflows/lint-py.yml
+++ b/.github/workflows/lint-py.yml
@@ -16,4 +16,6 @@ jobs:
       - uses: actions/checkout@v4
       - uses: astral-sh/ruff-action@v3
         with:
-          src: app/recordtransfer/views/
+          src: "./app/recordtransfer/views"
+          version-file: "./pyproject.toml"
+          args: "--config ./ruff.toml"
--- a/.github/workflows/lint-py.yml
+++ b/.github/workflows/lint-py.yml
@@ -1,0 +1,19 @@
+name: PR Python Lint Check
+
+on:
+  pull_request:
+    branches: [master]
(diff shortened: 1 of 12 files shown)


.github/workflows/lint-py.yml
```diff
@@ -0,0 +1,21 @@
...
+  lint-recordtransfer-views:
+    name: Lint RecordTransfer Views
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: astral-sh/ruff-action@v3
+        with:
+          src: "./app/recordtransfer/views"
+          version-file: "./pyproject.toml"
+          args: "--config ./ruff.toml check"
```
yenaing-oo, line 21: I had to specify the locations of `pyproject.toml` and `ruff.toml` since the action was looking for them in `src`. I also had to add `check` as part of `args` otherwise an invalid command like this gets run:
`ruff --config ./ruff.toml ./app/recordtransfer/views`

danloveg: Looks like you tried every possible combination of arguments before landing on that one, thanks for figuring that out!

yenaing-oo: Yeah this action is less intuitive than I thought 😆 . I wish I could define `config` the same way I'm able to define `src`.

app/recordtransfer/views/post_submission.py
```diff
@@ -29,21 +36,22 @@ class SubmissionDetail(UserPassesTestMixin, DetailView):
     template_name = "recordtransfer/submission_detail.html"
     context_object_name = "submission"
 
-    def get_object(self, queryset=None) -> Submission:
+    def get_object(self, queryset: Any = None) -> Submission:
```
danloveg, line 39: I think you can be more specific here:

```suggestion
    def get_object(self, queryset: Optional[QuerySet] = None) -> Submission:
```

danloveg (changes requested):
One *very* small comment, looks good otherwise

danloveg (approved):
I'll let you fix the merge conflicts before we merge this in, but everything else looks good.

danloveg (approved):

PR #560, opened 2025-05-09
Add a command to install gettext in Containerfile

Eashana0104:
I added the command to install and run gettext which helped in resolving the makemessages issue.
Currently .po and .mo files are getting creates. I tested it with 2 different languages as well ie en ad fr


Containerfile
```diff
@@ -14,6 +14,11 @@ RUN uv sync
 COPY package*.json ${PROJ_DIR}
 RUN npm install --no-color
 
+# 🔧 Install gettext for makemessages (includes msguniq)
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends gettext && \
+    rm -rf /var/lib/apt/lists/*
+
```
danloveg, line 21: This is correct, but you've encountered an interesting feature of containers: layer caching. This command need to go here instead:

```Dockerfile
WORKDIR ${PROJ_DIR}

# vvv The installation should go here instead vvv
RUN apt-get update && \
    apt-get install -y --no-install-recommends gettext && \
    rm -rf /var/lib/apt/lists/*

COPY pyproject.toml uv.lock ${PROJ_DIR}
RUN uv sync
```

See my review comments for more information on why this is.

danloveg (changes requested):
Each of the `COPY`, `RUN`, etc. commands in a Dockerfile is a separate *layer*. Docker will cache each layer, so that when the image needs to be built, it will use the cache **until it gets to a layer where something changes**.

This is an important concept to keep in mind. The way you've written the Dockerfile, anytime any of these files change, `gettext` will need to be re-installed in the image:

- `pyproject.toml`
- `uv.lock`
- `package.json`
- `package-lock.json`

Let's say for example we add a new Javascript dependency. That means that `package.json` will have changed. That also means that Docker's cache at this step: `COPY package*.json ${PROJ_DIR}` gets *invalidated*, and every future command in the Dockerfile has to get re-run. This means it will re-install `gettext`, but that doesn't sound right, because `gettext` has nothing to do with Javascript dependencies!

The way we address that is to move the installation as early in the Dockerfile as we can to leverage caching. If you move it to the location I've mentioned, none of the Python or Javascript dependencies changing will trigger re-installing `gettext`, which is what we want.

Here's another useful resource on image caching: https://docs.docker.com/build/cache/

Containerfile
Eashana0104: That does make sense. I am going to go through the article you attached as well.

danloveg (approved):

PR #561, opened 2025-05-12
Make tooltip work on signup page

Eashana0104:
import "htmx-ext-head-support";
import htmx from "htmx.org"; these were giving errors in the  base directory in js, So I removed this.
Move code for tooltip from widgets.js to tooltip.js inside base directory.
Corrected djlint errors from the files on which I worked on


app/recordtransfer/static/recordtransfer/js/base/index.js
```diff
@@ -1,10 +1,10 @@
-import "htmx-ext-head-support";
-import htmx from "htmx.org";
 import { setupMessages } from "./messages";
 import { setupNavbar } from "./navbar";
+import { setupHelpTooltips } from "./tooltip";
+
 
-window.htmx = htmx;
```
danloveg, line 6: Removing the `htmx` imports removes HTMX support in our application, can you add those lines back, and add back the line `window.htmx = htmx;`?

app/recordtransfer/templates/recordtransfer/submission_form_base.html
danloveg, line 1: Can you revert the changes to this file? There should be no template changes required in this issue besides the signupform

app/recordtransfer/templates/includes/submission_group_form.html
danloveg, line 1: Can you revert the changes to this file? There should be no template changes required in this issue besides the signupform

app/recordtransfer/static/recordtransfer/js/submission_form/widgets.js
```diff
@@ -1,17 +1,10 @@
...
-    flip,
-    shift,
-    offset,
-} from "@floating-ui/dom";
 import AirDatepicker from "air-datepicker";
 import localeEn from "air-datepicker/locale/en";
 import IMask from "imask";
 
+
+
```
danloveg, line 7: ESLint complains about these two empty lines, can you remove them?

app/recordtransfer/static/recordtransfer/js/submission_form/unsavedChanges.js
danloveg, line 1: `currentFormStep` is set in `submission_form_base.html`, I don't believe it's available in the window itself. I see where you're going, and agree that this should be refactored in the future. The currentFormStep is basically some context that Django is sending to the template, which then gets included in Javascript. The proper way to pass context that we've found is with [json_script](https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#json-script) - I can open up a new issue for that.

For now, can you revert the changes to this file?

app/recordtransfer/templates/recordtransfer/signupform.html
```diff
@@ -41,9 +35,10 @@
                 </div>
             {% endfor %}
             <div class="flex-item">
-                <label></label>
+                {{ field.label_tag }}
```
danloveg, line 38: Due to an unfortunate styling decision I made in the past, this is actually required to be an empty label:

```suggestion
                <label></label>
```

Additionally, `{{ field }}` is something that exists only in the for loop, there's no `{{ field }}` that's defined here.

app/recordtransfer/templates/recordtransfer/signupform.html
```diff
@@ -1,28 +1,22 @@
...
         {% endif %}
-
         <form method="post" action="{% url 'recordtransfer:create_account' %}">
             {% csrf_token %}
             {% for field in form %}
                 {% if field.errors %}
                     {% for error in field.errors %}
                         <div class="flex-error">
-                            <label></label>
+                            {{ field.label_tag }}
```
danloveg, line 19: This needs to be an empty label, because the way it's set right now, you get an extra label in the template which doesn't look right when there's an error:

Screenshot 2025-05-13 081849

```suggestion
                            <label></label>
```

app/recordtransfer/templates/recordtransfer/signupform.html
```diff
@@ -31,7 +25,7 @@
                     <label for="{{ field.id_for_label }}">{{ field.label }}</label>
                     {{ field }}
                     {% if field.help_text %}
-                        <div tooltip-content="{{ field.help_text }}" class="help-tooltip help-icon">
+                        <div tooltip-content={{ field.help_text }} class="help-tooltip help-icon">
```
danloveg, line 28: Removing the double quotes here has broken the tooltips because `field.help_text` may have spaces in it. See for example this tooltip which only says "This":

Screenshot 2025-05-13 082112

It should say what's in the `help_text` here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/c6cb705cdda41c025289c21b9461bc32d699d30e/app/recordtransfer/forms/__init__.py#L42

The fact that that form is in `__init__.py` is not great, I can open a new issue to address that.

```suggestion
                        <div tooltip-content="{{ field.help_text }}" class="help-tooltip help-icon">
```

app/recordtransfer/templates/recordtransfer/signupform.html
```diff
@@ -41,9 +35,10 @@
...
             {% endfor %}
             <div class="flex-item">
-                <label></label>
+                {{ field.label_tag }}
                 <button type="submit" class="blue-button">{% trans "Sign Up" %}</button>
             </div>
         </form>
     </div>
-{% endblock %}
+    <script src="{% static 'base.bundle.js' %}"></script>
```
danloveg, line 43: You can remove this line. The Javascript in `base.bundle.js` is included on every page in the `<head>` of the document, see here:

https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/c6cb705cdda41c025289c21b9461bc32d699d30e/app/recordtransfer/templates/recordtransfer/base.html#L10

app/recordtransfer/static/recordtransfer/css/base/widget.css
```diff
@@ -41,4 +41,4 @@
     .form-tooltip {
         width: 200px;
     }
-}
+}
```
danloveg, line 44: Avoid committing changes to files unnecessarily - this one is very minor so we can ignore it but try to only include changes that are relevant to the issue.

danloveg (changes requested):
This is a good start! The sign up form is a bit broken right now due to the template changes but if you make the changes I've mentioned it should be fixed again.

This is a bit of a wider discussion not related to this pull request specifically, but I'd like to stress the fact that a pull request should only include changes that are directly relevant to the issue being addressed. If there is some change you need to make to a different file that isn't relevant to the issue being fixed, then we can open up as many new issues as we need to to address all the issues in our code. For example, you found that `currentFormStep` is being set inside a `<script>` in the template itself, this isn't ideal and is worth refactoring. But to do that refactoring will require a separate issue and a separate pull request.

The reason I'm so adamant about this is for two main reasons:

1. The larger a pull request is, the more complex it is to review and the longer it takes me to provide feedback.
2. Each pull request turns into a separate commit on `master` (because I squash each pull request). The name of the pull request will reference the issue, and so that commit should only contain changes relevant to that issue. This is useful for when we need to track down when something changed in the past.

A perfect pull request is one where the changes made are surgical: meaning, only the bare minimum changes to address that issue have been made in that pull request.

Eashana0104:
I feel I learnt alot from making these changes, I honestly tried to only keep relevant changes in the pull request, Its probably just because first I made everything work keeping the code for tooltips in widgets.js , and then when that worked after that I moved the code to base directory in tooltips.js. But yeah what you said made sense, I have reverted the changes and now the pull request looks much cleaner. Now I also know that for the purpose of reviewing you look at files changed. Also moving forward, I will create other issues instead of making all the changes in one PR. I think tooltips work in the form now as well.

Eashana0104:
One thing i am encountering is, when I hover over confirm password, the tooltip doesn't always work for some reason.

danloveg:
> I feel I learnt alot from making these changes, I honestly tried to only keep relevant changes in the pull request, Its probably just because first I made everything work keeping the code for tooltips in widgets.js , and then when that worked after that I moved the code to base directory in tooltips.js. But yeah what you said made sense, I have reverted the changes and now the pull request looks much cleaner. Now I also know that for the purpose of reviewing you look at files changed. Also moving forward, I will create other issues instead of making all the changes in one PR. I think tooltips work in the form now as well.

Sounds good, thanks!

> One thing i am encountering is, when I hover over confirm password, the tooltip doesn't always work for some reason.

Are there any errors in the console when that happens? You may also need to re-build the .js bundles, you can do this by running this command in the container:

```shell
npm run watch
```

(If you're not already doing that)

Another thing to try is doing a hard-refresh in your browser. Sometimes your browser will cache the old .js before changes were made to it

danloveg:
> Yeah refreshing a few times worked i want to say. It's still acting bit weird. I am using the watch command though. When I am going on signupform for the first time, hovering on tooltip for confirm password work, but when I hover again, it doesn't. If I refresh same thing happens, it works for the first time but doesn't later. It could be just on my machine though.

I just tested it and mine's doing that, too. Let's take a look at that when we meet at 10

app/recordtransfer/templates/recordtransfer/submission_form_base.html
Eashana0104: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/561/commits/076e37744230ee18ae9de0263c84c0fa06078f6e

app/recordtransfer/templates/includes/submission_group_form.html
Eashana0104: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/561/commits/b9f33b472fc4fbff8d37b38ddf85a91ff2656ce4

app/recordtransfer/static/recordtransfer/js/submission_form/widgets.js
Eashana0104: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/561/commits/8fa0bf0b1b43b20a3538144225a3edeb63fcbab2

app/recordtransfer/static/recordtransfer/js/submission_form/unsavedChanges.js
Eashana0104: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/561/commits/836e64e5ab9dc128ee251cbdca1dca47eef2661d

app/recordtransfer/templates/recordtransfer/signupform.html
Eashana0104: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/561/commits/fb2aa83c8e3a11340cbdc5b35878dbd226b64431

app/recordtransfer/templates/recordtransfer/signupform.html
Eashana0104: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/561/commits/fb2aa83c8e3a11340cbdc5b35878dbd226b64431

app/recordtransfer/templates/recordtransfer/signupform.html
Eashana0104: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/561/commits/ef9bab4acdce0cedfffc595fb938ff2a6b469f87

app/recordtransfer/templates/recordtransfer/signupform.html
Eashana0104: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/561/commits/ef9bab4acdce0cedfffc595fb938ff2a6b469f87

Eashana0104:
I moved css code for tooltip here and added z-index = 1000 to make the tooltip pop up and not hidden by different layers.

app/recordtransfer/static/recordtransfer/css/base/tooltip.css

app/recordtransfer/static/recordtransfer/css/base/tooltip.css
```diff
@@ -0,0 +1,43 @@
...
+    font-size: 11pt;
+    background-color: #333333;
+    color: white;
+    border: 5px;
+    border-radius: 5px;
+}
+
+.help-tooltip.help-icon {
+    z-index: 1000;
+}
```
danloveg, line 25: One last thing, can you delete the `.help-icon` class from `submission_form.css` and move its styles here?

```suggestion
.help-tooltip.help-icon {
    z-index: 1000;
    margin-left: 10px;
}
```

danloveg (changes requested):
Works great now! I noticed there's still a `.help-icon` class in `app/recordtransfer/static/css/base/submission_form.css`, you may as well move that to `tooltip.css` so all the tooltip styles are in the same place.

Once that's done, we're good to go!

Eashana0104:
Everything is good now, This PR is ready to merge.

danloveg (approved):

Issue #536, opened 2025-04-30
Restyle Profile Page using DaisyUI


yenaing-oo:
As discussed in https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/531, the Profile page should be restyled using DaisyUI as a proof of concept to evaluate whether it is a viable component library for this project. The goal is to assess its customisability, and how well-documented and intuitive it is to use.

DaisyUI should initially be included as a CDN, and later on installed as a dependency for just the profile page if deemed to be a viable component library.

danloveg:
I agree with your approach to use a CDN at first. Looking forward to seeing how this turns out

PR #565, opened 2025-05-13
Restyle Profile Page using DaisyUI

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/536

DaisyUI offers the ability to easily customise a theme through CSS: https://daisyui.com/docs/themes/#how-to-add-a-new-custom-theme. For e.g., setting `--color-primary` and `--color-primary-content` allows us to use these colors to style components through CSS classes such as `btn-primary` and `text-primary-content`.

DaisyUI provides a variety of convenient components, although lack more complex components such as sortable/filterable data tables or text editors. Their documentation is adaquate and easy to understand. Since DaisyUI is simply a plugin built upon Tailwind CSS, we get access to Tailwind classes as well to manage things like layout, typography, and responsiveness. Tailwind itself is intuitive to use, and the documentation for it is extensive.

### DIFF
--- a/app/recordtransfer/static/recordtransfer/css/base/base.css
+++ b/app/recordtransfer/static/recordtransfer/css/base/base.css
@@ -1,3 +1,7 @@
+textarea,
+input,
+button,
+select,
 html {
     font-family: "Open Sans", sans-serif;
 }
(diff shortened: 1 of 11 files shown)


danloveg (approved):
This is really impressive, great job on this.

It all makes sense, I have no notes! I'm sold on DaisyUI now

Issue #571, opened 2025-05-14
Install DaisyUI as a Project Dependency


yenaing-oo:
DaisyUI should be installed in the project as an `npm` dependency instead of being used via a CDN. DaisyUI and Tailwind CSS should be made available to every page, i.e. included in `base.css`, instead of just the Profile page (`profile.html`).

See here for installation guide: https://daisyui.com/docs/install/

PR #577, opened 2025-05-15
Install DaisyUI as a Project Dependency

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/571

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -58,7 +58,10 @@ module.exports = {
     watchOptions: {
         aggregateTimeout: 500,
         poll: 1000,
-        ignored: "**/{node_modules,dist}/**"
+        ignored: [
+            "**/node_modules/**",
+            "**/dist/**",
+        ],
     },
     entry: {
         images: [
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -58,7 +58,7 @@ module.exports = {
     watchOptions: {
         aggregateTimeout: 500,
         poll: 1000,
-        ignored: /node_modules/,
+        ignored: "**/{node_modules,dist}/**"
     },
     entry: {
         images: [
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -120,7 +120,7 @@ module.exports = {
         rules: [
             {
                 test: /\.css$/,
-                use: [MiniCssExtractPlugin.loader, "css-loader"] // Extract CSS
+                use: [MiniCssExtractPlugin.loader, "css-loader", "postcss-loader"] // Extract CSS
             },
             {
                 test: /\.(jpe?g|png|webp)$/i,
--- a/Containerfile
+++ b/Containerfile
@@ -27,7 +27,7 @@ ARG WEBPACK_MODE
 ENV WEBPACK_MODE ${WEBPACK_MODE}
 
 # Run webpack to bundle and minify assets
-COPY webpack.config.js ${PROJ_DIR}
+COPY webpack.config.js postcss.config.mjs ${PROJ_DIR}
 RUN npm run build
 
 # Copy entrypoint script to image
--- a/app/recordtransfer/static/recordtransfer/css/base/base.css
+++ b/app/recordtransfer/static/recordtransfer/css/base/base.css
@@ -1,35 +1,3 @@
-@import "tailwindcss";
-
-@plugin "daisyui";
-
-@plugin "daisyui/theme" {
-    name: "nctr";
-    default: true;
-    prefersdark: false;
-    color-scheme: light;
-    /* NCTR Custom Theme Colors */
-
-    --color-base-100: rgb(255, 255, 255);
-    --color-base-200: rgb(250, 250, 250);
-    --color-base-300: rgb(242, 242, 242);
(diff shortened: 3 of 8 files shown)


yenaing-oo:
@danloveg 

Could you take a look at my setup for installing DaisyUI + Tailwind CSS?

The only issue I'm having right now is that the webpack watcher is constantly recompiling for some reason, without any changes being made to the static files. I tried setting `poll` to false, but it kept compiling. It must have something to do with how the `[postcss-loader` Webpack plugin](https://webpack.js.org/loaders/postcss-loader/) works. Perhaps you could help me fix this issue - aside from this, eveything works fine.

DaisyUI itself is simple to install, but it needs Tailwind to be installed first:
https://daisyui.com/docs/install/?lang=en
> How to install daisyUI as a Tailwind CSS plugin?
> 
> You need [Node.js](https://nodejs.org/en/download/) and [Tailwind CSS](https://tailwindcss.com/docs/installation/) installed.

For Tailwind setup, I followed some parts of that [guide](https://gist.github.com/bradtraversy/1c93938c1fe4f10d1e5b0532ae22e16a) you shared with me. It's a bit out of date since the instructions are meant for v3 of Tailwind. I had to reference some parts of Tailwind's [installation guide using PostCSS](https://tailwindcss.com/docs/installation/using-postcss), although it doesn't include any instructions for bundling with webpack. You can find the changes from v3 [here](https://tailwindcss.com/docs/upgrade-guide#changes-from-v3).

danloveg:
> @danloveg
> 
> Could you take a look at my setup for installing DaisyUI + Tailwind CSS?
> 
> DaisyUI itself is simple to install, but it needs Tailwind to be installed first: https://daisyui.com/docs/install/?lang=en
> 
> > How to install daisyUI as a Tailwind CSS plugin?
> > You need [Node.js](https://nodejs.org/en/download/) and [Tailwind CSS](https://tailwindcss.com/docs/installation/) installed.
> 
> For Tailwind setup, I followed some parts of that [guide](https://gist.github.com/bradtraversy/1c93938c1fe4f10d1e5b0532ae22e16a) you shared with me. It's a bit out of date since the instructions are meant for v3 of Tailwind. I had to reference some parts of Tailwind's [installation guide using PostCSS](https://tailwindcss.com/docs/installation/using-postcss), although it doesn't include any instructions for bundling with webpack. You can find the changes from v3 [here](https://tailwindcss.com/docs/upgrade-guide#changes-from-v3).
> 
> The only issue I'm having right now is that the webpack watcher is constantly recompiling for some reason, without any changes being made to the static files. I tried setting `poll` to false, but it kept compiling. It must have something to do with how the [`postcss-loader` Webpack plugin](https://webpack.js.org/loaders/postcss-loader/) works. Perhaps you could help me fix this issue - aside from this, eveything works fine.

Do you know where all the deprecated packages that were added came from?

Screenshot 2025-05-15 153843

One of them is asking for NodeJS <= 9, that's pretty ancient at this point.

yenaing-oo:
> The only issue I'm having right now is that the webpack watcher is constantly recompiling for some reason, without any changes being made to the static files. I tried setting `poll` to false, but it kept compiling. It must have something to do with how the [`postcss-loader` Webpack plugin](https://webpack.js.org/loaders/postcss-loader/) works. Perhaps you could help me fix this issue - aside from this, eveything works fine.

I think I figured out the issue that's causing the infinite recompilation. I found this in the docs for Tailwind v2: https://v2.tailwindcss.com/docs/just-in-time-mode#styles-rebuild-in-an-infinite-loop

I believe there's an infinite loop where:
1) Webpack triggers a build, running PostCSS to process our CSS files
2) PostCSS generates CSS output files in `dist/`
3) Any generated CSS file that made use of PostCSS is, for some reason, not ignored and detected by Webpack's file watcher
4) Webpack watcher sees these file changes and thinks: "Something changed, I need to rebuild."
5) This triggers another PostCSS run, which generates the same CSS files again

To fix this, I added `dist/` to the list of ignored directories in `watchOptions` in the webpack config. (see https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/577/commits/fd4b98e3529a88d71d6a8b4c2fa15b548db71d98_

webpack.config.js
```diff
@@ -58,7 +58,7 @@ module.exports = {
     watchOptions: {
         aggregateTimeout: 500,
         poll: 1000,
-        ignored: /node_modules/,
+        ignored: "**/{node_modules,dist}/**"
```
danloveg, line 61: I think this would be a bit more readable like this:

```suggestion
        ignored: [
            "**/node_modules/**",
            "**/dist/**",
        ],
```

danloveg (changes requested):
Glad to see this is working now, that makes complete sense that we would want to ignore the output folder.

I had one tiny readability-related comment but other than that this is working well.

webpack.config.js
yenaing-oo: This makes sense to me.

danloveg (approved):

PR #578, opened 2025-05-16
Make mobile navigation menu fixed

Eashana0104:
I made the nav-items-container and nav-toggle-button wrap into nav-wrapper. Did so that they both can be stay fixed and scrolled together. Sticky wasn't working very well. Fixed worked because it bypasses layout issues , it’s immune to parent styling.


danloveg (approved):
Looks great, no comments. I see you've already noted the icon colour needing to be changed here: #579. Thanks for opening a new issue for that!

PR #581, opened 2025-05-16
558 convert paginate by to a setting

Eashana0104:


app/recordtransfer/tests/unit/views/test_profile.py
Eashana0104, line 1: I only meant to override the settings paginate_by but when I saved the document, the following changes were also made, If any issues I could just rollback these extra changes

danloveg: That's completely fine in this instance, Ruff will auto-format the Python files we work on. The issue is when we commit auto-formatted changes to files we aren't working on. My goal is for all the Python files to slowly be fixed with proper auto-formatting so this is great!

app/recordtransfer/tests/unit/views/test_profile.py
```diff
@@ -7,7 +7,7 @@
 
 from django.conf import settings
 from django.contrib.messages import get_messages
-from django.test import TestCase
+from django.test import TestCase, override_settings
```
danloveg, line 10: This is a great utility you found, that works much better than `@patch`!

danloveg (changes requested):
This is great! I just had a small comment about where the documentation you wrote should be. That setting doesn't really fit within any of the existing setting headings, so I'm going to suggest you create a new "General Settings" heading and put the new setting's documentation there.

docs/settings/index.rst
```diff
@@ -1507,4 +1507,28 @@ SELENIUM_TESTS_HEADLESS_MODE
...
+    This setting controls how many items are shown per page in list views that support pagination.
+    You can adjust it to improve usability depending on the context (e.g., fewer items for mobile
+    views or more items for data-heavy admin views).
+
+    **.env Example:**
+
+    ::
+
+        # file: .env
+        PAGINATE_BY=20
```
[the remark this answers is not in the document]
Eashana0104: Is this how you thought of it to be https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/581/commits/82b8c25e4679b8d9b6c04d1893b977fd17e20add? If so, this commit resolves this. This makes sense though , I just couldn't figure out where to put it

docs/settings/index.rst
```diff
@@ -8,6 +8,30 @@ below.
 
 .. contents:: List of Settings
 
+PAGINATE_BY
+^^^^^^^^^^^
```
danloveg, line 12: This setting is in the right place now, it just needs a heading, like this:

```suggestion
General Settings
----------------

PAGINATE_BY
^^^^^^^^^^^
```

danloveg (changes requested):

docs/settings/index.rst
Eashana0104: Yeah, I did that. For some reason, It didn't get pushed. Sorry about that. Although I did it now, it should be good.

Eashana0104: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/581/commits/112a9a8e7ad0b2db3b984b66d9c31e7f4df587dd

Issue #574, opened 2025-05-14
Restyle Submission Group Detail Page


yenaing-oo:
The submission group detail page (`submission_group_detail.html`) should be restyled using DaisyUI. Restyling of the confirm modal should be carried out in a separate issue.

Blocked by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/571 and possibly https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/570 as the same form is used here as well.

danloveg:
This will require Tailwind+DaisyUI in the base bundle or will require the profile bundle to be included on this page. I'm not sure which makes the most sense

yenaing-oo:
> This will require Tailwind+DaisyUI in the base bundle or will require the profile bundle to be included on this page. I'm not sure which makes the most sense

I think in order to make UI changes gradually, temporarily including the profile bundle on this page might be a better choice.

PR #582, opened 2025-05-20
Modal to Delete In-Progress Submissions

### LINKED ISSUE #573: Modal to Delete In-Progress Submissions
Similar to https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/572, a [DaisyUI modal](https://daisyui.com/components/modal/) should be used to delete in-progress submissions from the Profile page. 

The endpoint and corresponding views to delete an in-progress submission already exists (see `submission/in-progress/<uuid:uuid>/` in `urls.py` and `views.pre_submission.DeleteInProgressSubmission`), so this issue deals with the UI only.

yenaing-oo:
Closes #573 

This PR adds the ability to delete an in-progress submission directly from the Profile page, by using a modal to ask the user for confirmation, instead of opening up a new page to do that. I have added [Toasts](https://daisyui.com/components/toast/) to notify users of a successful or failed deletion. All UI changes are limited to the Profile page currently.

Change notes:
- Pagination view now includes validation (see https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/582/commits/ee0a40662b14ee4a98508e46199c67e285c97c9a)
- Added a view and corresponding template to render a modal to delete an IP submission
- Scroll position on profile page is no longer saved
- Last active tab for tables is now saved and stored
- Removes modal on Submission Group Detail page


yenaing-oo:
Here's a summary of how it works:
- Clicking on an in-progress (IP) submission's delete button triggers HTMX to send a GET request to `delete_in_progress_modal`. HTMX fills the initially empty modal container with the response contents, then shows the modal when the "afterSwap" event is fired on `#modal-container`
- The "Yes" button to confirm delete has the `hx-delete` attribute, which sends a DELETE request for that IP submission upon click
- On the server side, `profile.delete_in_progress_submission()` handles the request and triggers a client event (either `showSuccess` or `showError`) using [`trigger_client_event`](https://django-htmx.readthedocs.io/en/latest/http.html#django_htmx.http.trigger_client_event) to include a message as well
- An event listener is set up to listen for the [htmx:afterRequest](https://htmx.org/events/#htmx:afterRequest) event, which fires regardless of whether the delete request was successful or not. The handler refreshes the table using [`htmx.ajax`](https://htmx.org/api/#ajax)
- Other event listeners exist that is set up upon page load, which listens for events to show a Toast (see `toast.js`)

yenaing-oo:
@danloveg 
This PR is now ready for a review.

app/recordtransfer/tests/unit/views/test_profile.py
```diff
@@ -508,3 +509,160 @@ def test_submission_group_table_pagination(self) -> None:
...
+        self.assertTrue(InProgressSubmission.objects.filter(uuid=self.in_progress.uuid).exists())
+
+    def test_cannot_delete_other_users_submission(self) -> None:
+        """Test that a user cannot delete another user's in-progress submission."""
+        other_url = reverse(
+            "recordtransfer:in_progress_submission", kwargs={"uuid": self.other_in_progress.uuid}
+        )
+
+        response = self.client.delete(other_url, headers=self.headers)
+        self.assertEqual(response.status_code, 500)
```
danloveg, line 603: I think this should be a 404 instead of a 500

app/recordtransfer/templates/includes/delete_in_progress_submission_modal.html
```diff
@@ -0,0 +1,24 @@
...
+    delete_in_progress_submission_modal
+{% endblock modal_id %}
+{% block modal_title %}
+    {% trans "Are you sure you want to delete this in-progress submission?" %}
+{% endblock modal_title %}
+{% block modal_content %}
+    <p>
+        {% trans "Do you wish to delete the in-progress submission from" %} {{ in_progress.last_updated|date:"F j, Y g:ia" }}
+        {% if in_progress.title %}titled "{{ in_progress.title }}"{% endif %}
+        ?
```
danloveg, line 13: This is split up in a way that makes translation difficult. How about changing it to this?

```suggestion
        {% if in_progress.title %}
            {% blocktrans with title=in_progress.title date=in_progress.last_updated|date:"F j, Y g:ia" %}
            Do you wish to delete the in-progress submission from {{ date }} titled "{{ title }}"?
            {% endblocktrans %}
        {% else %}
            {% blocktrans with date=in_progress.last_updated|date:"F j, Y g:ia" %}
            Do you wish to delete the in-progress submission from {{ date }}?
            {% endblocktrans %}
        {% endif %}
```

app/recordtransfer/views/profile.py
```diff
@@ -106,6 +114,34 @@ def form_invalid(self, form: BaseModelForm) -> HttpResponse:
...
+        in_progress.delete()
+        response = HttpResponse(status=204)
+        return trigger_client_event(
+            response, "showSuccess", {"value": "In-progress submission deleted."}
+        )
+    except Exception:
+        response = HttpResponse(status=500)
+        return trigger_client_event(
+            response, "showError", {"value": "Failed to delete in-progress submission."}
+        )
```
danloveg, line 142: I think get_object_or_404 should be used here, too. You'll have to change the Exception to let those exceptions through because that function raises an `Http404` exception. If you leave it as `except Exception` you'll mask the `Http404` exception as a 500 error

danloveg (changes requested):

app/recordtransfer/tests/unit/views/test_profile.py
yenaing-oo: This makes sense, I've made the necessary changes here:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/582/commits/dec95f6e009394c32dde36a1db541585834e1205

yenaing-oo:
I noticed I had accidentally removed the handler that DELETEs a submission group, whilst I was removing the confirmation modal on the Submission Group Detail page. I've added that back now, but I've removed the old modal, since it was conflicting with the new modal that I added. A restyled modal will be added back in https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/574
Addressed by: [Restore ability to delete submission group](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/582/commits/94f971a6c8ef4d529c70dd29774feb1c4e9f3316)

Issue #579, opened 2025-05-16
Nav menu bars icon not visible on white background


Eashana0104:
The navigation menu "bars" icon (hamburger menu) becomes invisible or difficult to see when the page background is white or light-colored. This occurs because the icon's default color is white (color: white), which lacks contrast against light backgrounds.

We need a different colour for the button or a background that looks good with the website as well. 

<img width="372" alt="Image" src="https://github.com/user-attachments/assets/436bf732-605f-4f37-8e8d-d644700b1636" />
<img width="328" alt="Image" src="https://github.com/user-attachments/assets/f909df3a-de11-47e9-b4f6-23c98ed5ac3b" />

<img width="267" alt="Image" src="https://github.com/user-attachments/assets/cb566a9a-c111-4615-be30-62c325125a05" />

danloveg:
Closed by #598

Issue #522, opened 2025-04-23
Make mobile navigation menu sticky


danloveg:
The mobile navigation menu disappears to the top of the screen when you scroll down.

![Image](https://github.com/user-attachments/assets/f6afd083-b6d8-4a75-b877-f8bcd987762a)

The menu should be made "sticky" so that the full menu can be seen no matter how far scrolled down the user is. Additionally, the menu icon should stay visible when the user scrolls down, so that the menu can be opened no matter how far scrolled down you are.

PR #584, opened 2025-05-20
Add background color to burger button when on light backgrounds

Eashana0104:
For now, things work as expected, only problem being it taking the color of banner and header into consideration making it add a background which should not happen, have to find a way to fix it. Positioning of the toggle button is handled by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/587

### DIFF
--- a/app/recordtransfer/static/recordtransfer/css/base/banner.css
+++ b/app/recordtransfer/static/recordtransfer/css/base/banner.css
@@ -5,7 +5,7 @@
 }
 
 .banner-text {
-    font-family: Merriweather, serif;
+    font-family: Merriweather,serif;
     font-size: 11pt;
 }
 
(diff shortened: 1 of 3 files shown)


danloveg (commented):
I think it would like nicer if the white button got a background when scrolling down. Like this sample I created:

Screenshot 2025-05-20 155338

I also noticed that the fixed position looks a bit weird in mobile:

Screenshot 2025-05-20 155453

I'm going to re-open ~~#579~~ #522 (sorry, wrong issue number) if you want to take a look at that again, sorry I didn't test the mobile menu at smaller resolutions.

Eashana0104:
I tried it to have a grey background when the background for the website was white, It was causing delays, and sometimes it won't have a background too. I can try adding it specifically to lighter backgrounds tomorrow

PR #587, opened 2025-05-21
522 make mobile navigation menu sticky

Eashana0104:
I changed the height and width of nav container to be according to the screen's height and width, removed flex-wrap from the main graphic text as well by adding media query for smaller resolutions. Nav title and icon gets dynamically resized based on screen size. TO DO - set background for toggle button, would be done in another issue/PR


danloveg:
> I am not sure for how small of a resolution should the website work, I tested for 375px , I think it works for fine for very small resolutions too but if we decrease it any further, some overlapping is there. Does it affect us if the screen size is smaller than this?

Is it possible to have the menu icon in line with the logo at smaller resolution? See this image, I'm hoping that there's a way to have the icon where I drew the three green lines.

Screenshot 2025-05-21 110005

Eashana0104:
Yeah for sure! I ll work on that

danloveg (changes requested):
At a first glance I noticed three things:

1. The navigation bar is broken on larger screens:

Screenshot 2025-05-23 122324

2. The mobile menu looks weirdly disjoint starting at 480px:

Screenshot 2025-05-23 122409

3. The mobile menu jumps between two positions when scrolling, can you add a smoothing transition so that moving the button doesn't look so jarring?

Eashana0104:
Fixed the UI related bugs in https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/587/commits/6f1cb194199dccb0c2b0a8070fc49980150cedd3, https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/587/commits/5c581679ae5d89aa3366b8d2e5b68411c2ae27ef, https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/587/commits/762ec3e9bb2c9d0bb3afec0e95bd2986d1de982d these commits

danloveg (commented):
Here's what I'm noticing now:

https://github.com/user-attachments/assets/a9f63b0c-b6c4-4cc7-b619-a1d2463b583d

*Before you make any more commits*, can you explain what your changes are supposed to do, or what the approach you're taking is? I see you're applying a fixed position when the screen scrolls down, but that won't update when the screen width is increased or decreased. Is there a way for us to set a fixed position with CSS and not have to use Javascript?

Eashana0104:
I did not make any commits yet but I removed the JavaScript code, and fixed the position for the toggle button so it only stays at that position even when the user scrolls, the position is set as top = custom px, depending on the size of the screen using media queries. On all the devices given in the Google Dev Tools, it works as expected. JavaScript code was breaking the nav bar but it was making the code more dynamic. I can try fixing the JavaScript code too or if using fixed in CSS works , I can do that as well.

danloveg:
> I noticed this but I think its happening because its not getting refreshed, I thought since its going to be different devices and the size of the device would not change, this issue wont occur if website is running on a single device from the beginning. I ll try to find an approach that I can use to not use Java script for this issue

It's entirely possible that someone will open the site on desktop, and have to resize their browser to fit another window side-by-side, for example. So we want to make sure none of the styles break when resizing the window.

Here, check out this branch and let me know what you think with this approach: [522-sticky-mobile-menu-alternative](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/tree/522-sticky-mobile-menu-alternative). I used a combination of your changes and some of my own.

Eashana0104:
I didnt even think about how someone else could be resizing our website too. Your approach makes the code cleaner as well, also I checked that approach with all different sizes of screen too and it works perfectly, I think I was over complicating it for no reason. I ll definitely learn from this and try not making such mistakes again.

Eashana0104:
I think you could merge this branch into master and then I can edit the code accordingly for the colour of the toggle button and adding a background, the current code might not work properly right now because of code changes.

danloveg:
> I think you could merge this branch into master and then I can edit the code accordingly for the colour of the toggle button and adding a background, the current code might not work properly right now because of code changes.

I'll close this PR and open a new one with my changes.

Issue #570, opened 2025-05-14
Use Modal to Create New Submission Groups on Profile Page


yenaing-oo:
Currently, a separate page (`submission-group/`) is used to display a form to create new submission groups. A DaisyUI [modal](https://daisyui.com/components/modal/) should be used to do this directly from the Profile page. This means that after this change, the url path `/submission-group` should be removed from `urls.py`. 

If `submission_group_form.html` is restyled, note that this will affect the Submission Form, where `submission_group_form.html` is used as well in the Group Submission step template (`submission_form_groupsubmission.html`). Note that this means that DaisyUI will have to be available to the Submission Form page as well.

The page to view and modify the details of a submission group should be kept separate and left as is (`submisison-group/{uuid}`)

PR #593, opened 2025-05-22
Use Modal to Create New Submission Groups on Profile Page

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/570

Change notes:
- Adds ability to create a new submission group using a modal directly from the Profile page, instead of loading a separate page
- Changes the way modals are rendered: modal's contents are being replaced now instead of the entire modal itself. This is to prevent the modal from flickering.

### DIFF
--- a/app/recordtransfer/tests/unit/views/test_profile.py
+++ b/app/recordtransfer/tests/unit/views/test_profile.py
@@ -676,7 +676,7 @@ class TestSubmissionGroupModalCreateView(TestCase):
         self.user = User.objects.create_user(username="testuser", password="password")
         self.client.login(username="testuser", password="password")
 
-        self.url = reverse("recordtransfer:submission_group_modal")
+        self.submission_group_modal_url = reverse("recordtransfer:submission_group_modal")
         self.headers = {
             "HX-Request": "true",
         }
@@ -689,7 +689,7 @@ class TestSubmissionGroupModalCreateView(TestCase):
             "name": "Test Group",
             "description": "Test Description",
         }
-        response = self.client.post(self.url, data=form_data, headers=self.headers)
+        response = self.client.post(self.submission_group_modal_url, data=form_data, headers=self.headers)
         self.assertTrue(SubmissionGroup.objects.filter(name="Test Group").exists())
 
         self.assertIn("HX-Trigger", response.headers)
@@ -701,7 +701,7 @@ class TestSubmissionGroupModalCreateView(TestCase):
             "name": "",
             "description": "Test Description",
         }
-        response = self.client.post(self.url, data=form_data, headers=self.headers)
+        response = self.client.post(self.submission_group_modal_url, data=form_data, headers=self.headers)
         self.assertEqual(response.status_code, 200)
         self.assertTemplateUsed(response, "includes/new_submission_group_modal.html")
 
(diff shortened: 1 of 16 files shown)


danloveg (commented):
It's not working right for me. When I click "Add New Submission Group" my screen looks like this:

Screenshot 2025-05-23 144039

And then when I click back in my browser, it brings me to `/submission/`

In the Network tab of my browser, I notice that a request to `/submission-group/` is sent when I click the button, not to `/submission-group-modal/`, could it have something to do with that?

yenaing-oo:
> It's not working right for me. When I click "Add New Submission Group" my screen looks like this:
> 
> And then when I click back in my browser, it brings me to `/submission/`
> 
> In the Network tab of my browser, I notice that a request to `/submission-group/` is sent when I click the button, not to `/submission-group-modal/`, could it have something to do with that?

Ah I somehow missed updating the reference to the modal path on the "Add New Submission Group" button. It should be working fine now.

danloveg (commented):
I haven't dug deep into the code yet, I'll do that after this is addressed:

You can't click Cancel to close the modal, it asks you to fill out the Group Name. You *can* click the X to close it but you should be able to click Cancel, too. This screenshot is what I got after I tried to click Cancel:

Screenshot 2025-05-23 151136

yenaing-oo:
> I haven't dug deep into the code yet, I'll do that after this is addressed:
> 
> You can't click Cancel to close the modal, it asks you to fill out the Group Name. You _can_ click the X to close it but you should be able to click Cancel, too. This screenshot is what I got after I tried to click Cancel:


That's a good find. I've fixed that here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/593/commits/e9e6e6d19754bf40b0bfaf5f82a117f2b5b51e08

app/recordtransfer/tests/unit/views/test_profile.py
```diff
@@ -666,3 +666,58 @@ def test_nonexistent_submission(self) -> None:
...
+
+class TestSubmissionGroupModalCreateView(TestCase):
+    """Tests for SubmissionGroupModalCreateView."""
+
+    def setUp(self) -> None:
+        """Set up test environment."""
+        self.user = User.objects.create_user(username="testuser", password="password")
+        self.client.login(username="testuser", password="password")
+
+        self.url = reverse("recordtransfer:submission_group_modal")
```
danloveg, line 679: Same comment as my other review, this would be easier to read if it was named something more descriptive like `submission_group_modal_htmx_url`

danloveg (changes requested):
Awesome, no comments beside the one change in the test I'm suggesting.

I see you've already worked out what's left for cleanup. I'm looking forward to seeing the delete button in the submission group table and having a modal for that, too.

app/recordtransfer/tests/unit/views/test_profile.py
yenaing-oo: I agree, I've made that change here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/593/commits/07a69a20557277046dfeddb14a3276faf55102ce

danloveg (approved):
Looks good, have a good weekend!

Issue #580, opened 2025-05-16
Ensure Usage of `get_object_or_404` Filters for User


yenaing-oo:
There are instances where `get_object_or_404` is being used without filtering for the user making the request.

For e.g., 

```python
return get_object_or_404(Submission, uuid=self.kwargs.get("uuid"))
```
**does not** check that the submission belongs to the user making the request. It only checks whether a Submission with the given uuid exists. It does not filter by user ownership.

To ensure that the submission belongs to the current user, add a filter for the user: 
```python
return get_object_or_404(
    Submission,
    uuid=self.kwargs.get("uuid"),
    user=self.request.user
)
```

PR #594, opened 2025-05-23
Simplify Permission Handling for Post Submission Views

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/580

The two instances where `get_object_or_404` were being used without filtering for the user did not actually pose any security risks, as the view utilized `test_func()` as part of `UserPassesTestMixin` to test that the object requested belonged to `request.user`. However, this is not ideal, since we want to prevent any objects that don't belong to `request.user` from making it to the view in the first place.

I've made use of user filtering directly in `get_queryset()`, so that it returns a queryset appropriate to the user or staff member.

I've also added missing tests for `SubmissionDetailView` and `SubmissionCsvView`.

### DIFF
--- a/app/recordtransfer/tests/unit/views/test_post_submission.py
+++ b/app/recordtransfer/tests/unit/views/test_post_submission.py
@@ -1,5 +1,7 @@
 from unittest.mock import MagicMock, patch
+
 from django.contrib.messages import get_messages
+from django.http import HttpResponse
 from django.test import TestCase
 from django.urls import reverse
 from django.utils.translation import gettext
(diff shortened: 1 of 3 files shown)


app/recordtransfer/tests/unit/views/test_post_submission.py
```diff
@@ -194,3 +196,120 @@ def test_delete_submission_group_error(self, mock_delete: MagicMock) -> None:
...
+    def setUp(self) -> None:
+        """Set up test environment."""
+        self.user = User.objects.create_user(username="testuser", password="password")
+        self.staff_user = User.objects.create_user(
+            username="staffuser", password="password", is_staff=True
+        )
+        self.submission = Submission.objects.create(user=self.user)
+        self.url = reverse(
+            "recordtransfer:submission_detail", kwargs={"uuid": self.submission.uuid}
+        )
```
danloveg, line 213: When I read the tests, I get confused remembering what `self.url` is and find I keep having to go back to `setUp` to remember what it was. Could you change this to `self.submission_detail_url` or something like that?

app/recordtransfer/tests/unit/views/test_post_submission.py
```diff
@@ -194,3 +196,120 @@ def test_delete_submission_group_error(self, mock_delete: MagicMock) -> None:
...
+    """Tests for SubmissionCsvView."""
+
+    def setUp(self) -> None:
+        """Set up test environment."""
+        self.user = User.objects.create_user(username="testuser", password="password")
+        self.staff_user = User.objects.create_user(
+            username="staffuser", password="password", is_staff=True
+        )
+        self.submission = Submission.objects.create(user=self.user)
+        self.url = reverse("recordtransfer:submission_csv", kwargs={"uuid": self.submission.uuid})
```
danloveg, line 271: Same comment as above, maybe this variable can be called `submission_csv_url`

danloveg (changes requested):
This makes sense, thanks for looking into this. I have a couple questions about test readability but functionality-wise everything looks good to me

app/recordtransfer/tests/unit/views/test_post_submission.py
yenaing-oo: This makes sense, I'll make these changes.

yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/594/commits/2f9f0129e7d35366a7ff16981e3c60a97e66967e

yenaing-oo:
Additionally, I've updated the URL variables for `TestSubmissionGroupCreateView` and `TestSubmissionGroupDetailView` in a similar fashion.

danloveg (approved):
The tests are much more readable now, thanks! This looks good.

PR #597, opened 2025-05-23
Make mobile menu stick to top when scrolled down

### LINKED ISSUE #522: Make mobile navigation menu sticky

danloveg:
Closes #522 

This change makes the mobile menu `position: fixed;` and ensures it stays at ~ 20px from the top of the screen. I've also made it so that the menu stays aligned with the title in the header so that the positioning of the menu icon looks good when scrolled all the way to the top.

This is a combo of my own changes and @Eashana0104's (see #587)

### DIFF
--- a/app/recordtransfer/static/recordtransfer/css/base/navbar.css
+++ b/app/recordtransfer/static/recordtransfer/css/base/navbar.css
@@ -182,6 +182,22 @@ header.homepage {
     transition: background-color 0.3s ease, color 0.3s ease;
 }
 
+@media all and (max-width: 450px) {
+    .nav-title {
+        margin-left: 10px;
+    }
+
+    #app-logo {
+        width: 35px;
+        height: 35px;
+    }
+
+    #app-title {
+        font-size: 16pt;
+    }
+}
+
+
 /* Mobile-only transitions */
 @media all and (max-width: 799px) {
     .nav-items-container {
--- a/app/recordtransfer/static/recordtransfer/css/base/navbar.css
+++ b/app/recordtransfer/static/recordtransfer/css/base/navbar.css
@@ -52,6 +52,16 @@ header.homepage {
     align-content: start;
     justify-content: flex-start;
     text-align: left;
+    /* Add padding to account for fixed position mobile menu */
+    padding-right: 40px;
+}
+
+/* Mobile navigation container - wraps both toggle and menu */
+.nav-wrapper {
+    position: fixed;
+    top: 20px;
+    right: 0;
+    z-index: 1001;
 }
 
 /* Sliding menu styles */
@@ -75,7 +85,7 @@ header.homepage {
 }
 
 .nav-items-container.open {
-    top: 80%;
+    top: 100%;
     visibility: visible;
     transition: top 0.3s ease, opacity 0.3s ease;
 }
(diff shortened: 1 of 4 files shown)


danloveg:
> Remember we were having a conversion about how if flex wrap is turned on the title gets into he next line and it doesn't look good 
>
> 📷
>
> Should I turn off the flex wrap? I created a function in my branch that would dynamically change the size of title and logo too depending on the screen size. Should I use that here ?

I've made a change in https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/597/commits/3ad2e4d9a906198a23e75737b7217b3d61f82157 that keeps the header looking good down to 320px (which is very small). Let me know how that looks

app/recordtransfer/templates/recordtransfer/header.html
```diff
@@ -75,6 +69,12 @@
                         {% endif %}
                     {% endif %}
                 </div>
+                <div class="nav-toggle-open">
+                    <a class="nav-link non-nav-link nav-toggle-button" href="#">
+                        <i class="fa fa-bars nav-icon-bars" aria-hidden="true"></i>
+                        <i class="fas fa-times nav-icon-close" aria-hidden="true"></i>
+                    </a>
+                </div>
             </div>
```
Eashana0104, line 78: I wanted to ask why you moved the nav-toggle-open div below the nav-items-container. I know HTML reads from top to bottom but what difference did this change make?

app/recordtransfer/static/recordtransfer/css/base/navbar.css
```diff
@@ -210,6 +201,16 @@ header.homepage {
         background: none;
     }
 
+    /* Reset mobile container for desktop */
+    .nav-wrapper {
+        position: static;
```
Eashana0104, line 222: I also really like how just using static can reset the page automatically, I think this was one of the issues I was facing, which was messing with the positioning of the nav bar

app/recordtransfer/templates/recordtransfer/header.html
danloveg: What I did was revert the changes back to v2025.04.30, and then made changes from there. In that version, the toggle was at the bottom. I think it makes more sense to move it to the top especially since this header is mobile-first, I'll make that change.

Eashana0104:
I see the changes. So there is no need to remove the flex wrap then as it looks good?

app/recordtransfer/templates/recordtransfer/header.html
danloveg: Addressed in https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/597/commits/9dbad1874c6b4745bc234f644db0f6b99c87ad0e

danloveg:
> I see the changes. So there is no need to remove the flex wrap then as it looks good?

If you think it looks good then yes we can just leave that flex wrap as it is.

Eashana0104 (approved):
Everything looks good now, thanks for helping me with this issue. I tested it for quite a bit and I think I can approve this now.

PR #598, opened 2025-05-23
Add background to toggle button so it shows on white background

Eashana0104:


app/recordtransfer/static/recordtransfer/js/base/navbar.js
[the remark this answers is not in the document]
danloveg: > I ll look into this. So this code works when it basses the header ie the main graphic, our footer is red too where white toggle bars are fully visible, should I add something like if the midpoint hasn't passed the footer, needs-background stays? What is the expected behaviour in that case?

The grey against the red still has enough contrast so I think it's OK to ignore the footer. See:

Screenshot 2025-05-26 105741

> Also, I did not quite understand the background part. If we change the background, suppose from white we did to some dark color, it would adjust the needs background automatically, right? I do think that your solution is very simple, but what if we change the background of our website altogether? Suppose we change it to dark mode, the background for the toggle button would still show then, right? I am not sure if I am thinking about this the right way>

If we made a dark mode of the website, we would also need to create an alternate version of the `.nav-wrapper.needs-background` class that works with the dark background. My solution is assuming that the background of the site will stay light which is a fair assumption.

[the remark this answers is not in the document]
Eashana0104: Okay i will add the background again then to the footer or should I keep it removed? Lemme know what you think looks better

app/recordtransfer/static/recordtransfer/js/base/navbar.js
```diff
@@ -61,9 +61,37 @@ export function setupNavbar() {
...
+            topElement = graphic.getBoundingClientRect();
+        } else if (header && !graphic) {
+            topElement= header.getBoundingClientRect();
+        }
+
+        if (navMiddle > topElement.bottom && navMiddle < footerRect.top) {
+            navWrapper.classList.add("needs-background");
+        } else {
+            navWrapper.classList.remove("needs-background");
+        }
```
danloveg, line 89: I like this solution better than mine, let's go with this.

Three things before this is good to be merged into `master`:

1. Can you rename the function `alignNavWrapper` to something more appropriate now that we've added the code that updates the background? Maybe something like `refreshMobileMenu`
2. The docstring for the function should be updated to reflect the additional functionality.
3. Can you auto-format the file after making those changes? (i.e., CTRL+Shift+P, "Format Document" in VS Code)

danloveg (changes requested):

app/recordtransfer/static/recordtransfer/js/base/navbar.js
Eashana0104: On it!

Eashana0104: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/598/commits/ba50a8c5f78e1cb42f9c48185cb3822d56bf5bdf. I did the format document option, it didn't change anything for me. I believe it automatically formats on save as well right?

danloveg: Thanks! And yeah I realize I may be using a different JS formatter, all good. We can sync our environments up in that regard in another issue.

danloveg (approved):

Issue #572, opened 2025-05-14
Modal to Delete Submission Groups


yenaing-oo:
Create a modal using DaisyUI (`https://daisyui.com/components/modal/`) to delete submission groups from the Profile page. This will also involve creating a new endpoint to delete a submission group, such as DELETE `submission-group/{uuid}` along with a view to handle this request.

Dependent on: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/536

yenaing-oo:
> [@danloveg](https://github.com/danloveg) [#582](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/582) was closed and merged in with the caveat that the Delete button on the Submission Group Details page was broken. Upon testing, the two issues I see are:
> 
> * A lack of a modal to confirm deletion (which was intentionally removed due to the modal needing restyling with DaisyUI)
> * A duplicate DELETE request being made on the Profile page after successful deletion causes a redirect to the profile page (possibly due to a lingering event listener). You'll have to click on "Preserve log" to see this error. I believe this error existed before the changes made in [Modal to Delete In-Progress Submissions #582](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/582) .
> 
> ![Image](https://github.com/user-attachments/assets/be77a5d3-c286-4658-9766-2ce6f70e45ce)

@danloveg 
I will work of addressing these issues in https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/574

PR #599, opened 2025-05-26
Modal to Delete Submission Groups

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/572

### DIFF
--- a/app/recordtransfer/views/profile.py
+++ b/app/recordtransfer/views/profile.py
@@ -227,7 +227,9 @@ def submission_table(request: HttpRequest) -> HttpResponse:
 
 
 class SubmissionGroupModalCreateView(CreateView):
-    """Renders a modal form to create a new submission group."""
+    """Renders a modal form to create a new submission group. Handles GET requests to show the
+    modal and POST requests to create the submission group. Both requests must be made by HTMX.
+    """
 
     model = SubmissionGroup
     form_class = SubmissionGroupForm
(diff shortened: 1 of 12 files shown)


app/recordtransfer/views/profile.py
```diff
@@ -252,3 +254,35 @@ def form_valid(self, form: BaseModelForm) -> HttpResponse:
...
+    except Http404:
+        response = HttpResponse(status=404)
+        return trigger_client_event(
+            response, "showError", {"value": "Submission group not found."}
+        )
+    except Exception:
+        response = HttpResponse(status=500)
+        return trigger_client_event(
+            response, "showError", {"value": "Failed to delete submission group."}
+        )
```
danloveg, line 288: I just realized that these toast messages should be translated, could you make that change here and also to the other toast messages in this file?

I think the submission group name in the success toast would be nice to have here. My first thought is that you may want to escape (`django.utils.html.escape`) the name before rendering it but not sure whether that's necessary.

```suggestion
        submission_group = get_object_or_404(SubmissionGroup, uuid=uuid, created_by=request.user)

        if request.method == "GET":
            context = {"submission_group": submission_group}
            return render(request, "includes/delete_submission_group_modal.html", context)

        group_name = submission_group.name
        submission_group.delete()
        response = HttpResponse(status=204)
        return trigger_client_event(
            response,
            "showSuccess",
            {
                "value": gettext('Submission group "%(name)s" deleted.')
                % {"name": escape(group_name)}
            },
        )
    except Http404:
        response = HttpResponse(status=404)
        return trigger_client_event(
            response, "showError", {"value": gettext("Submission group not found.")}
        )
    except Exception:
        response = HttpResponse(status=500)
        return trigger_client_event(
            response, "showError", {"value": gettext("Failed to delete submission group.")}
        )
```

yenaing-oo: Translation always seems to sneak past me! I've applied your suggestions for the entire file, including escaping the submission group name. I think it's not necessary to `escape`, since `textContent` automatically escapes HTML, but I think it's good practice to include it anyway, in case the front-end is using the message in other ways.

Addressed by:
- [Translate HTMX response messages](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/599/commits/ea8f2118fedd0920b10906ba9506d13de1e2ee7e)
- [Escape submission group name](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/599/commits/bd69553d555d1364d1dc13830cd717749d376d6b)

yenaing-oo: Good catch. I forgot there was one more there. I've fixed that here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/599/commits/9ab2b3abe1444fdbe2c996349434baf8b52f53f2

danloveg (approved):

PR #606, opened 2025-05-28
Fix duplicate in-progress submissions being created

danloveg:
It's possible for the `resume` GET parameter to be missing when the save_form_step button is clicked. When this happens, the view doesn't know the in-progress UUID and creates a new in-progress submission.

This pull request uses HTMX `hx-vals` to pass the in-progress UUID as a fallback to use in case resume is missing from the GET request.


danloveg:
@yenaing-oo your fix with `get_full_path` to fix the group UUID has fixed this issue, too.

PR #607, opened 2025-05-28
Simplify webpack config with main + vendor bundles

Eashana0104:


app/recordtransfer/static/recordtransfer/js/index.js
```diff
@@ -0,0 +1,49 @@
...
+
+
+
+import "htmx-ext-head-support";
+import htmx from "htmx.org";
+import { setupMessages } from "./base/messages";
+import { setupNavbar } from "./base/navbar";
+import { setupHelpTooltips } from "./base/tooltip";
+import { initialize } from "./profile/index";
+import { setup } from "./submission_form/index";
```
danloveg, line 25: Can you rename these functions to:

```suggestion
import { initializeProfile } from "./profile/index";
import { initializeSubmissionForm } from "./submission_form/index";
```

app/recordtransfer/static/recordtransfer/js/index.js
```diff
@@ -0,0 +1,49 @@
...
+    setupHelpTooltips();
+});
+
+document.addEventListener("htmx:afterSwap", (event) => {
+    if (event.detail.target.id === "main-container") {
+        setup();
+    }
+});
+
+document.addEventListener("DOMContentLoaded", initializeSubmissionGroup);
```
danloveg, line 49: There's an extra unnecessary `"DOMContentLoaded"` event listener at the end of this file.

```suggestion

document.addEventListener("DOMContentLoaded", () => {
    setupNavbar();
    setupMessages();
    initializeProfile();
    initializeSubmissionForm();
    initializeSubmissionGroup();
});

window.addEventListener("load", () => {
    setupHelpTooltips();
});

document.addEventListener("htmx:afterSwap", (event) => {
    if (event.detail.target.id === "main-container") {
        setup();
    }
});
```

danloveg (changes requested):
This is really good, good job on this.

Two things I noticed:

1. The styles on this page are not working at all: http://localhost:8000/submission/sent/ . I'm guessing Tailwind has its own list styles that are removing the basic list styles. Specifically, the links do not look like links. If you'd like to re-style this page with TailwindCSS I think that's OK to include in this pull request since it's a small page.

Screenshot 2025-05-30 152926

2. The text size in `textarea` elements is smaller than it is in `input` elements. Can you make it so that `textarea`s have the same font size as the `input`s?

Screenshot 2025-05-30 152727

And I'll also suggest merging the latest `master` into this branch just to be doubly sure there were no JS changes between the last time you merged master into here and now.

danloveg (approved):
Everything is working as expected. The styles on some of the other pages will also need a refresh but those pages are pretty simple. I'll open up new issues to restyle the others.

PR #626, opened 2025-05-30
Update Login account link

Eashana0104:
Summary:
The issue with the "Log in to your account" button has been resolved — the link was missing /account/ in the path, and it's now correctly directing users.

Additional Note:
While reviewing, I noticed that the submission_submit_success and submission_submit_failure test emails may not send properly because the recipients list is empty. In both functions, the list is instantiated inside the function rather than being passed as a parameter.

Temporary Fix:
I manually populated the recipients list during testing, and the emails were sent successfully.


Eashana0104:
I see what you mean, i did not realise this url was already provided in urls.py. I have commited the suggestions, I tested it too, it works nicely.

danloveg (approved):

Issue #627, opened 2025-05-30
Unable to Access Uploaded Files from Review Step


yenaing-oo:
Clicking on a link to view an uploaded file from the Review step opens up a page with a JSON error:

![Image](https://github.com/user-attachments/assets/a0715281-3ed4-4264-a35a-d08a72ae710f)

![Image](https://github.com/user-attachments/assets/0b7a380c-a522-4f0c-9eb5-195e9d696bca)

PR #629, opened 2025-05-30
Fix Not Being Able to Access Uploaded Files on Review Step

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/627

This issue was caused by the `requires_upload_step` decorator requiring both file upload and file access operations to be restricted to the upload step. I've changed it so that:
1) file uploads and deletes are restricted to just the Upload Files step
2) uploaded file access is restricted to the Upload Files and Review step


danloveg (approved):
Works good!

Issue #628, opened 2025-05-30
Issues with the Unsaved Changes Protection Modal


yenaing-oo:
On a new submission, I'm unable to click on any nav links after:
1) Going to Step 2
2) Going back to Step 1. Note how the "unsaved changes" modal does not show up.

On resuming an in-progress submission:
1) Going back to step 1
2) Clicking on any link to take you to another page
3) The "unsaved changes" modal appears, but the link to save the form does not work

PR #630, opened 2025-05-30
Fix Issues with Unsaved Changes Protection Modal

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/628


yenaing-oo:
@danloveg 
I've managed to fix the first issue, but I'm having trouble explaining why the modal displays fine for all the other steps, and only has an issue for the Step 1.

The second issue with the save link on the modal not working on the first step makes sense to me, because that button/link was piggybacking the save button on the form to perform the save. But that button does not show up if it's on Step 1. I fixed that by making sure it shows up on Step 1 if the form has been edited in any way.

Issue #612, opened 2025-05-29
Move Common Template Constants to a Context Processor


yenaing-oo:
Currently, HTML element IDs and other static values, such as URL query parameter names, are defined in `constants.py`, imported into views, and manually passed into templates as context variables.

I propose moving these constants to a global context processor, or importing them there, so that they are available across all templates without needing to be explicitly passed in each view. I think that this approach is cleaner and reduces code clutter in views.

PR #651, opened 2025-06-02
Move Common Template Constants to a Context Processor

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/612

Change notes:
- Added a context process to make common constants available to templates without having to pass them from the view. Any constants being passed from views to templates as context for JS remains unchanged.
- Fixed bug where selecting "Other" option on "Province or state" select field of Contact Information Step on Submission Form does not display textbox to type in alternative value.


app/recordtransfer/context_processors.py
```diff
@@ -1,9 +1,50 @@
...
+        "ID_SOURCE_INFO_SOURCE_ROLE": constants.ID_SOURCE_INFO_SOURCE_ROLE,
+        "ID_SOURCE_INFO_OTHER_SOURCE_ROLE": constants.ID_SOURCE_INFO_OTHER_SOURCE_ROLE,
+        # Query Name Constants
+        "PAGINATE_QUERY_NAME": constants.PAGINATE_QUERY_NAME,
+        "SUBMISSION_GROUP_QUERY_NAME": constants.SUBMISSION_GROUP_QUERY_NAME,
+        # Pagination Table Container IDs
+        "ID_SUBMISSION_GROUP_TABLE": constants.ID_SUBMISSION_GROUP_TABLE,
+        "ID_IN_PROGRESS_SUBMISSION_TABLE": constants.ID_IN_PROGRESS_SUBMISSION_TABLE,
+        "ID_SUBMISSION_TABLE": constants.ID_SUBMISSION_TABLE,
+    }
```
danloveg, line 50: If you need every variable LIKE_THIS included in the context, you could do something like this, which would automatically include any new constants we add:

```suggestion
CONSTANT_VARIABLE = re.compile(r"[A-Z][A-Z_]+[A-Z]")


def constants_context(request: HttpRequest) -> dict[str, str]:
    """Make constants available globally in all templates."""
    constant_variables = [name for name in dir(constants) if CONSTANT_VARIABLE.match(name)]

    return {
        name: getattr(constants, name)
        for name in constant_variables
        if isinstance(getattr(constants, name), str)
    }
```

I don't believe the performance here would be very good but since this data doesn't change during runtime you could probably calculate this once and then return the same dict every time.

danloveg (changes requested):
There's got to be a simpler way to get all these constants into the context. I've suggested a way here, which feels a little bit hacky, but has the added benefit that we wouldn't need to remember to add the constant to the context whenever we add a new variable.

app/recordtransfer/context_processors.py
yenaing-oo: There are some constants that I would like to avoid including in the global context, specifically these:
```python
# Define constants for dictionary keys in SubmissionFormWizard
TEMPLATEREF = "templateref"
OTHER_PROVINCE_OR_STATE_VALUE = "Other"
FORMTITLE = "formtitle"
INFOMESSAGE = "infomessage"
FORM = "form"
```

Any suggestions on how to best deal with these?

yenaing-oo: Another solution is to organize the constants in classes, like this:
```python
class Htmlds:
    # User Profile Form
    FIRST_NAME = "id_first_name"
    LAST_NAME = "id_last_name"
    GETS_NOTIFICATION_EMAILS = "id_gets_notification_emails"
    CURRENT_PASSWORD = "id_current_password"
    NEW_PASSWORD = "id_new_password"
    CONFIRM_NEW_PASSWORD = "id_confirm_new_password"

class QueryParams:
   ...

class SubmissionFormWizardKeys:
   TEMPLATEREF = "templateref"
   FORMTITLE = "formtitle"
   INFOMESSAGE = "infomessage"
   FORM = "form"
```

app/recordtransfer/static/recordtransfer/js/submission_form/contactInfo.js
```diff
@@ -8,6 +8,6 @@ export function setupContactInfoForm(context) {
     setupSelectOtherToggle(
         context["id_province_or_state"],
         context["id_other_province_or_state"],
-        context["other_province_or_state_id"]
+        context["other_province_or_state_value"]
```
yenaing-oo, line 11: Unrelated to this issue, but I changed the name of this context variable to accurately reflect what it represents

app/recordtransfer/static/recordtransfer/js/submission_form/contactInfo.js
danloveg, line 1: This uncovered a bug: selecting the Other field in Province or State does not show an other field. This is because of [this line](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/48ab20397cf9a66b9c85622a9a203141ca26f78e/app/recordtransfer/static/recordtransfer/js/submission_form/otherField.js#L31). It's assuming that the select value is a number, but that's not the case here. Can you fix that?

app/recordtransfer/constants.py
```diff
@@ -1,42 +1,72 @@
...
+    INFOMESSAGE: str = "infomessage"
+    FORM: str = "form"
+
+    def asdict(self) -> dict[str, str]:
+        """Return the dataclass as a dictionary."""
+        return dataclasses.asdict(self)
+
+
 # Value to use for the "Other" option in the Province/State dropdown
 OTHER_PROVINCE_OR_STATE_VALUE = "Other"
```
danloveg, line 72: I don't love that everything is neatly organized except this one variable. Could you include this in a dataclass too?

danloveg (changes requested):

app/recordtransfer/constants.py
yenaing-oo: Agreed, I've doen that here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/651/commits/4c423e5910303f89b554fd8f3965b076ec88777c

danloveg (approved):
Looking at the `SubmissionFormWizard`, I think the _TEMPLATES would look cleaner like this:

```python
@dataclasses.dataclass
class SubmissionStepMeta:
    template: str
    title: str
    form: forms.Form

class SubmissionFormWizard(SessionWizardView):
    _TEMPLATES: ClassVar[dict[str, SubmissionStepMeta]] = {
        SubmissionStep.ACCEPT_LEGAL: SubmissionStepMeta(
            template="recordtransfer/submission_form_legal.html",
            title=gettext("Legal Agreement"),
            form=forms.AcceptLegal,
        )
    }

    # ...

    # example usage:

    def get_template_names(self) -> list[str]:
        return [self._TEMPLATES[self.current_step].template]
```

That way we wouldn't need to keep using constant strings, and we would get typing from the type annotations on the data class.

Not necessary for this issue of course, just something I was thinking of.

Issue #639, opened 2025-06-02
Address submission form feedback


danloveg:
This issue will be used to organize the submission form feedback we received recently from an NCTR archivist regarding changes we should make to make the form more user-friendly.

danloveg:
@Eashana0104 @yenaing-oo let me know before either of you work on these issues and I'll set up a new feature branch to work on these.

yenaing-oo:
I think it'll be a good idea to begin working on these issues once #640 is complete and merged into `master`, so as to avoid potential merge conflicts.

Eashana0104:
I was thinking of starting to work on these issues now! @danloveg, whenever you get a chance, would you mind creating a feature branch for this, please?

PR #655, opened 2025-06-02
Refactor Submission.make_bag

danloveg:
What was initially supposed to be a quick removal of an unexpected argument has turned into a refactor of how BagIt bags are generated by the application.

Two glaring issues that this pull request addresses are:

1. Concurrent Bag generation was not supported. If two users wanted to generate a Bag at the same time, at most one of them would generate correctly because they would be generated in the same place at the same time.
2. A Bag does not need to be completely re-generated if it the files have not changed. Logic was added in this pull request to detect whether a Bag should be completely re-generated or if its metadata can be updated in-place.

Changes include:

- Simplify the `make_bag` function in `Submission` by using exceptions instead of complicated return values
- Return `Bag` from `make_bag`
- Add logic to detect whether a Bag needs to be completely re-generated
- Simplify the `create_downloadable_bag` function by removing the nested `try:` blocks
- Fixes issues with unexpected `logger` argument in `make_bag`
- Add tests for `jobs.create_downloadable_bag`
- Remove `BAG_STORAGE_FOLDER` setting
  - Bags are ephemeral now, since they don't need to be kept on the file system.


app/recordtransfer/models.py
```diff
@@ -838,51 +840,28 @@ def make_bag(self, algorithms: Union[str, list] = "sha512", file_perms: str = "6
...
-            LOGGER.error(
-                'The BAG_STORAGE_FOLDER "%s" does not exist!', settings.BAG_STORAGE_FOLDER
-            )
-            return {
-                "missing_files": [],
-                "bag_created": False,
-                "bag_valid": False,
-                "time_created": None,
-            }
+        os.makedirs(settings.BAG_STORAGE_FOLDER, exist_ok=True)
```
yenaing-oo, line 843: Is it safe to assume that `settings.BAG_STORAGE_FOLDER` is set **and** that it is a directory that exists?

`BAG_STORAGE_FOLDER` is already being verified in `app/recordtransfer/management/commands/verify_settings.py`.

 `BAG_STORAGE_FOLDER` also has a default value set in `base.py`, although I do see that a check is being done for this in `Submission.user_folder()` regardless of this.

I assume you also removed this part:
```python
if not os.path.exists(settings.BAG_STORAGE_FOLDER) or not os.path.isdir(
            settings.BAG_STORAGE_FOLDER
        ):
``` 
to allow 
`os.makedirs(settings.BAG_STORAGE_FOLDER, exist_ok=True)` to raise its own errors when creating that directory. I wonder if we can remove `os.makedirs(settings.BAG_STORAGE_FOLDER, exist_ok=True)` altogether and assume it already exists.

app/recordtransfer/models.py
```diff
@@ -838,51 +840,28 @@ def make_bag(self, algorithms: Union[str, list] = "sha512", file_perms: str = "6
...
 
         if not os.path.exists(self.user_folder) or not os.path.isdir(self.user_folder):
```
yenaing-oo, line 845: Can we assume that `self.user_folder` is always a directory? I think the second condition (`not os.path.isdir(self.user_folder)` ) might be redundant.

app/recordtransfer/jobs.py
```diff
@@ -49,55 +49,45 @@ def create_downloadable_bag(submission: Submission, user_triggered: User) -> Non
...
-                or result["time_created"] is None
-            ):
-                new_job.job_status = Job.JobStatus.FAILED
-                new_job.save()
-                return
-
-        try:
+            submission.make_bag(algorithms=settings.BAG_CHECKSUMS)
+
+        if not os.path.exists(settings.TEMP_STORAGE_FOLDER):
```
yenaing-oo, line 54: This made me realise that `TEMP_STORAGE_FOLDER` is not being verified in `verify_storage_folder_settings()` as part of `app/recordtransfer/management/commands/verify_settings.py`. I can open up an issue to verify that as well.

app/recordtransfer/jobs.py
```diff
@@ -49,55 +49,45 @@ def create_downloadable_bag(submission: Submission, user_triggered: User) -> Non
...
+    except Exception as exc:
+        new_job.job_status = Job.JobStatus.FAILED
+        new_job.save()
+        LOGGER.error("Creating zipped bag failed due to exception!", exc_info=exc)
+
     finally:
+        if os.path.exists(submission.location):
+            LOGGER.info("Removing bag from disk after zip generation.")
+            submission.remove_bag()
+
```
yenaing-oo, line 90: Could you remind me why we're removing the bag, even after a successful generation? Is it because we want each bag generation to be fresh for each time the job is called?

yenaing-oo (commented):
I have not taken a close look at the tests yet, but I just had some questions first. Relying on exceptions instead of a dict with "empty" values makes sense to me.

app/recordtransfer/jobs.py
yenaing-oo: Perhaps `make_bag` can store the generated bag in a unique location within `self.location` like this:
```python
new_dir = Path(self.location) / unique_id
new_dir.mkdir(parents=True, exist_ok=False)
...
bag = bagit.make_bag(new_dir, bagit_info, checksums=algorithms)
...
return  new_dir
```
and return the path in which the bag was stored.

That way each job will generate a bag in a unique location, which can be safely deleted once done.

danloveg: I ended up going a bit wild with the refactoring, but there's no problems now with multiple users asking for the same Bag to be generated concurrently.

Each time the Job to create a Bag is run, a new temporary directory is created to generate the Bag in. That way, you can be generating the Bag as many times as necessary in parallel.

https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/af68cfd6e352f1199e6d0c64938d0346ff7d8a36/app/recordtransfer/jobs.py#L60

app/recordtransfer/jobs.py
danloveg: I opened up #704 for this

yenaing-oo (commented):
The code is much more readable now, and the changes you made _does_ address the two concerns described in the PR description. But from what I can tell, the ability to update an existing bag is not utilised anywhere in the codebase currently. Every bag is generated from scratch, even when it's just a submission's metadata that has changed. Is this functionality intended for a future use case?

app/recordtransfer/models.py
```diff
@@ -1061,147 +1028,205 @@ def extent_statements(self):
...
-
-    def make_bag(self, algorithms: Union[str, list] = "sha512", file_perms: str = "644"):
+    def make_bag(
+        self,
+        location: Path,
+        algorithms: Iterable[str] = ("sha512",),
+        file_perms: str = "644",
+    ) -> bagit.Bag:
         """Create a BagIt bag on the file system for this Submission. The location of the BagIt bag
         is determined by self.location. Checks the validity of the Bag post-creation to ensure that
```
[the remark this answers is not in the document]
danloveg: Good catch, I've updated the docstring here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/655/commits/3f49e1c509507b88c755ca9466952480db683d7a

app/recordtransfer/models.py
```diff
@@ -1061,147 +1028,205 @@ def extent_statements(self):
...
 
-        if not algorithms:
-            raise ValueError("algorithms cannot be empty")
+        if not self.bag_name:
+            self.bag_name = self.generate_bag_name()
+            self.save()
 
-        if not self.upload_session:
-            raise ValueError("This submission has no associated upload session")
+        if location.exists() and (location / "data").exists():
```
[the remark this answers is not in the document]
danloveg: This path is not used in the application, you're right, but may be used in the future. I tried to write this function in a way that makes it efficient when given a path to an already existing bag.

yenaing-oo:
> @yenaing-oo I'm planning to remove the `bag_name` field on `Submission`, but that would require a migration. I'm going to wait until #639 is addressed and merged into master because there are migrations as part of that change. Should hopefully be a quick review process after that!

That sounds good to me, and yes, it should be a quick review after that.

yenaing-oo (approved):
Everything looks good to me! Glad the bag creation code is now much more simplified and working as expected.

Issue #642, opened 2025-06-02
Require a SECRET_KEY to be set in `docker_prod` configuration


danloveg:
You are allowed to run the production configuration of the app right now without setting a SECRET_KEY. Since this is a security risk, we should require a SECRET_KEY to be set when using the `docker_prod` configuration.

yenaing-oo:
Right now, `SECRET_KEY` is being set in `base.py` for both the `docker_dev` and `docker_prod` configuration. To make it required for production, should we set a default value for `SECRET_KEY` only for `docker_dev` and not for `docker_prod` so that `decople` raises an error if that it is not found in the environment?

```python
# docker_dev.py
SECRET_KEY = config("SECRET_KEY", default="default_secret_key")

# docker_prod.py
SECRET_KEY = config("SECRET_KEY")
```

PR #660, opened 2025-06-03
Make SECRET_KEY required for production

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/642

### DIFF
--- a/app/app/settings/base.py
+++ b/app/app/settings/base.py
@@ -7,8 +7,6 @@ BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__fil
 
 TESTING = False
 
-SECRET_KEY = config("SECRET_KEY", default="q9n%k!e3k8vuoo9vnromslji*hsczyj84krzz1$g=i$wp2r!s-")
-
 FORM_RENDERER = "django.forms.renderers.TemplatesSetting"
 
 INSTALLED_APPS = [
--- a/app/app/settings/docker_dev.py
+++ b/app/app/settings/docker_dev.py
@@ -14,6 +14,8 @@ ALLOWED_HOSTS = [
     "localhost",
 ]
 
+SECRET_KEY = config("SECRET_KEY", default="q9n%k!e3k8vuoo9vnromslji*hsczyj84krzz1$g=i$wp2r!s-")
+
 # MySQL Database
 
 DEV_DATABASE_NAME = config("DEV_DATABASE_NAME", default="development_database.sqlite3")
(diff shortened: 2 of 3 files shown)


danloveg:
> EDIT: Copilot is recommending I use "Github Secrets" instead of setting a variable, even though it's just for testing. To follow good practice, could you set it as a secret instead of a variable?

I've added the secret but I re-ran the workflow and it still can't find it. Any idea why that might be? Maybe you need to push a new commit, it could be that there's some cached state right now and it won't be able to find the new secret until some change is made.

Screenshot 2025-06-06 085409

danloveg:
> > > EDIT: Copilot is recommending I use "Github Secrets" instead of setting a variable, even though it's just for testing. To follow good practice, could you set it as a secret instead of a variable?
> > 
> > 
> > I've added the secret but I re-ran the workflow and it still can't find it. Any idea why that might be? Maybe you need to push a new commit, it could be that there's some cached state right now and it won't be able to find the new secret until some change is made.
> 
> I think the SECRET_KEY that we're setting in the workflow is not available to the Django app, since it's running within a container with a separate environment. The only solution I can think of right now is to create a `.prod.env` containing `SECRET_KEY` before running `docker compose` in the workflow.

That could work, then how about this:

```yaml
      - name: Copy environment file
        run: cp example.prod.env .prod.env

      - name: Add SECRET_KEY to environment file
        run: echo "SECRET_KEY=${{ secrets.SECRET_KEY }}" >> .prod.env
```

yenaing-oo:
> > > > EDIT: Copilot is recommending I use "Github Secrets" instead of setting a variable, even though it's just for testing. To follow good practice, could you set it as a secret instead of a variable?
> > > 
> > > 
> > > I've added the secret but I re-ran the workflow and it still can't find it. Any idea why that might be? Maybe you need to push a new commit, it could be that there's some cached state right now and it won't be able to find the new secret until some change is made.
> > 
> > 
> > I think the SECRET_KEY that we're setting in the workflow is not available to the Django app, since it's running within a container with a separate environment. The only solution I can think of right now is to create a `.prod.env` containing `SECRET_KEY` before running `docker compose` in the workflow.
> 
> That could work, then how about this:
> 
> ```yaml
>       - name: Copy environment file
>         run: cp example.prod.env .prod.env
> 
>       - name: Add SECRET_KEY to environment file
>         run: echo "SECRET_KEY=${{ secrets.SECRET_KEY }}" >> .prod.env
> ```

Nice, this worked! Just had to add a new line before adding the secret key environment variable.

danloveg (changes requested):

danloveg (approved):

Issue #647, opened 2025-06-02
Re-style about page with DaisyUI


danloveg:
The about page looks awkward now that it's missing list styles and a bunch of spacing. It should be restyled with DaisyUI.

yenaing-oo:
I'll be experimenting with the `[tailwindcss-typography](https://github.com/tailwindlabs/tailwindcss-typography)` plugin to how well it can style whole sections of text without having to explicitly define classes for each text HTML tag.

Eashana0104:
@yenaing-oo  I ll check this out too.

yenaing-oo:
I've discovered the reason why the list styles and spacing were missing is because Tailwind by default strips away these default browser styles: https://tailwindcss.com/docs/preflight

It seems that the [Tailwind Typography ](https://github.com/tailwindlabs/tailwindcss-typography) plugin adds its own styles back, with the added ability to style a whole section using `prose` styles such as:

```html
<article class="prose prose-img:rounded-xl prose-headings:underline prose-a:text-blue-600">
  {{ markdown }}
</article>
```
In the example, any images in this article will be rounded, headings underlines, etc.

For now, I don't feel that we need the ability to style whole sections this way, so I'm choosing not to add this as a dependency. We can continue to style individual HTML elements using existing Tailwind and DaisyUI classes.

Eashana0104:
Yeah, i feel with daisyUI and Tailwind alone, it has made everything so much easier. I didn't look at this plugin yet, but if I will for sure, It sounds useful for some situations but you are right that i dont think we necessarily need it for now.

PR #675, opened 2025-06-06
Restyle About page with daisyUI

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/647


danloveg (approved):
I like this. I want the same title + main intro text on the help page, I think that would look good - I'll ask Ash about that.

PR #678, opened 2025-06-06
609 restyle login and signup page using daisyui

Eashana0104:


app/recordtransfer/views/login_view.py
```diff
@@ -0,0 +1,35 @@
+import logging
+
+from django.contrib.auth.forms import AuthenticationForm
+from django.contrib.auth.views import LoginView
+from django.http import HttpResponse
+from django.shortcuts import redirect
+from django.template.loader import render_to_string
+
+logger = logging.getLogger(__name__)
```
danloveg, line 9: I don't believe this is necessary, I'm not seeing any usage of this logger

app/recordtransfer/views/login_view.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class HTMXLoginView(LoginView):
```
danloveg, line 12: I think this class would fit better in `account.py`. Can you also rename it `Login`? We have other views that use HTMX responses, but we're not adding HTMX to the class name

app/recordtransfer/views/login_view.py
```diff
@@ -0,0 +1,35 @@
...
+    def form_invalid(self, form: AuthenticationForm) -> HttpResponse:
+        """Handle invalid login form submissions."""
+        if self.request.headers.get("HX-Request"):
+            html = render_to_string(
+                "registration/login_errors.html",  # Changed to form-only template
+                {"form": form},
+                request=self.request,
+            )
+            return HttpResponse(html)
+        return super().form_invalid(form)
```
danloveg, line 35: Can you add some unit tests for this class in `recordtransfer.tests.unit.views.test_account`?

There should be a new class like:

```python
class TestLogin(TestCase):
    """Tests for the Login view."""

    # ... test here ...
```

app/recordtransfer/views/account.py
```diff
@@ -38,8 +39,25 @@ def form_valid(self, form: BaseModelForm) -> HttpResponse:
...
+        """Handle invalid signup form submissions."""
+        if self.request.headers.get("HX-Request"):
+            # Return only the error template for HTMX requests
+            html = render_to_string(
+                "recordtransfer/signup_form.html",
+                {"form": form},
+                request=self.request,
+            )
+            return HttpResponse(html)
+        return super().form_invalid(form)
```
danloveg, line 59: Can you add more tests to the `TestCreateAccount` unit test class that checks this logic?

app/recordtransfer/templates/registration/login_errors.html
```diff
@@ -0,0 +1,13 @@
...
+        {% for error in form.non_field_errors %}<div class="bg-red-50 border border-red-200 text-red-700 px-1 py-1 rounded-md mb-2">{{ error }}</div>{% endfor %}
+    {% endif %}
+    {% for field in form %}
+        {% if field.errors %}
+            {% for error in field.errors %}
+                <div class="bg-red-50 border border-red-200 text-red-700 px-1 py-1 rounded-md mb-2">{{ field.label }}: {{ error }}</div>
+            {% endfor %}
+        {% endif %}
+    {% endfor %}
+</div>
```
danloveg, line 13: Can you add an animation to this error message? I found if I clicked **Login** multiple times with errors each time, there was no feedback to indicate that the form was submitted.

danloveg (changes requested):
Looks like you're already a Javascript pro, so no comments there!

The functionality is all working great, and the styling is very nice. I have a few suggestions for cleaning the code up, and increasing test coverage. I'm also thinking some sort of animation on the error message on the login page would be nice since the error message doesn't change if the form is submitted multiple times in a row with errors.

app/recordtransfer/views/login_view.py
Eashana0104: I thought so too that i should probably move Login class in account.py but stopped as I thought maybe I should be mindful about Single Responsibility principle. In account.py, classes were related to account creation. In login i thought it was more of authentication. Now i think about it, I think i overthinked it.

danloveg: > I thought so too that i should probably move Login class in account.py but stopped as I thought maybe I should be mindful about Single Responsibility principle. In account.py, classes were related to account creation. In login i thought it was more of authentication. Now i think about it, I think i overthinked it.

All good, that's a good thought. Sometimes we need to balance SRP with simplicity, in this case it's simpler to have the login account grouped in with the account creation stuff

app/recordtransfer/templates/registration/login_errors.html
Eashana0104: I added a custom animation to all the error fields, this way when the user clicks on submit multiple times, the error would get animated. The animation was custom made as well as daisy ui doesnt have a professional looking animation . Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/678/commits/c0c68041e633f5d56168d48848d08d9e64f1052f

app/recordtransfer/views/account.py
Eashana0104: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/678/commits/9e7829a66039b8f5f4be5055309bc50c27c108b4. I added 3 different tests in order to test when the form is valid, when the form is invalid and when the form is empty, which is similar to form bein ginvalid.

app/recordtransfer/tests/unit/views/test_account.py
```diff
@@ -230,6 +231,88 @@ def test_user_defaults_set_correctly(self) -> None:
...
+        }
+
+        for field, error_message in required_fields.items():
+            # Verify field exists in the response
+            self.assertIn(f"id_{field}", str(response.content))
+            # Verify the specific error message appears for this field
+            self.assertIn(error_message, str(response.content))
+
+        # User should not be created
+        self.assertEqual(User.objects.count(), 1)
```
danloveg, line 314: Instead of hard-coding the `1`, can you do this instead?

- Before making the request, store `user_count_before_request = User.objects.count()`

```suggestion
        # User should not be created
        self.assertEqual(User.objects.count(), user_count_before_request)
```

danloveg (changes requested):
I'm impressed by how clean these new tests are, good job.

I had one tiny comment about checking that the user object count doesn't change, but other than that, looks good!

Eashana0104:
Yaya! Thanks, I spent some time on making sure the tests are clean and consistent. I have committed the sugegstion and I think this should be good to be merged into the feature branch

Issue #559, opened 2025-05-09
Allow Overriding Settings in Database


danloveg:
It may be useful to be able to set certain settings in the database that can override settings set in the environment variable file.

For example, say some administrator of the website wants to change the ARCHIVIST_EMAIL setting. To do that right now, the whole application would have to be re-deployed to change the `.prod.env` file.

What I'm proposing is that the administrator could log in to the admin site, and change the email directly from the admin website.

There are a number of changes that this *doesn't* make sense to implement for. For example, we don't want someone to be able to change the `REDIS_HOST` setting because that would break the app. But there are certain settings where changing them would be useful and also would not break the website. More good candidates to add to the database would be the CAAIS defaults - those settings feel like something that should be allowed to be controlled by an archivist.

I asked Gemini what this might look like and it recommended this approach:

Add a new `SiteSetting` model:

```python
# models.py

from django.db import models

class SiteSetting(models.Model):
    key = models.CharField(max_length=255, unique=True)
    value = models.TextField()

    def __str__(self):
        return f"{key}={value}"
```

And load the key=value pairs in `settings.py`:

```python
# settings.py

import os
import django
from django.core.exceptions import ImproperlyConfigured

SITE_NAME = "My Site"

try:
    from django.apps import app
    if apps.ready:
        from models import SiteSetting

        # Load all settings
        site_settings = {setting.key: setting.value for setting in SiteSetting.objects.all()}

        # Override settings if they're set in the database
        SITE_NAME = site_settings.get("SITE_NAME", SITE_NAME)

except (ImproperlyConfigured, ImportError):
    # Handle the case when Django isn't ready
    pass

**Comment by yenaing-oo** [comment]
From reading online and experimenting myself, it appears that you cannot load models within `settings.py` because the settings are read first before you can use the ORM. I get this error:

```python
# settings.py
from recordtransfer.models import User
```

```log
...
2025-05-29 16:25:30   File "/opt/secure-record-transfer/app/app/settings/base.py", line 5, in <module>
2025-05-29 16:25:30     from recordtransfer.models import User
2025-05-29 16:25:30   File "/opt/secure-record-transfer/app/recordtransfer/models.py", line 13, in <module>
2025-05-29 16:25:30     from caais.models import Metadata
2025-05-29 16:25:30   File "/opt/secure-record-transfer/app/caais/models.py", line 32, in <module>
2025-05-29 16:25:30     from caais.managers import (
[...]

PR #683, opened 2025-06-09
Allow Overriding Settings in Database

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/559

Change notes:
- Migrated the following environment variable settings to the database under a `SiteSetting` model:

1. DO_NOT_REPLY_USERNAME
2. ARCHIVIST_EMAIL
3. PAGINATE_BY
4. CAAIS_DEFAULT_REPOSITORY
5. CAAIS_DEFAULT_ACCESSION_TITLE
6. CAAIS_DEFAULT_ARCHIVAL_UNIT
7. CAAIS_DEFAULT_DISPOSITION_AUTHORITY
8. CAAIS_DEFAULT_ACQUISITION_METHOD
9. CAAIS_DEFAULT_STATUS
10. CAAIS_DEFAULT_SOURCE_CONFIDENTIALITY
11. CAAIS_DEFAULT_PRELIMINARY_CUSTODIAL_HISTORY
12. CAAIS_DEFAULT_DATE_OF_MATERIALS
13. CAAIS_DEFAULT_EXTENT_TYPE
14. CAAIS_DEFAULT_QUANTITY_AND_UNIT_OF_MEASURE
15. CAAIS_DEFAULT_CONTENT_TYPE
16. CAAIS_DEFAULT_CARRIER_TYPE
17. CAAIS_DEFAULT_EXTENT_NOTE
18. CAAIS_DEFAULT_PRELIMINARY_SCOPE_AND_CONTENT
19. CAAIS_DEFAULT_LANGUAGE_OF_MATERIAL
20. CAAIS_DEFAULT_STORAGE_LOCATION
21. CAAIS_DEFAULT_PRESERVATION_REQUIREMENTS_TYPE
22. CAAIS_DEFAULT_PRESERVATION_REQUIREMENTS_VALUE
23. CAAIS_DEFAULT_PRESERVATION_REQUIREMENTS_NOTE
24. CAAIS_DEFAULT_APPRAISAL_TYPE
25. CAAIS_DEFAULT_APPRAISAL_VALUE
26. CAAIS_DEFAULT_APPRAISAL_NOTE
27. CAAIS_DEFAULT_ASSOCIATED_DOCUMENTATION_TYPE
28. CAAIS_DEFAULT_ASSOCIATED_DOCUMENTATION_TITLE
29. CAAIS_DEFAULT_ASSOCIATED_DOCUMENTATION_NOTE
30. CAAIS_DEFAULT_GENERAL_NOTE
31. CAAIS_DEFAULT_RULES_OR_CONVENTIONS
32. CAAIS_DEFAULT_LANGUAGE_OF_ACCESSION_RECORD
33. CAAIS_DEFAULT_SUBMISSION_EVENT_TYPE
34. CAAIS_DEFAULT_SUBMISSION_EVENT_AGENT
35. CAAIS_DEFAULT_SUBMISSION_EVENT_NOTE
36. CAAIS_DEFAULT_CREATION_TYPE
37. CAAIS_DEFAULT_CREATION_AGENT
38. CAAIS_DEFAULT_CREATION_NOTE

- Set up an admin for `SiteSetting` with validation to enable admins to change setting values
- Added admin docs for modifying setting values
- Added `recordtransfer.enums` to docs under Developer Resources
- Added tests for `SiteSetting` and `SiteSettingModelForm`


yenaing-oo:
Todo:
`CAAIS_UNKNOWN_START_DATE` and `CAAIS_UNKNOWN_END_DATE` should be stored with a `value_type` of a new type `SiteSetting.SettingType.DATE`. This way, dates can be retrieved in a Python date type using a new method `get_date_value()`. **Dependent on (1)**

Questions:
1) How should we deal with `CAAIS_UNKNOWN_*` settings being used in the`caais` app? The `SiteSetting` model is currently part of the `recordtransfer` app. Since we want the `caais` and `recordtransfer` app to be separate, we cannot use the `SiteSetting` model to fetch `CAAIS_UNKNOWN_*` setting values from the database. Should these remain as an environment variable?

2) How detailed should the description for each setting be? I'm concerned with `recordtransfer.models` being bloated with strings, although having the setting names and descriptions in one place is good for maintainability: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/660782f5d3d7be944c8049f01e29b3d0076b49d5/app/recordtransfer/models.py#L228-L377

yenaing-oo:
> I've just spent a bunch of time thinking about this. I think we need a slightly different data structure. How about this?
> 
> ```python
> from typing import NamedTuple
> 
> class SettingKeyMeta(NamedTuple):
>     """Describes a setting's type, description, and default value."""
> 
>     value_type_name: str
>     description: str
>     default_value: Optional[str] = None
> 
> class SiteSetting(models.Model):
>     
>     class Key(Enum):
>         ARCHIVIST_EMAIL = SettingKeyMeta(
>             "str",
>             _("The email displayed for people to contact an archivist."),
>             default_value="archivist@example.com",
>         )
> 
>         # (example with no default)
>         CAAIS_DEFAULT_REPOSITORY = SettingKeyMeta(
>             "str",
>             _("Default value to fill in metadata for CAAIS sec. 1.1 - Repository."),
>         )
> 
>         @property
>         def key_name(self) -> str:
>             return self.name
> 
>         @property
>         def value_type_name(self) -> str:
>             return self.value.value_type_name
> 
>         @property
>         def description(self) -> str:
>             return self.value.description
> 
>         @property
>         def default_value(self) -> Optional[str]:
>             return self.value.default_value
> ```

This looks good to me, although I have one concern. Are we okay with using raw strings for the `value_type_name`? This means we would also have to check the type like this:
```python
def get_value(self):
   if setting.value_type_name == "str":
       # casting done here
```
Perhaps we can keep the existing `SiteSetting.SettingType` enum for this?



> This also enables you to do this:
> 
> ```python
> # file: 0050_sitesetting.py
> 
> def populate_site_settings(apps, schema_editor):
>     """Populate SiteSetting model with initial values."""
>     SiteSetting = apps.get_model("recordtransfer", "SiteSetting")
> 
>     settings_to_create = []
>     for setting in SiteSetting.Key:
>         settings_to_create.append(
>             SiteSetting(
>                 key=setting.key_name,
>                 value=setting.default_value or "",
>                 value_type=setting.value_type_name,
>             )
> 
>     SiteSetting.objects.bulk_create(settings_to_create)
> ```

I believe accessing `SiteSetting.Key` like this in a migration is not allowed because enum classes under the model class are not tracked in migrations. There is no guarantee that the enum class would still be there when the migration gets run since the code could have changed.

> I _would_ say we could add the description to the database, but that's not ideal for translation.

This makes sense to me.

danloveg:
> This looks good to me, although I have one concern. Are we okay with using raw strings for the `value_type_name`? This means we would also have to check the type like this:
> 
> ```python
> def get_value(self):
>    if self.value_type == "str":
>        # casting done here
> ```
> 
> Perhaps we can keep the existing `SiteSetting.SettingType` enum for this?

My bad, I didn't see you already created an Enum for the variable type. Yes, keep that enum because that works better.

> Also, just to confirm, is it okay to include detailed paragraph descriptions of a setting within the enum class, or should we try to keep this as concise as possible? I'm assuming we would still want to document these settings somewhere in the docs, except no longer in `settings/index.rst`, in which case short descriptions should be fine.

Yes, I think it makes the most sense to keep these descriptions in the enum, the code is self-documenting in that way.

We'll need to document those in the docs too, yes, but that's no more tedious than the current method of manually documenting all the settings.

> > This also enables you to do this:
> > ```python
> > # file: 0050_sitesetting.py
> > 
> > def populate_site_settings(apps, schema_editor):
> >     """Populate SiteSetting model with initial values."""
> >     SiteSetting = apps.get_model("recordtransfer", "SiteSetting")
> > 
> >     settings_to_create = []
> >     for setting in SiteSetting.Key:
> >         settings_to_create.append(
> >             SiteSetting(
> >                 key=setting.key_name,
> >                 value=setting.default_value or "",
> >                 value_type=setting.value_type_name,
> >             )
> > 
> >     SiteSetting.objects.bulk_create(settings_to_create)
> > ```
> 
> I believe accessing `SiteSetting.Key` like this in a migration is not allowed because enum classes under the model class are not tracked in migrations. There is no guarantee that the enum class would still be there when the migration gets run since the code could have changed. We'll have to manually list out every item of data we want to add.

Hmm.. I wonder if there's a better way to do that than manually listing them out. Would moving the enum out of the class work?

yenaing-oo:
@danloveg 
I changed the placement of the classes `SettingType`, `Key` and `SettingKeyMeta` by moving them (and renaming some of them) to `enums.py`. I did this because of forward reference issues due to having both `SettingType` and `Key` defined in `SiteSetting`, where `SiteSetting.SettingType` was being referenced in `Key` without `SiteSetting` being fully defined yet.

In my changes, I placed `SettingKeyMeta` under the `SiteSettingKey` class since it was only going to be used there.

Let me know what you think about the way I've organised the code. 

> Hmm.. I wonder if there's a better way to do that than manually listing them out. Would moving the enum out of the class work?

As for avoiding listing out the initial values for `SiteSetting` manually in the migration, I don't think there is any way around this. I got this error from importing `SiteSettingKey` from `recordtransfer.enums` in the migration:

```log
...
  File "/opt/secure-record-transfer/.venv/lib/python3.10/site-packages/django/db/backends/base/base.py", line 531, in validate_no_broken_transaction
    raise TransactionManagementError(
django.db.transaction.TransactionManagementError: An error occurred in the current transaction. You can't execute queries until the end of the 'atomic' block.
```

Even if moving the enum out of the model class and importing it in the migration does work, we still have to be concerned that the application code might change. I found multiple posts on Stackoverflow that warn against doing this:
- https://stackoverflow.com/a/59142840/15193287
- https://stackoverflow.com/a/78950687/15193287

app/recordtransfer/migrations/0050_sitesetting.py
danloveg, line 1: Since I merged #693 in, the `dependencies` for this migration will need to change

yenaing-oo:
@danloveg 
I will add screenshots to `docs/admin_guide/site_settings.rst` once everything is finalized, in case any changes to the admin interface for `SiteSetting` is needed.

yenaing-oo:
I've realised that since not all settings have a default value, `get_value_str` and `get_value_int` could also return `None`. This does make caching a bit inefficient, since the cache has no way of knowing whether a value is actually `None` or we just haven't made our first fetch to cache the value yet. This means `set_cache` will do a lookup everytime a value is `None`.

See commit: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/683/commits/963b553435cebf59b96421d2ab3c93bba8db2fe1

yenaing-oo:
I'll add screenshots once we're happy with the UI for resetting a SiteSetting.

danloveg:
> I've realised that since not all settings have a default value, `get_value_str` and `get_value_int` could also return `None`. This does make caching a bit inefficient, since the cache has no way of knowing whether a value is actually `None` or we just haven't made our first fetch to cache the value yet. This means `set_cache` will do a lookup everytime a value is `None`.
> 
> See commit: [963b553](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/commit/963b553435cebf59b96421d2ab3c93bba8db2fe1)

That's OK for now. After I merge this in, can you open up a new issue to look into that? We can take another look at that in July.

yenaing-oo:
> > I've realised that since not all settings have a default value, `get_value_str` and `get_value_int` could also return `None`. This does make caching a bit inefficient, since the cache has no way of knowing whether a value is actually `None` or we just haven't made our first fetch to cache the value yet. This means `set_cache` will do a lookup everytime a value is `None`.
> > See commit: [963b553](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/commit/963b553435cebf59b96421d2ab3c93bba8db2fe1)
> 
> That's OK for now. After I merge this in, can you open up a new issue to look into that? We can take another look at that in July.

Sounds good, I'll do that.

yenaing-oo:
Took me a while to figure out how to set the `changed_by` field on an instance of `SiteSetting` for an update to it done through `SiteSettingAdmin`. I was trying to pass `request.user` to the form instance through `SiteSettingAdmin`, but it doesn't support this functionality since a `ModelAdmin` is different from an `UpdateView`.

See my initial janky solution: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/683/commits/3ef58484c1849f7b5d9a6defe36bbd42ae482927
After digging through Django's docs for `ModelAdmin`: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/683/commits/33861d88b1101fb185c743b00f9b3653c8e43ae9

danloveg (approved):
This is really great work. One of our more complicated changes, that's for sure! I'm happy with how clean this turned out.

PR #686, opened 2025-06-09
Restyle heading and buttons for password reset page

Eashana0104:


Eashana0104:
Yeah i see what you mean with this. Since the border for the email field was set usind daisy ui input, i created a new css rule for the rounded borders which could be used at various places, its resolved in https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/686/commits/1dc804f9f04539fe4c50f78494a1ca5970da895e

PR #693, opened 2025-06-12
Make viewing individual files from the admin interface easier

### LINKED ISSUE #624: Make viewing individual files from the backend easier
Right now, permanent uploaded files cannot be downloaded or viewed from the admin backend. There is a use case for looking at individual files, so we should look at ways to make it easier to see these files without having to create a job to create a bag just to view one file.

I don't want to make temp files viewable in this way, though. What "temporary" files are is confusing so I want to abstract some of the temp file admin stuff away.

I have some suggestions to make this happen:

1. Enable viewing files directly from the permanent file admin page. To do this, we can remove the `admin_uploadedfile.bundle.js`.
2. Enable viewing files inline in the `PermUploadedFileInline` admin (do not enable this same functionality for the `TempUploadedFileInline`.
3. Remove the admin page for `TempUploadedFile`. I don't think this page is necessary, if someone *really* wants to see in-transit temp files they can access those from the `UploadSessionAdmin`.
4. Use [get_inlines](https://docs.djangoproject.com/en/4.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.get_inlines) to only show the temporary uploaded file inline when the upload session is in the CREATED, UPLOADING, EXPIRED, or COPYING_IN_PROGRESS state.
5. Change the "TEMPORARY UPLOADED FILES" title in the inline admin to "FILES CURRENTLY BEING UPLOADED" (see picture below)

Image

Eashana0104:
Closes #624 

Change notes:
- Removed [`@validate_upload_access`](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/619d49a142d84ad2223a6f0314706c8b8c0bc551/app/recordtransfer/decorators.py#L12) decorator from `views.media.uploaded_file` to allow access to uploaded files from outside the Submission Form
- Refactored and renamed `recordtransfer.models.UploadSession.get_temp_file_by_name` to get both permanent and temp files


app/recordtransfer/admin.py
```diff
@@ -222,15 +226,23 @@ class UploadSessionAdmin(ReadOnlyAdmin):
...
     ordering = [
         "-started_at",
     ]
 
+    def get_inlines(self, request, obj=None) -> list:
+        """Return the inlines to display for the UploadSession."""
+        if obj and obj.status in {
+            UploadSession.SessionStatus.CREATED,
+            UploadSession.SessionStatus.UPLOADING,
+            UploadSession.SessionStatus.COPYING_IN_PROGRESS,
```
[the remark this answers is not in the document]
yenaing-oo: Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/693/commits/75af558c32c882c9970ef6b1bb9b791eb21a1ee1

app/recordtransfer/admin.py
```diff
@@ -77,6 +78,12 @@ def format_upload_size(obj: BaseUploadedFile) -> str:
     return get_human_readable_size(int(obj.file_upload.size), 1000, 2)
```
danloveg, line 78: I kept getting an error because one of the files I had uploaded didn't exist.

Screenshot 2025-06-16 133217

Screenshot 2025-06-16 133234

Can you change this to check that the file exists first?

```suggestion
    if not obj.file_upload or not obj.exists:
        return "N/A"
    return get_human_readable_size(int(obj.file_upload.size), 1000, 2)
```

app/recordtransfer/admin.py
```diff
@@ -77,6 +78,12 @@ def format_upload_size(obj: BaseUploadedFile) -> str:
...
 
 
+@display(description=gettext("File Link"))
+def media_url(obj: BaseUploadedFile) -> SafeText:
+    """Return the media URL for a BaseUploadedFile instance."""
+    return format_html('<a href="{}">{}</a>', obj.get_file_access_url(), obj.name)
```
danloveg, line 84: Similar to my above comment, in case the file doesn't exist, can we do this?

I also think it'd be better if the file opened in a new tab.

```suggestion
    if not obj.file_upload or not obj.exists:
        return mark_safe(gettext("File was removed"))
    return format_html('<a target="_blank" href="{}">{}</a>', obj.get_file_access_url(), obj.name)
```

danloveg (changes requested):
All the changes I mentioned in #624 have been addressed here, so that's great.

There is no handling in the admin for when a file was removed, so I'm requesting two changes regarding that.

app/recordtransfer/admin.py
yenaing-oo: Just tested it, works as expected:
- Inline:
image
- Admin:
image

danloveg (approved):

Issue #697, opened 2025-06-13
Register Password Reset to `send_test_email` command


danloveg:
The Password Reset email should be registered to the `send_test_email` function so that this email can be sent by that command.

yenaing-oo:
Digging into the code for resetting a password, I've discovered that the built-in Django view for password resets, [PasswordResetView](https://docs.djangoproject.com/en/4.2/topics/auth/default/), sends an email synchronously. See:

- https://github.com/django/django/blob/8e2249bc79a74d96f9ad20c89f82a89c78d4b648/django/contrib/auth/views.py#L218
- https://github.com/django/django/blob/8e2249bc79a74d96f9ad20c89f82a89c78d4b648/django/contrib/auth/forms.py#L391

I'll be overriding this behaviour to send this email asynchronously using `django-rq` in the while addressing the original issue.

PR #699, opened 2025-06-13
Send password reset email asynchronously

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/697

### DIFF
--- a/app/app/urls.py
+++ b/app/app/urls.py
@@ -17,26 +17,22 @@ Including another URLconf
 
 from django.contrib import admin
 from django.contrib.auth import views as auth_views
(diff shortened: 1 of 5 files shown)


yenaing-oo:
@danloveg 
I see that the subjects of all the `emails.py` aren't being translated. I know Django has a way to set this for their built-in email templates, such as password resets, using text files, such as: https://github.com/django/django/blob/main/django/contrib/auth/templates/registration/password_reset_subject.txt

I think this is a bit excessive for each email subject. Perhaps we can call translation in code instead?

danloveg (approved):

Issue #622, opened 2025-05-29
Make groups form more user friendly


danloveg:
To address feedback to make the submission group part of the submission form, we should look to add context to that page as to *why* a user may want to create a group or assign a submission to a group.

Eashana0104:
Closed by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/725

PR #725, opened 2025-06-19
622-Address submission group feedback

Eashana0104:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/622

### DIFF
--- a/app/recordtransfer/views/pre_submission.py
+++ b/app/recordtransfer/views/pre_submission.py
@@ -139,7 +139,7 @@ class SubmissionFormWizard(SessionWizardView):
             title=gettext("Assign Submission to Group (Optional)"),
             form=forms.GroupSubmissionForm,
             info_message=gettext(
-                "If you plan to submit multiple files that together exceed 40 files or 256 MiB, we recommend organizing them into a submission group. "
+                "If you plan to submit multiple files that together exceed 40 files or 256 MiB, we recommend organizing them into a submission group. "
                 "Submission groups allow you to keep related submissions together. To do this, Select an existing group from the dropdown below or Create a new one. "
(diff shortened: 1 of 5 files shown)


app/recordtransfer/views/pre_submission.py
```diff
@@ -139,8 +139,9 @@ class SubmissionStepMeta:
             title=gettext("Assign Submission to Group (Optional)"),
             form=forms.GroupSubmissionForm,
             info_message=gettext(
-                "If this submission belongs in a group with other submissions you have made or will "
-                "make, select the group it belongs in in the dropdown below, or create a new group"
+                "If you plan to submit multiple files that together exceed 40 files or 256 MiB, we recommend organizing them into a submission group. "
+                "Submission groups allow you to keep related submissions together. To do this, Select an existing group from the dropdown below or Create a new one. "
+                "When creating a new group, consider using a clear and consistent name format, such as YourLastName_YYYY-MM-DD, so you can easily identify and assign related submissions to the same group in the future."
```
danloveg, line 144: The "40 files" and "256 MiB" should not be hard-coded because these are configurable parameters. See [here](https://secure-record-transfer.readthedocs.io/en/latest/settings/index.html#max-single-upload-size-mb).

I also think this message is too long for the form - people aren't going to read it. How about making a new section on the help page for this, and add a link to the template that says "Why would I want to make a group?" that takes them to that section of the help page.

Then we can add some examples for why you would want to create a group:

- The total size of your donation exceeds our limits
- You are planning to send more related files at a later date

danloveg (changes requested):

Eashana0104:
> This is really nice! I think the situations you've outlined in the FAQ are perfect, and also having the reason under each bullet point **bolded** makes it easy to read.

Thanks! I have committed all the suggestions; they make so much sense! This PR should be good now

Issue #620, opened 2025-05-29
Make rights form more user friendly


danloveg:
We've received some feedback to make this form more user friendly. To address this feedback:

1. Rename this step to "Record Rights and Restrictions"
2. Add more text to the top of the page so the user knows what this page is for. Something along the lines of: "Depending on the records you are transferring, there may be specific rights or restrictions that govern access to your records. Enter any associated rights or restrictions that apply to the records. Add as many sections as you like using the + More button. You may enter another type of rights or restriction if the dropdown does not contain the type you're looking for."
  - This information comes from the help page, maybe there's a better way to have that information in two places.
3. Make this form optional. The rights section is optional in CAAIS 1.0, so a person does not *need* to enter any.

PR #727, opened 2025-06-19
620 make rights form more user friendly

Eashana0104:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/620

### DIFF
--- a/app/recordtransfer/constants.py
+++ b/app/recordtransfer/constants.py
@@ -63,16 +63,3 @@ class OtherValues:
     def asdict(self) -> dict[str, str]:
         """Return the dataclass as a dictionary."""
         return dataclasses.asdict(self)
-
-
-@dataclasses.dataclass(frozen=True)
-class HelpMessages:
-    """Class to hold help messages used throughout the application."""
-
-    # Rights form help message
-    RIGHTS_INFO: str = (
-        " Depending on the records you are submitting, there may be specific rights that govern "
-        "the access of your records. The following is a brief description about the types of "
-        " rights that are available to be used. If none suit your needs, select Other in"
-        "the rights dropdown."
-    )
--- a/app/recordtransfer/constants.py
+++ b/app/recordtransfer/constants.py
@@ -63,3 +63,16 @@ class OtherValues:
     def asdict(self) -> dict[str, str]:
         """Return the dataclass as a dictionary."""
         return dataclasses.asdict(self)
+
+
+@dataclasses.dataclass(frozen=True)
+class HelpMessages:
+    """Class to hold help messages used throughout the application."""
+
+    # Rights form help message
+    RIGHTS_INFO: str = (
+        " Depending on the records you are submitting, there may be specific rights that govern "
+        "the access of your records. The following is a brief description about the types of "
+        " rights that are available to be used. If none suit your needs, select Other in"
+        "the rights dropdown."
+    )
--- a/app/recordtransfer/forms/submission_forms.py
+++ b/app/recordtransfer/forms/submission_forms.py
@@ -698,7 +698,7 @@ class RightsFormSet(BaseFormSet):
 
     def __init__(self, *args, **kwargs):
         super(RightsFormSet, self).__init__(*args, **kwargs)
-        self.forms[0].empty_permitted = True  # Changed to True to make it optional
+        self.forms[0].empty_permitted = True
 
 
 class RightsForm(SubmissionForm):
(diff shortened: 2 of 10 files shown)


app/recordtransfer/templates/recordtransfer/help.html
```diff
@@ -92,10 +92,7 @@
                 <span class="inline-block text-primary text-md font-semibold rounded px-2 py-1 mb-2">Step 5 of Submission Form</span>
                 <p class="mb-4 prose">
                     {% blocktrans %}
-                Depending on the records you are submitting, there may be specific rights that govern
-                the access of your records. The following is a brief description about the types of
-                rights that are available to be used. If none suit your needs, select <b>Other</b> in
-                the rights dropdown.
+                        {{ help_messages }}
```
danloveg, line 95: I like how you're thinking here, but unfortunately for translation it won't work. Can you change it back so that the help message is directly in the template?

For example, if you ran `makemessages` to extract all translatable strings the way you have it right now, the string in `{{ help_message }}` won't be translatable.

app/recordtransfer/constants.py
```diff
@@ -63,3 +63,16 @@ class OtherValues:
...
+class HelpMessages:
+    """Class to hold help messages used throughout the application."""
+
+    # Rights form help message
+    RIGHTS_INFO: str = (
+        " Depending on the records you are submitting, there may be specific rights that govern "
+        "the access of your records. The following is a brief description about the types of "
+        " rights that are available to be used. If none suit your needs, select Other in"
+        "the rights dropdown."
+    )
```
danloveg, line 78: I like this, but it won't work with translation, so we'll need to remove it.

app/recordtransfer/forms/submission_forms.py
```diff
@@ -698,7 +698,7 @@ class Meta:
 
     def __init__(self, *args, **kwargs):
         super(RightsFormSet, self).__init__(*args, **kwargs)
-        self.forms[0].empty_permitted = False
+        self.forms[0].empty_permitted = True  # Changed to True to make it optional
```
danloveg, line 701: This comment is unnecessary

```suggestion
        self.forms[0].empty_permitted = True
```

app/recordtransfer/forms/submission_forms.py
```diff
@@ -745,7 +745,8 @@ def clean(self) -> dict:
         )
         .order_by("sort_order_other_first"),
         label=gettext("Type of rights"),
-        empty_label=gettext("Please select one"),
+        empty_label=gettext("Select a rights type (optional)"),
+        required=False,
```
danloveg, line 749: This behaviour allows someone to submit a rights form with only the note set, which is behaviour I don't want to allow. See:

Screenshot 2025-06-22 202027

I think this needs to be marked as `required=True`, but that does make it so that there's a red asterisk that makes it seem like the field is always required. I'll let you come up with some solutions to address that.

danloveg (changes requested):
Besides the changes I flagged, there are two other issues with the rights form page.

1. The "Click Here" link doesn't look like a link. This should be fixed.

Screenshot 2025-06-22 201753

2. The tooltips for extra rights forms don't work when you hover your mouse over them. To re-create this issue, click "+ More" on step 5 of the form. The tooltips in the newly created forms don't work.

app/recordtransfer/forms/submission_forms.py
Eashana0104: Maybe to address this, if there is no rights provided in type of rights which is optional, the notes for rights is not even visible. If the user adds a type of right, then the user would be seeing a notes for rights option?

app/recordtransfer/templates/recordtransfer/help.html
Eashana0104: I see the issue here now, but if i just keep the text in the template that would mean i wont be able to reuse the text in help page and rights page of submission form. Is that gonna be okay?

app/recordtransfer/forms/submission_forms.py
Eashana0104: So I made notes for rights only being available to the user when the types of rights is filled, if not Notes for rights doesnt even appear, this way user cannot just submit a note for a right which is not even mentioned. Addressed in https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/727/commits/3f4e061aead8f0a75de6e875e32f5da6ec146951

danloveg (approved):
My first reaction was that showing/hiding the notes field was overly complicating the solution, but on thinking about it and interacting with the form, I think this is neat solution that feels right.

Good job!

Issue #696, opened 2025-06-13
Update Styling of Password Reset Email


danloveg:
The password reset email styling should be updated to match the design of the other emails.

Right now, this email looks like:

![Image](https://github.com/user-attachments/assets/818d5d6e-a45d-40da-9bba-e22a52164b5d)

PR #733, opened 2025-06-23
Update styling of password reset email

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/696

I updated the styling of the password reset email to be consistent with the new email styling.

### DIFF
--- a/app/recordtransfer/templates/registration/password_reset_email.html
+++ b/app/recordtransfer/templates/registration/password_reset_email.html
@@ -1,46 +1,38 @@
-<!DOCTYPE html>
-<html style="
-    font-family: 'Segoe UI', Frutiger, 'Frutiger Linotype', 'Dejavu Sans', 'Helvetica Neue', Arial, sans-serif;
-    margin: 0;">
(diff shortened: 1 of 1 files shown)


yenaing-oo:
@danloveg 
The emails are currently not responsive. Content goes past screen on mobile and user has to scroll to see the full content. The styles needed to address this are not supported by some email clients (such as `max-` and `min-size` for use in media queries), which makes it challenging to achieve responsiveness.

danloveg:
> @danloveg The emails are currently not responsive. Content goes past screen on mobile and user has to scroll to see the full content. The styles needed to address this (such as `max-` and `min-size` for use in media queries) are not supported by some email clients, which makes it challenging to achieve full cross-client responsiveness.

Rather than worry about that too much in this issue, want to open up a new issue for that? I agree that we want them to be responsive but that's beyond the scope of this issue.

If this is ready to be reviewed, want to resolve the merge conflicts and request my review?

Issue #705, opened 2025-06-16
Create non-root user in prod configuration and run the app as that user


danloveg:
It's a security concern for applications in Docker to run as the root user. In the prod part of the `Containerfile`, we should create a non-root user and run the application as that user.

PR #739, opened 2025-06-24
643 add smoke test workflow for podman new

Eashana0104:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/705


Eashana0104:
Key Differences Between Podman and Docker Workflows
1. Database Readiness Check
Docker: No explicit database check.

Podman: Adds a step using mysql --no-defaults to wait for the database.

Reason: Podman's orchestration doesn't guarantee services are fully ready, unlike Docker Compose.

2. Application Startup Detection
Docker: Simple one-step check for the message:
"Listening at: http://0.0.0.0:8000"

Podman: Two-step check: first look for ">> Starting app", then wait for Gunicorn logs.

Reason: Podman streams logs differently and more slowly, so a more robust detection method is needed.

3. Timeouts
Docker: 30-second timeout.

Podman: 60 seconds for the app, 90 seconds for the database.

Reason: Podman containers, especially databases, generally take longer to initialize.

4. Log Capture Strategy
Docker: Uses direct piping (e.g., logs | grep).

Podman: Logs are saved to a file first, then searched (logs > app_logs.txt followed by grep).

Reason: Podman's log output is buffered, making direct piping unreliable.

5. Installation Requirements
Docker: No installation needed (already available on GitHub Actions).

Podman: Requires installation of Podman and podman-compose.

Reason: Podman isn’t pre-installed in GitHub Actions environments.

.github/workflows/smoke-test-prod-podman.yml
```diff
@@ -0,0 +1,108 @@
...
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
+
+          podman --version
+          which podman-compose
+
+      - name: Start containers
+        run: podman-compose -f compose.prod.yml up -d
+
+      - name: Wait for database to be ready
+        run: |
```
danloveg, line 41: The indentation for this block is not consistent with the other blocks. There should be two spaces of indentation after `run: |` and the `until` and `if` blocks are not indented, either.

.github/workflows/smoke-test-prod-podman.yml
```diff
@@ -0,0 +1,108 @@
...
+            if [ $elapsed -ge $timeout ]; then
+            echo "Database initialization timed out"
+            podman-compose -f compose.prod.yml logs db
+            exit 1
+            fi
+
+            echo "Database is ready"
+
+      - name: Wait for app to start
+        run: |
```
danloveg, line 62: Same comment as above. Use two spaces of indentation because that's how it is in all the other `run:` blocks. The `until` block and the first `if` block are not indented , either.

.github/workflows/smoke-test-prod-podman.yml
```diff
@@ -0,0 +1,108 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Copy environment file
+        run: cp example.prod.env .prod.env
+
+      - name: Add SECRET_KEY to environment file
+        run: |
+          echo -e "\nSECRET_KEY=${{ secrets.SECRET_KEY }}" >> .prod.env
+      - name: Set up Podman
```
danloveg, line 26: Add an extra space here for consistency:

```suggestion
          echo -e "\nSECRET_KEY=${{ secrets.SECRET_KEY }}" >> .prod.env

      - name: Set up Podman
```

danloveg (changes requested):
I'm surprised at how much more complicated this new workflow is! But I'm glad to see it's working now.

The new Action workflow file has inconsistent indenting, my comments are here just to make sure this workflow is easy to read.

Eashana0104:
Hi Daniel! I am not sure if I should be making any commits to this branch, As kelvin linked the new PR, we both did some changes for making the workflow better and do its job. In the new PR, there are health checks performed and removing the step to wait for the database.

Issue #643, opened 2025-06-02
Add smoke test workflow for Podman


danloveg:
Right now, Docker is being used for the production smoke test, but we should also be using Podman to make sure the app runs in Podman as well. This is important because we're not often using Podman for development, but all of our documentation for running the application refers to Podman.

PR #743, opened 2025-06-25
Set up Health Check for Production App Container and Smoke Test Workflow using Podman

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/643

- Set up a [health check ](https://docs.docker.com/reference/compose-file/services/#healthcheck) for the production app container
- Used this health check in both smoke test workflows for Docker and Podman to check if app is ready before fetching homepage

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -93,8 +93,15 @@ COPY --from=builder ${PROJ_DIR}/.venv ${PROJ_DIR}/.venv
 COPY --from=builder ${PROJ_DIR}/dist ${PROJ_DIR}/dist
 COPY --from=builder ${APP_DIR} ${APP_DIR}
 
-# Set ownership AFTER copying files
-RUN chown -R myuser:myuser ${PROJ_DIR}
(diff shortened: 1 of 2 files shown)


yenaing-oo:
This PR is an alternative to https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/739. I was testing out https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/739 and came across container health checks, which seems like a cleaner approach. Waiting for the DB to be ready as a separate step in the workflow will not be necessary, as the app container checks for this upon startup in `entrypoint.sh`.

danloveg (approved):
Made a couple minor modifications to get the smoke test workflow to run, but this looks great.

It seems that there was a problem with the `init-volumes` service, since the Podman smoke test workflow was hanging right after that service would exit. I've refactored the permission setting into the Containerfile and now it's working.

Issue #735, opened 2025-06-23
Inefficient Caching When SiteSetting Value Is `None`


yenaing-oo:
The current implementation of `get_value_str` and `get_value_int` methods makes caching inefficient when a setting has a value of `None`. Since Django's cache uses `None` to indicate a cache miss, the code cannot distinguish between "not cached yet" and "value is actually None." As a result, every request for a setting with a real `None` value hits the database rather than using the cache.

See [here](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/fd5878844c002ba1ace8a1f78fd4a3986e6e15d2/app/recordtransfer/models.py#L205-L264) for the referred code section.

I asked Copilot for a possible fix to this and it suggested this:

Introduce a unique sentinel object (e.g. NOT_CACHED = object()) and use it as the default for cache.get(). This allows distinguishing between a cache miss and a cached `None` value.

Example:

```python
NOT_CACHED = object()

def get_value_str(key: SiteSettingKey) -> Optional[str]:
    val = cache.get(key.name, default=NOT_CACHED)
    if val is not NOT_CACHED:
        return val
    # ... fetch from DB and cache result ...
```

This will ensure that even settings with a value of `None` are cached, reducing unnecessary DB queries.

PR #744, opened 2025-06-26
Fix Inefficient Caching When SiteSetting Value Is None

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/735


danloveg (approved):
Smart, a sentinel object is perfect for this use case.

Issue #740, opened 2025-06-25
E2E tests for Help Page


Eashana0104:
Currently, we do not have any end-to-end (e2e) tests written for the help page. This might result in some bugs not being detected and the website failing altogether without us noticing. There needs to be some E2E tests for help page to test its functionality

PR #745, opened 2025-06-26
740 e2e tests for help page

Eashana0104:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/740

This was similar to About page e2e test errors!

The error we were encountering wasn't due to any issues in the code itself, but rather due to vulnerable dependencies in Node.js. Running `npm audit fix` resolved the problem.

Additionally, I faced an issue locating the Help page link from the home page during testing. While the test ran fine locally in VS Code, it was failing in the CI environment. This was because the test window in CI had a smaller viewport, causing the `nav-help` element to be hidden. To address this, I included a toggle button click along with a try-catch block to ensure the element is visible before interacting with it.


package-lock.json
danloveg, line 1: Can you uncommit the changes made to this file? It looks like eslint may have been updated by accident

danloveg (changes requested):

package-lock.json
danloveg, line 1: Can you merge master into this branch? I *think* if you do that, this file shouldn't appear in the list of changed files anymore.

danloveg (changes requested):

Issue #695, opened 2025-06-13
Re-style Reset Password Email


danloveg:
The password reset email should be re-styled, and the `send_test_email` should have this email added to it as well. See sub-issues for more details.

danloveg:
We can use a feature branch to address all the sub-issues.

PR #747, opened 2025-06-26
Restyle reset password email and make job asynchronous

yenaing-oo:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/695


danloveg (approved):
Looks good. I tested out sending the email with the command and that's working well.

PR #764, opened 2025-07-07
Address security issue by upgrading glob

danloveg:
This change is for CVE-2025-5889. The offending package was the brace-expansion package (version 2.0.1), which has been replaced with a different package that does not have this vulnerability.


yenaing-oo (approved):
The version upgrade makes sense, the lock file is also up-to-date.

PR #765, opened 2025-07-09
Fix Group Description does not populate when returning to form

### LINKED ISSUE #757: Group Description does not populate when returning to form
The submission group description does not auto-populate when you return back to the submission group form.

Image

You can replicate this issue by completing the form to the review step (making sure to select a group with a description), and then go back to the submission group step of the form.

yenaing-oo:
Closes #757

Although defined as an async function, `fetchSubmissionGroups` was not actually waiting on the fetch to resolve first (no await is used within . It was returning immediately, even before groups and group descriptions have been fetched & populated. `updateGroupDescription` was being called before `fetchSubmissionGroups` could resolve, meaning the selected option had no `data-attribute` populated.

I've fixed this so that `fetchSubmissionGroups` `await`s for a response before continuing, and updating the group description display upon form initialization.


danloveg (approved):
Nice!

PR #780, opened 2025-07-14
Fix Saving Form from Unsaved Changes Modal Does Not Work on Review Step

### LINKED ISSUE #776: Unsaved changes modal does not work on review step
The unsaved changes modal does not let you save your submission from the review step.

Steps to reproduce:

- Fill out the form to the review step
- Try to click away to the profile page
- In the unsaved changes modal, click "Save form and resume later"

Nothing will happen, and there is an error in the dev console.

**The pull request that resolves this issue should also add an E2E test to catch if this behaviour happens in the future, too.**

Image

yenaing-oo:
Closes #776 

This issue is that the "Click to save form and resume later" button does not exist on the Review Step. The reason for this is that the user would not be interested in saving the form at the final Review Step.

However, the "Save form and resume later" button on the unsaved changes modal relies on the former button to perform the save. To address this, I've made it so that the "Click to save form and resume later" button is included as a hidden button in every step, and is displayed conditionally, so that it can still be used by the modal even when it's not visible.

### DIFF
--- a/app/recordtransfer/templates/recordtransfer/submission_form_base.html
+++ b/app/recordtransfer/templates/recordtransfer/submission_form_base.html
@@ -82,13 +82,14 @@
         </div>
         <div class="{% block savelinkclass %}flex-item{% endblock savelinkclass %}">
             <label></label>
-            {% if SHOW_SAVE_BUTTON %}
-                <button id="form-save-button"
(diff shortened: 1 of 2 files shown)


danloveg (approved):
This works, thanks.

PR #781, opened 2025-07-14
Fix XSS vulnerability in formset.js

danloveg:
Addresses a code quality security issue regarding DOM text being re-interpreted as HTML.


Eashana0104 (approved):
I like how this code solves the XSS vulnerability. This could have been malicious. Good catch! I looked at the code, read a few docs on what this vulnerability is to begin with and i feel the solution makes sense and it doesnt break anything!

PR #788, opened 2025-07-15
Simplify downloading Job files from admin

### LINKED ISSUE #783: Simplify downloading files from Job admin
The `JobAdmin` class has complicated download logic that can be simplified by using Django's default read-only file field for the `attached_file` field.

I believe the following changes can be made to simplify this admin:

- Remove the file `recordtransfer/templates/admin/job_change_form.html`
- Remove the function `recordtransfer.models.Job.get_admin_download_url`
- Remove the function `recordtransfer.models.Job.get_admin_change_url` - unrelated, but it's not used anywhere
- Remove these functions in `recordtransfer.admin.JobAdmin`:
  - `get_urls`
  - `download_file`
  - `get_fields`

By the end, the job's attached file field should look like the green checkmark, and the bottom of the admin should not be present.

Image

yenaing-oo:
Closes #783 

Change notes:
- Added a view to access a Job instance's attached file
- Refactored construction of a response to access a file through nginx into its own function for reusability
- Removed complicated logic to display links to download attached job file


app/recordtransfer/views/media.py
```diff
@@ -253,17 +280,33 @@ def _handle_uploaded_file_get(session: UploadSession, file_name: str) -> HttpRes
...
+    """
+    try:
+        try:
+            job = Job.objects.get(uuid=job_uuid)
+        except Job.DoesNotExist:
+            return HttpResponseNotFound("Job not found")
+
+        if not job.attached_file:
+            return HttpResponseNotFound("File not found for this job")
+
```
yenaing-oo, line 304: Note to self: Restrict requests to this view to only staff

danloveg (approved):

PR #790, opened 2025-07-15
Remove redundant media request view

### LINKED ISSUE #787: Remove redundant `media_request` view
# Description

The [media_request](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/b293e6fa666356be3d0bf3b28556445c04a8e28e/app/recordtransfer/views/media.py#L29) view is unreachable in production due to this [line](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/b293e6fa666356be3d0bf3b28556445c04a8e28e/docker/nginx/templates/nginx.conf.template#L29) in the nginx configuration. Since the nginx server sits in front of the Django app, it will block any `/media/` request without passing it onto the Django app. This means that even if you are a logged-in staff member trying to access a file in the media directory using a valid path, you will get a 404 from nginx.

I tested this out by doing the following in production:
1) Add this line to the top of `media.py:media_request`:
```python
    LOGGER.info("Media request for path: %s", path)
```

2) (Optional) Add this line to the top of `profile.py:UserProfile` which you can use later to check that logging works:
```python
     LOGGER.info("Rendering user profile
```

3) Reuild and run the production containers:
```log
docker compose -f compose.prod.yml up -d --build --remove-orphans
```

4) Generate a downloadable bag from the Submission admin

5) Go to the page for the generated job and copy the `attached_field` text:
Image

6) Append it to the end of this string: `http://localhost:1337/media/` and enter that link in the browser

7) You should be getting a 404 error:

Image

8) There should be no logs coming from `media.py` within the `app` container. However, you should be able to see a similar log to this in `nginx` logging the 404:

```log
2025-07-15 12:12:07 172.19.0.1 - - [15/Jul/2025:17:12:07 +0000] "GET /media/jobs/attachments/admin_20250612-213120_meeting-minutes.zip HTTP/1.1" 404 843 "http://localhost:1337/admin/recordtransfer/job/1/change/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/138.0.0.0 Safari/537.36" "-"
```

# Proposed Solution

Remove the `media_request` view and the `else` conditional block in `app/recordtransfer/urls.py` that includes this endpoint in production.

# Going Forward

We should create endpoints with paths other than `/media/` to access restricted files such as those under `/opt/secure-record-transfer/app/media/`. We can do this using views that perform permission checking and then including headers in the response that tell nginx intercept it, similar to the GET handler for [uploaded_file](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/b293e6fa666356be3d0bf3b28556445c04a8e28e/app/recordtransfer/views/media.py#L191) view.

yenaing-oo:
Closes #787

### DIFF
--- a/app/recordtransfer/tests/unit/views/test_media.py
+++ b/app/recordtransfer/tests/unit/views/test_media.py
@@ -288,57 +288,6 @@ class TestUploadFilesView(TestCase):
         self.assertEqual(session.file_count, 0)
 
 
-class TestMediaRequestView(TestCase):
-    """Test the recordtransfer:media view."""
-
-    @classmethod
-    def setUpClass(cls) -> None:
-        """Disable logging."""
-        super().setUpClass()
-        logging.disable(logging.CRITICAL)
-
-    @classmethod
-    def setUpTestData(cls) -> None:
-        """Create user accounts."""
-        cls.non_admin_user = User.objects.create_user(
-            username="non-admin", password="1X<ISRUkw+tuK"
-        )
-        cls.staff_user = User.objects.create_user(
-            username="admin", password="3&SAjfTYZQ", is_staff=True
-        )
-
-    def test_302_returned_logged_out(self) -> None:
-        """Check that 302 is returned if the client is logged out.
-
-        A 302 is returned because of the login_required() validator.
-        """
-        uri = reverse("recordtransfer:media", kwargs={"path": "test.txt"})
-
-        response = self.client.get(uri)
-
-        self.assertEqual(response.status_code, 302)
-
-    def test_403_returned_not_staff(self) -> None:
-        """Check that 403 is returned if the user is not staff."""
-        self.client.login(username="non-admin", password="1X<ISRUkw+tuK")
-        uri = reverse("recordtransfer:media", kwargs={"path": "test.txt"})
-
-        response = self.client.get(uri)
-
-        self.assertEqual(response.status_code, 403)
-
-    def test_200_returned_staff(self) -> None:
-        """Check that 200 is returned if the user is staff."""
-        self.client.login(username="admin", password="3&SAjfTYZQ")
-        uri = reverse("recordtransfer:media", kwargs={"path": "test.txt"})
-
-        response = self.client.get(uri)
-
-        self.assertTrue(response.has_header("X-Accel-Redirect"))
-        self.assertFalse(response.has_header("Content-Type"))
-        self.assertEqual(response.status_code, 200)
-
-
 class TestUploadedFileView(TestCase):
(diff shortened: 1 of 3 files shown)


danloveg:
> > > So if I understand this correctly, this view was never being hit in production, right? So the user protection I had wrote in that view was never even being used.
> > > We'll have to use the intended behaviour somehow, because I think right now anyone would be able to access any media file.
> > 
> > 
> > I believe `nginx` blocks all incoming requests with a `/media/` path from the client to achieve this behaviour, without passing that request onto Django. I'll double-check that this is the case.
> 
> Yes, in production, `nginx` _does_ block any request with the `/media/` path and responds with a 404: 📷
> 
> The only permitted way to access an uploaded image for example, is to send a GET request to`views.media.uploaded_file` view: 📷

Got it. I see what you're talking about now with creating separate URLs for media files so that `/media/` doesn't get intercepted by NGINX, you did that in #788

yenaing-oo:
> Before I approve this, can you do an audit of where `media_request` was being used? You've covered the job files in #788, but there are more places, such as the temp uploaded files in the form, and the perm uploaded files in the admin.
> 
> Once you figure that out, can you open a new issue to address restricting access to those files in production?

I believe permission checking for uploaded files (both temp and perm) is already being done in the `media.uploaded_file` view. Did you want me to double check this behaviour in production?

yenaing-oo:
> > > Before I approve this, can you do an audit of where `media_request` was being used? You've covered the job files in #788, but there are more places, such as the temp uploaded files in the form, and the perm uploaded files in the admin.
> > > Once you figure that out, can you open a new issue to address restricting access to those files in production?
> > 
> > 
> > I believe permission checking for uploaded files (both temp and perm) is already being done in the `media.uploaded_file` view. Did you want me to double check this behaviour in production?
> 
> If you wouldn't mind, that would give me some peace of mind. I really don't want this to slip through the cracks, it's a big security issue if people can see each others' files.

I've verified that:
- no user (including an admin) can access an uploaded file using the `/media/` URL
- a regular user can only access their own uploaded files using this path: `upload-session/<session_token>/files/<file_name>/`
- an admin can access any user's uploaded files using the same path as above: `upload-session/<session_token>/files/<file_name>/`

- no user (including an admin) can access a job file using the `/media/` URL
- a regular user cannot access any job file using this path: `job/<uuid:job_uuid>/file/`
- an admin can access any job file using this path: `job/<uuid:job_uuid>/file/`

PR #791, opened 2025-07-15
Fix Stretched Country Icon in UserAdmin Change Form

### LINKED ISSUE #782: Fix stretched country icon when editing User in admin
The country icon's aspect ratio is stretched out on the Country field in the User change form. (e.g., `/admin/recordtransfer/user/1/change/`)

Related to #383

Image

yenaing-oo:
Closes #782 

Change notes:
- Duplicate the `CustomCountrySelectWidget` that already exists in `caais.widgets` within `recordtransfer.widgets`

### DIFF
--- a/app/recordtransfer/forms/mixins.py
+++ b/app/recordtransfer/forms/mixins.py
@@ -1,9 +1,9 @@
 from django import forms
 from django.utils.translation import gettext
 from django_countries.fields import CountryField
-from django_countries.widgets import CountrySelectWidget
 
 from recordtransfer.constants import HtmlIds, OtherValues
+from recordtransfer.widgets import CustomCountrySelectWidget
 
 
 class ContactInfoFormMixin(forms.Form):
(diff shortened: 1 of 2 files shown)


yenaing-oo:
@danloveg 
Does my approach of duplicating `CustomCountrySelectWidget` in `recordtransfer.widgets` make sense? Since `recordtransfer` is dependent on `caais` (and not the other way round), I'm questioning whether it'd be okay to use `caais.widgets` within the `recordtransfer` app instead of duplicating like I've currently done.

danloveg:
> @danloveg Does my approach of duplicating `CustomCountrySelectWidget` in `recordtransfer.widgets` make sense? Since `recordtransfer` is dependent on `caais` (and not the other way round), I'm questioning whether it'd be okay to use `caais.widgets` within the `recordtransfer` app instead of duplicating like I've currently done.

Yes, duplicating it makes sense. The more we keep `caais` and `recordtransfer` decoupled, the easier it will be to separate those modules in the future.

danloveg (approved):

PR #795, opened 2025-07-16
Fix Submission Broken in Production

### LINKED ISSUE #794: Submission Broken in Production
In production, upon clicking "Submit" at the end of the Review Step of a Submission Form, this error appears:

Image

The form does not submit, possibly due to an issue with reCAPTCHA validation.

yenaing-oo:
Closes #794 

The submit button was carrying out its default behaviour (submitting the form) when clicked on, without waiting for reCAPTCHA verification to be run first. This meant the reCAPTCHA (`g-recaptcha-response` ) field did not get a chance to be populated before submission, hence the validation error preventing submission.

I've fixed that by preventing the default behaviour of the button when clicked on.

One thing I'm not sure of is why the form managed to submit without issue in development.


danloveg (approved):
Makes sense, thanks for taking a look into this.

PR #799, opened 2025-07-16
Add E2E for successful form submission

### LINKED ISSUE #796: Add E2E test for successful submission
We have an E2E test that fills out the form till the Review Step and checks that the contents of that page is accurate. However, we are lacking a test that checks whether a submission can be made successfully.

yenaing-oo:
Closes #796


danloveg (approved):
You're right, the tests don't really feel slower at all. This looks good!

Issue #784, opened 2025-07-14
Expand Ruff linting to include admin, apps, caais


danloveg:
Add the following files to the `lint-py.yml` workflow and address the issues in each:

- `./app/recordtransfer/admin.py`
- `./app/recordtransfer/apps.py`
- `./app/recordtransfer/caais.py`

PR #803, opened 2025-07-16
784 expand ruff linting to include admin apps caais new

Eashana0104:
I have resolved all the Ruff linting issues in the 3 files ie, apps.py, caais.py, and admin.py, from the recordtransfer directory. There are some errors in those files still but they are not linting errors. For the linting errors, it was typically declaring types of the lists, adding docstrings, adding annotations, and return types for different functions, and unused imports! This PR closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/784


danloveg (approved):
Very nice!

PR #837, opened 2025-07-22
Improve UI for Error Pages

### LINKED ISSUE #789: Improve UI for Error Pages
The error page templates in `app/templates` are too basic and only include a link back to the Home page. The templates should include the navigation bar and footer, and the body should contain user-friendly information to explain the error. This may involve overriding the default views for the error pages. Here is an example to override the 404 view handler:

```python
# in app/app/urls.py
handler404 = 'recordtransfer.views.home.custom_404'
```
See [here](https://docs.djangoproject.com/en/5.2/ref/urls/#handler400) for docs to URL configuration settings for error handlers.

Additionally, `recordtransfer.views.home.SystemErrorPage` is not being used anywhere, and can be used as the custom view for 500 errors.

yenaing-oo:
Closes #789 

Change notes:
- Updated the error page templates in `app/templates` to extend from `app/recordtransfer/templates/recordtransfer/base.html`, so that the header and footer is available.
- Although 400.html and 403.html have also been updated for consistency, these pages are not expected to be shown to the user. In our views, bad requests are typically handled through validation errors returned with a 200 status code, and forbidden resources are generally obscured by returning a 404 status code instead
- I've gone through the views to make sure that server errors are handled by Django where appropriate, and that Http404 exceptions are raised when necessary for Django to handle as well

### DIFF
--- a/app/recordtransfer/views/media.py
+++ b/app/recordtransfer/views/media.py
@@ -190,7 +190,8 @@ def _handle_upload_file(request: HttpRequest, session: UploadSession) -> JsonRes
                 "file": _file.name,
                 "accepted": False,
                 "uploadSessionToken": session.token,
-                "error": gettext(f'File "{_file.name}" is too large to scan for malware'),
+                "error": gettext('File "%(name)s" is too large to scan for malware')
+                % {"name": _file.name},
             },
             status=400,
         )


danloveg (changes requested):
I was able to see the 404 page on the site pretty easily but I am having trouble seeing the 403 and 400 pages. Could you give me a couple URLs I can use to test those pages, or a quick snippet of code I could insert to cause these errors to appear if not?

Issue #836, opened 2025-07-22
Improve E2E test performance by executing in parallel


danloveg:
The E2E tests tend to be really slow when running in pull requests. We can likely speed this up by running them in parallel within Actions.

To do this:

- Install [pytest-xdist](https://github.com/pytest-dev/pytest-xdist). `uv add pytest-xdist --group dev`
- Change the `pytest` command in the E2E workflow to: `uv run pytest -m e2e -n auto`

We'll want to do some testing to make sure that this does in fact speed the tests up.

danloveg:
Doesn't really work in github actions!

PR #840, opened 2025-07-23
836 improve e2e test performance by executing in parallel

Eashana0104:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/836

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -43,4 +43,4 @@ jobs:
       - name: Run E2E tests
         env:
           SELENIUM_TESTS_HEADLESS_MODE: True
-        run: uv run pytest -m e2e -n 2
+        run: uv run pytest -m e2e
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -43,4 +43,4 @@ jobs:
       - name: Run E2E tests
         env:
           SELENIUM_TESTS_HEADLESS_MODE: True
-        run: uv run pytest -m e2e
+        run: uv run pytest -m e2e -n auto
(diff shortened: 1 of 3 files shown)


Eashana0104:
@danloveg I'm not entirely sure if running tests in parallel with our local SQLite database is viable. Even during local development, I am seeing a lot of flakiness in the test runs. There are a few likely reasons:

- Concurrency limitations: SQLite doesn’t handle multiple concurrent writes well. When using parallel test workers, this leads to database locking issues and overall instability.

- Performance bottlenecks: While SQLite performs well in simple, single-user setups, it tends to struggle under the load of parallel testing.

I’m also unsure whether switching to MySQL for local testing would be the right move — it might resolve these issues but could complicate the setup for developers.

Aside from this, we’re seeing parallel test runs fail entirely on CI, likely due to differences in the testing environment compared to local setups (e.g. available resources, OS-level behavior, or concurrency handling). These inconsistencies are making reliable parallel testing a challenge across the board.

danloveg:
> @danloveg I'm not entirely sure if running tests in parallel with our local SQLite database is viable. Even during local development, I am seeing a lot of flakiness in the test runs. There are a few likely reasons:
> 
>     * Concurrency limitations: SQLite doesn’t handle multiple concurrent writes well. When using parallel test workers, this leads to database locking issues and overall instability.
> 
>     * Performance bottlenecks: While SQLite performs well in simple, single-user setups, it tends to struggle under the load of parallel testing.
> 
> 
> I’m also unsure whether switching to MySQL for local testing would be the right move — it might resolve these issues but could complicate the setup for developers.
> 
> Aside from this, we’re seeing parallel test runs fail entirely on CI, likely due to differences in the testing environment compared to local setups (e.g. available resources, OS-level behavior, or concurrency handling). These inconsistencies are making reliable parallel testing a challenge across the board.

This all makes sense. For now, I'll close this PR and the issue as well. It was worth a try!

PR #841, opened 2025-07-23
Fix archivist_user group, remove transfer_user

### LINKED ISSUE #819: Fix archivist_user group, remove transfer_user
If a user is a staff member, but is *not* a superuser, they see nothing when going to `/admin/`. The `archivist_user` is intended to add some basic staff functionality without providing full superuser access to the site.

Additionally, there is a group that has no function, and should be removed called `transfer_user`.

This is what an `archivist_user` sees right now when going to `/admin/` (many things are missing):

Image

What I would like to see for an `archivist_user` is:

Image

Basically, everything in `caais` and `recordtransfer`, except the **Site settings** model should be read-only for these users.

yenaing-oo:
Closes #819 

Change notes:
- I refactored the migrations so that all the permissions for `archivist_user` is set in the latest migration file. This should be fine since the app is still in development.
- I've also had to add some checks in the admin forms to avoid `KeyError`s due to certain fields not showing for non-superusers


yenaing-oo:
@danloveg 
Do we want everything that shows up for a superuser when editing a `Metadata` instance to also show up for an `archivist_user`?:

For superuser:
image

Currently, for an `archivist_user`:
image

danloveg:
> @danloveg Do we want everything that shows up for a superuser when editing a `Metadata` instance to also show up for an `archivist_user`?:

Yes, the archivist user should have complete freedom when editing `Metadata` models. I'll wait for you to add that permission and then I'll add my review.

danloveg:
> > > @danloveg Do we want everything that shows up for a superuser when editing a `Metadata` instance to also show up for an `archivist_user`?:
> > 
> > 
> > Yes, the archivist user should have complete freedom when editing `Metadata` models. I'll wait for you to add that permission and then I'll add my review.
> 
> I've actually already gone ahead and made that change, since I figured that would be the case (see [d98ebd5](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/commit/d98ebd5e5dba33b40834e10b08f7153992b459ec))

Great, I'll take a look now

Issue #838, opened 2025-07-23
Add UUID field to `User` model and use in views


yenaing-oo:
There is a URL path that includes the user's id (the primary key) as its argument -  `get_user_submission_groups`.  A UUID field should be added to the `User` model and used for the mentioned path instead, as this obscures details such as the number of users and the IDs of other users.

PR #842, opened 2025-07-23
838 Remove user_id being accessible in get_submission_group() url

Eashana0104:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/838

This PR addresses a security concern by removing the` user_id` from the `get_submission_group` link, which was previously exposing user identifiers. Instead of relying on the passed user_id, the updated implementation now securely retrieves the groups directly for the authenticated user making the request.

The functionality has been tested and verified to ensure everything works as expected without relying on the exposed identifier.


Eashana0104:
To test this PR, I navigated to Step 7 of the submission form—Assign Submission to Group. Using the browser’s developer tools, I inspected the network requests and located the API call that fetches all groups associated with the specific user. This confirmed that the UUID is now being used instead of the user ID.
Screenshot 2025-07-23 at 4 33 56 PM

app/recordtransfer/views/post_submission.py
```diff
@@ -204,9 +204,15 @@ def form_invalid(self, form: BaseModelForm) -> HttpResponse:
...
-    """Retrieve the groups associated with the current user."""
-    user = get_object_or_404(User, pk=user_id)
+def get_user_submission_groups(request: HttpRequest, user_uuid: str) -> JsonResponse:
+    """Return a JSON response containing all submission groups created by the specified user."""
+    user = get_object_or_404(User, uuid=user_uuid)
+    if request.user != user and not request.user.is_staff and not request.user.is_superuser:
+        return JsonResponse(
+            {"error": gettext("You do not have permission to view these groups.")},
+            status=403,
+        )
```
danloveg, line 214: This code was removed in #837, can you remove this?

app/recordtransfer/views/post_submission.py
```diff
@@ -204,9 +204,15 @@ def form_invalid(self, form: BaseModelForm) -> HttpResponse:
         )
 
 
-def get_user_submission_groups(request: HttpRequest, user_id: int) -> JsonResponse:
-    """Retrieve the groups associated with the current user."""
-    user = get_object_or_404(User, pk=user_id)
+def get_user_submission_groups(request: HttpRequest, user_uuid: str) -> JsonResponse:
```
danloveg, line 207: There are no unit tests for this view. Can you add some in `recordtransfer.tests.unit.views.test_post_submission`?

danloveg (changes requested):

app/recordtransfer/views/post_submission.py
Eashana0104: I was reviewing the code and noticed a potential security issue during testing. Currently, it seems that any user can access the groups associated with a specific user, which shouldn’t be allowed.

The issue appears to stem from the line:

`user = get_object_or_404(User, uuid=user_uuid)`

This only verifies whether a user with the given UUID exists in the database — it doesn't check whether the requesting user is the same as the one associated with that UUID.

To enforce proper access control, we should explicitly verify that the user making the request is indeed the same user who owns the data. Removing this check is likely what introduced the vulnerability.

danloveg: I think this view also needs a filter on what type of HTTP methods it allows:

```python
@require_http_methods(["GET"])
def get_user_submission_groups(request: HttpRequest) -> JsonResponse:
    # ...
```

danloveg: From the main site, that'd be true, yes. Staff can still see people's submission groups from the admin site, that's good enough. There's no need for a staff member to request a user's groups via this view

Eashana0104: I see this makes sense, I have added a test to portray that as well!

app/recordtransfer/views/post_submission.py
Eashana0104: Added some unit tests here https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/842/commits/5930df08fc8cbd12065f677d2666b1b02cf7aa32. Let me know if there are some scenarios I missed and should include in the tests perhaps.

app/recordtransfer/views/post_submission.py
Eashana0104: Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/842/commits/8126c156bcf9a95070a767f3ac098a9ab6ed32a2, https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/842/commits/5b29d682692c771308a5ba2f10d4bae96cfede31,
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/842/commits/b097b37c98d38d2cfc38255ee69606d0b5daa772,
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/842/commits/76bb7de4beea9a001fb50dbccf3624888a0cf1e1, 
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/842/commits/80fa54d9f15e56c7ceaff805fef240c00f084b99

danloveg:
> Checked the logs on Docker and Network tabs on Google chrome, the HTTP get request to get_submission_group works and does not contain user_id now!

Great!

I'm going to review this now, but can you please change the title of the pull request? There's no UUID field on User now, but you'd think there was from the PR title.

PR #854, opened 2025-07-24
Fix login redirects being ignored

### LINKED ISSUE #850: Fix login redirects being ignored
Steps to re-produce:

- Start on the homepage, **not logged in**
- Click "Submit Your Records"
- Log in with your account credentials

You are redirected to the homepage, rather than to the "next" URL. The URL has the proper location the user should be sent to next:

`/account/login/?next=/submission/`

But the next parameter is being ignored. We should update the login view to obey the next parameter, if it's set.

yenaing-oo:
Closes #850 

The main issue was that the URL to which the login form data was posted to did not include the `?next=` query parameter.

### DIFF
--- a/app/recordtransfer/templates/registration/login.html
+++ b/app/recordtransfer/templates/registration/login.html
@@ -13,7 +13,7 @@
             <!-- Form -->
             <div id="login-form-container">
                 <form class="space-y-4"
-                      hx-post="{% url 'login' %}"
+                      hx-post="{% url 'login' %}?next={{ next }}"
                       hx-target="#login-error-container"
                       hx-swap="innerHTML">
                     {% csrf_token %}
--- a/app/app/urls.py
+++ b/app/app/urls.py
@@ -30,7 +30,7 @@ urlpatterns = [
     # Override the login view with redirect behavior
     path(
         "account/login/",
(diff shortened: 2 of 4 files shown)


app/recordtransfer/templates/registration/login.html
```diff
@@ -13,7 +13,7 @@
             <!-- Form -->
             <div id="login-form-container">
                 <form class="space-y-4"
-                      hx-post="{% url 'login' %}"
+                      hx-post="{% url 'login' %}?next={{ next }}"
```
danloveg, line 16: Just a question, `{{ next }}` may be empty, does it matter to worry about that case? I see it re-directs just fine even when next is in the parameters but is empty, just wondering if you think we need to consider that at all.

We could do something like this:

```suggestion
                      hx-post="{% url 'login' %}{% if next }?next={{ next }}{% endif %}"
```

danloveg (commented):
This all makes sense, but I had one question before approving (see below).

app/recordtransfer/templates/registration/login.html
yenaing-oo: I think it wouldn't hurt to do it this way to avoid any unexpected issues.

danloveg (approved):

PR #856, opened 2025-07-24
Prevent brute force login attempts using `django-axes`

### LINKED ISSUE #844: Add `django-axes` to prevent brute force login attempts
Consider adding https://github.com/jazzband/django-axes to prevent users from trying to brute force logging in to the application.

There a number of configuration options outlined here: https://django-axes.readthedocs.io/en/latest/4_configuration.html#configuring-project-settings.

By default, it looks like users are locked out permanently. We should consider maybe having an `AXES_COOLOFF_TIME`.

yenaing-oo:
Closes #844


app/app/settings/base.py
```diff
@@ -43,8 +51,18 @@
     "django.middleware.locale.LocaleMiddleware",
     "django.middleware.gzip.GZipMiddleware",
     "django_htmx.middleware.HtmxMiddleware",
+    # AxesMiddleware should be the last middleware in the MIDDLEWARE list.
+    "axes.middleware.AxesMiddleware",
 ]
 
+# Axes configuration
+AXES_FAILURE_LIMIT = config("AXES_FAILURE_LIMIT", default=5, cast=int)
+AXES_COOLOFF_TIME = config("AXES_COOL_OFF_TIME", default=0.01, cast=float)  # in hours
```
danloveg, line 60: Can you add these to the documentation? The documentation doesn't have to be extensive, but maybe you could link to the django-axes documentation for each of these

danloveg, line 60: I think thirty six seconds is too short. Let's make it half an hour by default (people can always change it)

```suggestion
AXES_COOLOFF_TIME = config("AXES_COOL_OFF_TIME", default=0.5, cast=float)  # in hours
```

danloveg (changes requested):
Can you add a warning message if the user has two or fewer remaining attempts to log in before they are locked out?

app/app/settings/base.py
yenaing-oo: Yes, 36 seconds is way too short. I initially had it at 15 mins but I must've accidentally changed it during testing.

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/856/commits/94d8aa86bfc39e7278bc2f35ad0838df8734eb62

yenaing-oo:
> Can you add a warning message if the user has two or fewer remaining attempts to log in before they are locked out?

That's a good idea, I've added that here: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/856/commits/ecfa63c67456aba1d3732fa1bc829a4010c31a5e

I've also added verification for those settings, as well as tests for the toast events.

app/recordtransfer/views/account.py
```diff
@@ -122,18 +129,56 @@ def form_valid(self, form: AuthenticationForm) -> HttpResponse:
...
+            response = trigger_client_event(
+                response,
+                "showWarning",
+                {
+                    "value": gettext(
+                        "You have %d login attempt(s) left before your account is locked out."
+                    )
+                    % num_tries_left,
+                },
+            )
```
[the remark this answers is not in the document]
yenaing-oo: That's a pretty neat thing to have - will be using this in the future.

danloveg (approved):
I will merge this in for August 2025

Issue #861, opened 2025-07-25
Update submission form docs


danloveg:
[The submission form documentation](https://secure-record-transfer.readthedocs.io/en/latest/application_overview/submission-form.html) needs to be updated again because we've re-styled everything in the submission form.

PR #869, opened 2025-07-25
Update submission form docs

Eashana0104:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/861

- Updated all images to reflect the newly restyled submission form
- Added documentation for the new progress bar feature on the submission form
- Included details on the behavior of the Save button, including how it remains accessible throughout the form
- Included explanation on how donors can now see the remaining upload limit (in MB) while uploading files


danloveg (changes requested):
Can you crop the images so that they don't have the header or footer? You don't need to take new screenshots, just remove the header and footer from the images you took.

There are also some images that have a grey bar on the left, that should be removed as well.

For example, this image:

Image

Should look like this. Notice I cropped the left grey bar, and I cropped out the header and the footer.

Image

If we change the footer or header, we don't need to also update these images.

Eashana0104:
I have cropped all the required images accordingly. Resolved by https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/869/commits/e17e288ef4cdd84016b34c61ab011604eb7cc8f7

PR #876, opened 2025-07-28
Set up Javascript translations

### LINKED ISSUE #831: Set up Javascript translations
Set up Javascript translation. This is a bit tricky due to [the issues mentioned here](https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#internationalization-in-javascript-code).

From what I understand, we need to:

1. Add the [JavaScriptCatalogView](https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#module-django.views.i18n)
2. Include the [JavaScript catalog](https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#using-the-javascript-translation-catalog) in the base template.
3. Add `gettext` calls in some of the JavaScript files. [Here is a good example of where this would be needed](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/8be73c181b97ac369ccac38eed360d559d68cb02/app/recordtransfer/static/recordtransfer/js/submission_form/customUppyPlugin.js#L30).
4. Ensure `makemessages -e djangojs` can extract messages from JavaScript files.
5. Translate some of the strings extracted from JavaScript.
6. Ensure that the string translations are loaded from the catalog in the JavaScript.

yenaing-oo:
Closes #831

### DIFF
--- a/app/recordtransfer/utils.py
+++ b/app/recordtransfer/utils.py
@@ -1,7 +1,6 @@
-import glob
-import hashlib
 import logging
 import os
+from pathlib import Path
 from typing import TYPE_CHECKING, Union
 from zipfile import ZipFile
 


danloveg:
That's interesting, the `/jsi18n/` endpoint returns translations depending on the language in the `django_language` cookie.

I'm not seeing the translations, do I have to run `compilemessages`?

danloveg:
> That's interesting, the `/jsi18n/` endpoint returns translations depending on the language in the `django_language` cookie.
> 
> I'm not seeing the translations, do I have to run `compilemessages`?

Looks like yes. I'll add an issue to compile those files in the Containerfile

yenaing-oo:
Without caching, subsequent fetches take:
image

With caching, subsequent fetches take less time:
image

app/app/urls.py
```diff
@@ -34,4 +35,5 @@
         name="login",
     ),
     path("account/", include("django.contrib.auth.urls")),
-]
+    path("jsi18n/", JavaScriptCatalog.as_view(), name="javascript-catalog"),
```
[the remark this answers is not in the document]
yenaing-oo: `get_version()` is a custom function left to be implemented according to the user's needs. See this helpful comment in Django's forums: https://code.djangoproject.com/ticket/6195#comment:9

app/recordtransfer/utils.py
```diff
@@ -354,3 +356,18 @@ def accept_session(filename: str, filesize: Union[str, int], session: "UploadSes
...
+    """
+    base_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+    locale_dir = os.path.join(base_dir, "locale")
+    mo_files = glob.glob(os.path.join(locale_dir, "*/LC_MESSAGES/djangojs.mo"))
+    latest_mtime = 0
+    for mo_file in mo_files:
+        mtime = os.path.getmtime(mo_file)
+        if mtime > latest_mtime:
+            latest_mtime = mtime
+    return hashlib.md5(str(latest_mtime).encode()).hexdigest()
```
danloveg, line 373: There are a few simplifications we can make to this function:

- Use `max` instead of manually tracking the latest time.
- Use `LOCALE_PATHS` instead of finding them yourself.
- Use `pathlib.Path` instead of glob. It has a very nice [rglob](https://docs.python.org/3/library/pathlib.html#pathlib.Path.rglob) function
- Just use the time instead of hashing the time. The time is unique enough, it doesn't also need to be hashed.

```suggestion
def get_js_translation_version() -> str:
    """Return the latest modification time of all djangojs.mo files in the locale directory.

    This changes whenever compiled JS translations are updated.
    """
    return str(
        max(
            [
                item.stat().st_mtime
                for locale_dir in settings.LOCALE_PATHS
                for item in Path(locale_dir).rglob("djangojs.mo")
            ]
        )
    )
```

danloveg (changes requested):
Let me know what you find regarding how often `get_js_translation_version` is being called.

I do have some suggestions for simplifying that function (see below)

app/recordtransfer/utils.py
yenaing-oo: Beautiful - so much cleaner.

yenaing-oo:
I want to see if it'll be easy to implement client-side caching as well.

yenaing-oo:
> I want to see if it'll be easy to implement client-side caching as well.

I couldn't get client-side caching to work with [ConditionalGetMiddleware](https://docs.djangoproject.com/en/5.2/ref/middleware/#django.middleware.http.ConditionalGetMiddleware), which sets [ETags](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/ETag) on responses for versioning and client-side caching.

See:
- https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#note-on-performance
- https://docs.djangoproject.com/en/5.2/ref/middleware/#module-django.middleware.http
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/ETag

yenaing-oo:
It might not be worth the trouble getting this to work since the response time is around 16ms, which is already pretty fast.

PR #881, opened 2025-07-28
Compile messages on app start

### LINKED ISSUE #878: Compile messages in Containerfile
Before the app is started, the messages should be compiled so that they're ready before the app starts up.

I don't know if a separate `compilemessages` is required for each language, but this command will at least compile messages for French:

```shell
uv run python manage.py compilemessages -l fr
```

danloveg:
Closes #878 

Runs `compilemessages` when the image is run so that `.mo` files are present when the app starts.

`gettext` was missing from the production container so I've added that in.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -14,6 +14,8 @@ if [ "$SERVICE_NAME" = 'rq' ]; then
   python manage.py migrate --no-input
   echo ">> Verifying settings."
   python manage.py verify_settings
+  echo ">> Compiling messages."
+  python manage.py compilemessages --ignore .venv
   echo ">> Starting RQ worker(s)"
 
 elif [ "$SERVICE_NAME" = 'rq-scheduler' ]; then
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -14,8 +14,16 @@ if [ "$SERVICE_NAME" = 'rq' ]; then
   python manage.py migrate --no-input
   echo ">> Verifying settings."
   python manage.py verify_settings
-  echo ">> Compiling messages."
-  python manage.py compilemessages --ignore .venv
+
+  # This is run in the containerfile for production. This is included here for
+  # dev because the locale/ directory is in the app/ folder, which gets shared
+  # as a volume locally with the container; this means the compiled messages
+  # would get overwritten when the volume is shared.
+  if [ "$ENV" = 'dev' ]; then
+    echo ">> Compiling messages."
+    python manage.py compilemessages --ignore .venv
+  fi
+
   echo ">> Starting RQ worker(s)"
 
 elif [ "$SERVICE_NAME" = 'rq-scheduler' ]; then
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -15,7 +15,7 @@ if [ "$SERVICE_NAME" = 'rq' ]; then
   echo ">> Verifying settings."
   python manage.py verify_settings
 
-  # This is run in the containerfile for production. This is included here for
+  # This runs in the containerfile for production. This is included here for
   # dev because the locale/ directory is in the app/ folder, which gets shared
   # as a volume locally with the container; this means the compiled messages
   # would get overwritten when the volume is shared.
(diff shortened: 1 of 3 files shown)


docker/entrypoint.sh
```diff
@@ -14,6 +14,8 @@ if [ "$SERVICE_NAME" = 'rq' ]; then
   python manage.py migrate --no-input
   echo ">> Verifying settings."
   python manage.py verify_settings
+  echo ">> Compiling messages."
+  python manage.py compilemessages --ignore .venv
```
yenaing-oo, line 18: I'm conflicted between having this called in the Containerfile vs here. In production, the translations would have no reason to change, so it makes sense for this to be part of the container build. In development, where translations are prone to change, having this as part of the entrypoint makes sense (although a manual restart of the app service would be required for this to take effect - simply calling `compilemessages` during runtime would not work). It's possible to cater to both situations using the `$ENV` environment variable.

The alternative is to leave this as is to avoid complexity. What do you think?

yenaing-oo: The change you made in https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/commit/fb7f6c5da8329517eefbcfa76c42f14510e7b26b makes sense to me, and is what I would've done as well.

> I put it here because in dev, generating the `.mo` files in the Containerfile, then sharing a volume with your local directory overwrites those files.

Although, I don't quite understand why overwriting the `.mo` files in our local directory would be an issue. Additionally, wouldn't compiling in the entrypoint overwrite the `.mo` files as well? Those files aren't being ignored in `.dockerignore`.

My reason for wanting to compile in the entrypoint in dev is so that it gets run automatically whenever the service is restarted.

danloveg: There are two reasons why overwriting the `.mo` files is an issue in dev.

# 1. Local .mo Files Don't Exist Yet

- We don't commit `.mo` files because of the [.gitignore](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/7be6f6116e817a755917da9dfcb95804a26073d5/.gitignore#L73). So, you might not have any yet.
- You run `compilemessages` in the Containerfile, which will create `.mo` files. These will be in `/opt/secure-record-transfer/app/locale/...` in the image.
- But when [this volume](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/7be6f6116e817a755917da9dfcb95804a26073d5/compose.dev.yml#L95) is shared with the dev image, everything at `/opt/secure-record-transfer/app/` *in the image* is effectively replaced with what's in your local directory at `./app/`.

Any "new" files you tried making in the container within `/opt/secure-record-transfer/app/` get removed when this volume is specified. This includes the `.mo` files.

# 2. Local .mo Files are Outdated

- You have some uncommited `.mo` files locally. But these are old, and haven't been compiled in a while.
- You run `compilemessages` in the Containerfile, which will create `.mo` files. These will be in `/opt/secure-record-transfer/app/locale/...` in the image.
- But when [this volume](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/7be6f6116e817a755917da9dfcb95804a26073d5/compose.dev.yml#L95) is shared with the dev image, everything at `/opt/secure-record-transfer/app/` *in the image* is effectively replaced with what's in your local directory at `./app/`. Your old `.mo` files overwrite the fresh ones you generated.

This could be an issue because you would be using stale `.mo` files without realizing it.

---

The reason this is not an issue in production is because there is no volume shared locally with the production container. There is no risk of removing `.mo` files or providing stale `.mo` files.

---

I'll say, none of this is an issue with the Webpack generated static files because those are *outside of* `/opt/secure-record-transfer/app/`. They're one directory above. The reason we're running into this issue here is because the image is trying to generate files that are within a shared volume.

We could address this by moving the locale directory to the root of the repository but I'm not sure how common a pattern that is in Django.

danloveg: The way it's written now works because the entrypoint script runs *after* the volume is created.

yenaing-oo (approved):
Looks and works good, thank you for the explanation!

docker/entrypoint.sh
danloveg: It took a while to wrap my head around too!

Issue #889, opened 2025-07-29
Long file names overflow Step 8 summary card on Review step


Eashana0104:
On the Review step of the submission form, the Step 8 summary section displaying uploaded files visually breaks when file names are too long. The text spills out of the card container, affecting the layout and overall readability.

We should ensure that long file names:

Wrap or truncate appropriately within the card

Do not overflow or distort the layout

<img width="592" height="730" alt="Image" src="https://github.com/user-attachments/assets/a71a80f3-3ea2-4caf-9060-4a5e66eb677a" />

PR #891, opened 2025-07-29
889 long file names overflow step 8 summary card on review step

Eashana0104:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/889

Long file names now break and wrap to the next line if they exceed the container width. They break at any character though.  Using word-break: break-all ensures that overly long names don’t overflow the layout.

I avoided using word-break: keep-all because in many browsers, file names (especially with underscores or no spaces) don’t break naturally, which could cause overflow issues again.

I also tried using overflow-wrap:break-word, it didn't work; it was still overflowing even after breaking, and it wasn't understanding the file-entry boundaries

### DIFF
--- a/app/recordtransfer/templates/recordtransfer/submission_form_review.html
+++ b/app/recordtransfer/templates/recordtransfer/submission_form_review.html
@@ -40,9 +40,7 @@
                                                         <i class="fas fa-file text-primary/60"></i>
                                                         <a class="non-nav-link text-primary/60 font-medium underline hover:text-primary transition break-all block"
                                                            style="word-break: break-all;
-                                                                  overflow-wrap: anywhere;
-                                                                  max-width: 100%;
-                                                                  display: block"
+                                                                  max-width: 100%"
                                                            href="{{ file.url }}"
                                                            target="_blank">{{ file.name }}</a>
                                                     </div>
--- a/app/recordtransfer/templates/recordtransfer/submission_form_review.html
+++ b/app/recordtransfer/templates/recordtransfer/submission_form_review.html
@@ -38,9 +38,11 @@
                                                 {% if forloop.counter <= 5 %}
                                                     <div class="file-entry mb-2 flex items-center gap-2">
                                                         <i class="fas fa-file text-primary/60"></i>
-                                                        <a class="non-nav-link text-primary/60 font-medium underline hover:text-primary transition block"
+                                                        <a class="non-nav-link text-primary/60 font-medium underline hover:text-primary transition break-all block"
                                                            style="word-break: break-all;
-                                                                  max-width: 100%"
+                                                                  overflow-wrap: anywhere;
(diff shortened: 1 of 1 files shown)


app/recordtransfer/templates/recordtransfer/submission_form_review.html
```diff
@@ -38,7 +38,9 @@
                                                 {% if forloop.counter <= 5 %}
                                                     <div class="file-entry mb-2 flex items-center gap-2">
                                                         <i class="fas fa-file text-primary/60"></i>
-                                                        <a class="non-nav-link text-primary/60 font-medium underline hover:text-primary transition"
+                                                        <a class="non-nav-link text-primary/60 font-medium underline hover:text-primary transition break-all block"
+                                                           style="word-break: break-all;
+                                                                  max-width: 100%"
```
danloveg, line 43: You should be able to just use Tailwind classes rather than include inline styles.

1. For `word-break: break-all;`, the `break-all` class you've included already does that.

Image

2. For `max-width: 100%;`, you could just use [max-w-full](https://tailwindcss.com/docs/max-width#using-a-percentage)

```suggestion
                                                        <a class="non-nav-link text-primary/60 font-medium underline hover:text-primary transition break-all max-w-full block"
```

app/recordtransfer/templates/recordtransfer/submission_form_review.html
```diff
@@ -56,7 +58,9 @@
                                                         {% if forloop.counter > 5 %}
                                                             <div class="file-entry mb-2 flex items-center gap-2">
                                                                 <i class="fas fa-file text-primary/60"></i>
-                                                                <a class="non-nav-link text-primary/60 font-medium underline hover:text-primary transition"
+                                                                <a class="non-nav-link text-primary/60 font-medium underline hover:text-primary transition break-all block"
+                                                                   style="word-break: break-all;
+                                                                          max-width: 100%"
```
danloveg, line 63: Same comment as above

```suggestion
                                                                <a class="non-nav-link text-primary/60 font-medium underline hover:text-primary transition break-all max-w-full block"
```

danloveg (changes requested):

app/recordtransfer/templates/recordtransfer/submission_form_review.html
Eashana0104: That makes a lot more sense now — I was initially confused about why the DaisyUI classes didn’t seem to be working, but it turns out I was testing on the wrong server at the time. That was my oversight. This approach definitely simplifies the code and works as expected!

Resolved by 
 https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/891/commits/6e4914ef70ffee0ed992770ae7630d4a3b43a9ea

danloveg (approved):
This fixes the issue, thanks.

PR #892, opened 2025-07-29
Fix job zip file names

### LINKED ISSUE #890: .zip files in job are downloading with random names
This affects the production configuration.

Clicking this file name:

Image

Downloads a zip file named `zQlPQxxH.zip`.

The intended behaviour is that the zip file downloaded matches the name in the link.

danloveg:
Closes #890 

Adds the **Content-Type** and **Content-Disposition** headers to zip file responses. This makes it so NGINX does not change the file name when it's downloaded.

This change can only be tested in production mode.


danloveg:
> Everything looks great to me — the code is clean, the behavior is exactly as expected, and it's already well-tested. The implementation feels refined and thoughtful.

Thanks for testing it!

PR #896, opened 2025-07-30
Fix Page Flicker While Navigating Between Submission Form Steps

### LINKED ISSUE #895: Page Flicker While Navigating Between Submission Form Steps
The page flickers a bit moving between steps of the submission form. To improve user experience, we might consider adding a skeleton component to display while the step data is being processed, and the next step is being rendered. Or perhaps we could add a transition effect.

yenaing-oo:
Closes #895


yenaing-oo:
> > Hi Kelvin! I was looking into this issue as well and hadn’t realized you were already working on it. I just tested the updated functionality, and it’s working really well now. While exploring it, I also found an alternative approach — simply adding transition:true to hx-swap seems to handle the transitions smoothly without any flickering. Let me know what you think about this!
> 
> Nicee, the transition looks even smoother with that. Do you want to make that edit on my branch directly?

Or open up a separate PR

yenaing-oo:
> > > Hi Kelvin! I was looking into this issue as well and hadn’t realized you were already working on it. I just tested the updated functionality, and it’s working really well now. While exploring it, I also found an alternative approach — simply adding transition:true to hx-swap seems to handle the transitions smoothly without any flickering. Let me know what you think about this!
> > 
> > 
> > Nicee, the transition looks even smoother with that. Do you want to make that edit on my branch directly?
> 
> How about i create a new branch regarding this, its just a small line change and the PR might look cleaner that way

Yeap, I figured that would be better. Please go ahead.

yenaing-oo:
Closed in favour of https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/897

PR #897, opened 2025-07-30
Fix Flickering by Enabling Transitions in HTMX Swaps

Eashana0104:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/895

This update adds transition:true to the hx-swap attribute, allowing HTMX to automatically handle transitions before and after content swaps. This results in a much smoother visual experience and eliminates the flickering that was previously occurring during swaps.


danloveg (approved):
Looks good!

PR #898, opened 2025-07-30
Restrict admin access to submission group detail page

### LINKED ISSUE #893: Restrict admin access to submission group detail page
Currently, when an admin accesses another user's submission group detail page, they will see no submissions even if that group _does have submissions_. This is because only the request user's submissions are displayed.

To avoid confusion from this, an admin user should not be allowed access to a submission group detail page unless they own the group themselves. 

Additionally, links to view the submission group detail page of submission groups from the admin site should be removed.

yenaing-oo:
Closes #893

### DIFF
--- a/app/recordtransfer/admin.py
+++ b/app/recordtransfer/admin.py
@@ -370,6 +370,7 @@ class SubmissionGroupAdmin(ReadOnlyAdmin):
         "export_atom_2_2_csv",
         "export_atom_2_1_csv",
     ]
+    view_on_site = False
 
     def has_delete_permission(self, request: HttpRequest, obj: object = None) -> bool:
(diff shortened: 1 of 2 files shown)


yenaing-oo:
@danloveg 
> I think enabling changing the group in the admin is as easy as removing [this line](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/cc9aede19d70bf29a9ce556ef8385a1741ac501b/app/recordtransfer/forms/admin_forms.py#L113)

Removing that line didn't seem to do anything. Additionally, I think this would require a bit more thought, as we want to restrict setting submission groups which only belong to the user that made that submission.

Issue #827, opened 2025-07-22
Create language switcher


danloveg:
Create a language switcher for the application. I think the best place for this would be right at the top of the screen inside the `banner-container` element.

There is an example view for how to make this work:

https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#the-set-language-redirect-view

I'd suggest writing an E2E test that checks that when the language is switched to Hindi, that the language cookie is set properly.

See:

- https://www.selenium.dev/documentation/webdriver/interactions/cookies/
- https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#language-cookie

I think the one that contains the name of the language is `LANGUAGE_COOKIE_NAME`

PR #904, opened 2025-07-31
827 create language switcher

Eashana0104:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/827

Change Notes:

Added a URL route for language switching using Django’s `set_language` view, which is responsible for setting the language cookie.

Implemented a language switcher form within the `banner-container`. This dropdown dynamically lists the languages defined in the `LANGUAGES` setting in `base.py`.

Added an end-to-end (E2E) test to verify that the `LANGUAGE_COOKIE_NAME` is correctly set to `'hi'` when Hindi is selected from the language dropdown.

### DIFF
--- a/app/recordtransfer/tests/e2e/test_language_switch.py
+++ b/app/recordtransfer/tests/e2e/test_language_switch.py
@@ -7,8 +7,6 @@ from selenium.webdriver.common.by import By
 from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.support.ui import Select, WebDriverWait
 
-from app.settings.base import LANGUAGE_COOKIE_NAME
-
 from .selenium_setup import SeleniumLiveServerTestCase
 
 
@@ -36,10 +34,10 @@ class TestLanguageSwitch(SeleniumLiveServerTestCase):
         Select(lang_dropdown).select_by_value("hi")
 
         def get_language_cookie(driver: webdriver.Remote):
-            cookie = driver.get_cookie(LANGUAGE_COOKIE_NAME)
+            cookie = driver.get_cookie(settings.LANGUAGE_COOKIE_NAME)
             if cookie is not None and cookie.get("value") == "hi":
                 return cookie
             return None
 
         language_cookie = WebDriverWait(driver, 10).until(get_language_cookie)
-        assert language_cookie is not None, f"{LANGUAGE_COOKIE_NAME} cookie was not set"
+        self.assertIsNotNone(language_cookie)


app/recordtransfer/templates/recordtransfer/banner.html
```diff
@@ -1,6 +1,21 @@
...
+                {% get_available_languages as LANGUAGES %}
+                {% get_language_info_list for LANGUAGES as languages %}
+                {% for language in languages %}
+                    <option value="{{ language.code }}"
+                            {% if language.code == LANGUAGE_CODE %}selected{% endif %}>
+                        {{ language.name_local }} ({{ language.code }})
+                    </option>
+                {% endfor %}
+            </select>
+        </form>
```
danloveg, line 19: Two things:

1. Can you increase the contrast of this form? There white on light grey is too hard to see

Image

2. Can you move the language switcher to the top right of the screen in the banner?

danloveg (changes requested):

app/recordtransfer/templates/recordtransfer/banner.html
Eashana0104: I wasn’t able to replicate the contrast issue on my device in either Light or Dark mode—it might be browser-specific. Styling for <select> dropdowns is often handled by the browser itself, and many browsers tend to ignore custom styles for native select elements. That said, I’ve tried adding some styling to the select and option elements—let me know if it makes a difference on your end. 
Its done in https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/904/commits/c27f7e9a2b28cdd7b4aca92d5a82d10d388de4a1

Eashana0104: I checked on Safari as well it looks fine, which browser do you use again? Was it Microsoft?

danloveg: That looks good now!

Eashana0104: For sure! We should be good to go for this then

app/recordtransfer/templates/recordtransfer/banner.html
```diff
@@ -1,6 +1,20 @@
-<div class="banner-container">
-    <div>
+{% load i18n %}
+<div class="banner-container flex items-center justify-between gap-6 w-full">
+    <div class="flex items-center">
```
danloveg, line 3: This phone line gets squished when the screen is narrow. Let's hide it on `sm` screens. I was worried about doing this but if you look at https://nctr.ca they do the same thing so I'm OK with doing that here.

```suggestion
    <div class="hidden sm:flex items-center">
```

app/recordtransfer/templates/recordtransfer/banner.html
```diff
@@ -1,6 +1,20 @@
...
         <span class="banner-text">The National Residential School Crisis Line&nbsp;</span>
         <a class="banner-link non-nav-link" href="tel:1-866-925-4419">1-866-925-4419</a>
     </div>
+    <form action="{% url 'set_language' %}" method="post" class="ml-4">
```
danloveg, line 7: A side effect of hiding the phone line is that the language picker gets moved to the left when the phone line disappears. You can fix this with this class:

```suggestion
    <form action="{% url 'set_language' %}" method="post" class="ml-auto">
```

danloveg (changes requested):
Besides the two suggestions here I have one more suggestion, can you change the banner-container's padding to be `padding: 12px 20px` instead of `padding: 12px 40px`?

danloveg (changes requested):
Sorry, this is my bad for not looking at the test case until now! There are a few things that need to be changed in this test (see below).

Once that's done this pull request will be ready to go.

Eashana0104:
> Sorry, this is my bad for not looking at the test case until now! There are a few things that need to be changed in this test (see below).
> 
> Once that's done this pull request will be ready to go.

These make sense, I ll try to use unit test asserts instead, code looks better this way!

danloveg (approved):

PR #912, opened 2025-07-31
Fix issue with empty list passed to `max()`

danloveg:
This PR fixes an issue with the `get_js_translation_version` function when there are no ~~`djangojs.po`~~ `djangojs.mo` files.


Eashana0104 (changes requested):
Just this typo perhaps, everything else looks good to me, I had similar code to fix the issue!

Eashana0104 (approved):
Everything looks good! I ran e2e and unit tests on this branch as well and everything passes!

PR #915, opened 2025-07-31
Add info to admin for finding files

### LINKED ISSUE #884: Add Uploaded file inline to Submission
I think it would be useful to be able to see individual uploaded files right from the Submission admin without having to go into the session object to find files.

danloveg:
Closes #884 

Unfortunately, since `PermUploadedFile` is not a foreign key of `Submission`, it's be very difficult to render an inline of these indirectly related models.

I've settled for a simpler solution: updating the `__str__` of the UploadSession to make it more human readable, and adding a help text that says to click the upload session's name to view files uploaded in that submission.

Before:

upload_session_before

After:

upload_session_after

It looks more friendly now, in my opinion.

### DIFF
--- a/app/recordtransfer/forms/admin_forms.py
+++ b/app/recordtransfer/forms/admin_forms.py
@@ -106,6 +106,10 @@ class SubmissionModelForm(RecordTransferModelForm):
             "uuid",
         )
 
+        help_texts: ClassVar[dict] = {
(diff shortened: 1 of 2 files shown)


yenaing-oo (approved):
Looks good - the `UploadSession` string representation is much more readable now. It's too bad there's no built in way to add inlines for indirect relationships between models.

Issue #829, opened 2025-07-22
Verify strings in code can be translated


danloveg:
Translate a few strings found in the Python code, and ensure that when the language is switched to Hindi, that the strings are indeed translated from English. For example, maybe some of the field names in the submission form or the help texts can be translated.

To compile message files, see:

https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#compiling-message-files

danloveg:
Closed by #918

PR #918, opened 2025-08-01
829 verify strings in code can be translated

Eashana0104:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/829


danloveg (approved):
Looks great. That also makes sense about `gettext_lazy`.

PR #919, opened 2025-08-01
Upgrade python to 3.10, 3.11

### LINKED ISSUE #811: Upgrade Python to 3.10, 3.11, drop support for 3.9
Python 3.9 will be end-of-life in October 2025. We should bump the Python version we're using to 3.11, and support 3.10 as well, dropping support for 3.9.

Changes that need to be made:

- [x] Change Python version in `.python-version` to 3.11
- [x] Change Python version in `Containerfile` to 3.11
- [x] Change Python version matrix in [django-tests.yml](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/master/.github/workflows/django-tests.yml) to 3.10, 3.11

Changes to typing:

- [ ] Change `Union` typing to [type union operator](https://docs.python.org/3/whatsnew/3.10.html#pep-604-new-type-union-operator)
- [ ] Change `Optional[type]` to `type | None`
- [ ] Replace `List[type]` with `list[type]`
- [ ] Replace `Dict[type]` with `dict[type]`

For example, this type:

```python
Union[str, List[str]]
```

Will be changed to:

```python
str | list[str]
```

danloveg:
Closes #811 

I've upgraded the supported python version from 3.9 + 3.10 to 3.10 + 3.11.

I thought I'd fix the typings app-wide but that's a lot more work than I realized it would be. I refactored `recordtransfer.admin` and that was already pretty complex. I think I'll just add `admin.py` to this pull request as a good example of how the typings *should* be written going forward.


danloveg:
@yenaing-oo I forgot to mention since it's been a little while, but I updated your reset to default functionality in this PR to fix the type errors with that functionality.

yenaing-oo (changes requested):
I like the refactors made to `admin.py` -  including `formatted_upload_size` and `uploaded_file_url` as a mixin in the relevant admin classes looks much cleaner. The changes you made to the site setting reset functionality also makes sense to me. I like how you found the appropriate method (`response_change`) to override to perform the redirect and fix the type error. 

Python versions are updated in all the required places, and all the workflows run correctly.

Aside from the one suggestion I made in `admin.py`, everything looks good!

yenaing-oo:
> @yenaing-oo I forgot to mention since it's been a little while, but I updated your reset setting to default functionality in this PR to fix the type errors with that function.

I see, yes. I recall struggling to fix that type error, but I'm glad you found the right method to override to return the desired redirect response.

yenaing-oo (approved):

PR #920, opened 2025-08-01
Set up django debug toolbar

yenaing-oo:
- Reorganized middleware according to doc recommendations: https://docs.djangoproject.com/en/4.2/ref/middleware/
- Set up django debug toolbar so that it's only used for the development environment, without affecting tests

Note: The placement of django debug toolbar's middleware is intentional.


danloveg (approved):
This is really cool! I can see the query and CPU tabs being really useful in the future.

I see that it doesn't show up in prod as well.

Issue #879, opened 2025-07-28
Fix questionable gettext calls


danloveg:
There are a few `gettext` calls that could be updated to both give translators more context and to be more consistent in how translations are created app-wide.

# No. 1

```
#: caais/forms.py:100
msgid ""
". Use the date format YYYY-MM-DD for a single date or YYYY-MM-DD - YYYY-MM-"
"DD for a date range"
msgstr ""
```

Should be called like this instead:

```python
gettext("%(help_text)s. Use the date format ....") % {"help_text": original_help_text}
```

# No. 2

```
#: recordtransfer/caais.py:539
msgid "{TermClass._meta.verbose_name} was noted as {other_term!r}"
msgstr ""
```

Should be called like this instead:

```python
pgettext(
    context="class_name is the name of a Python class",
    message="%(class_name)s was noted as %(term_name)s",
) % {
        "class_name": TermClass._meta.verbose_name,
        "term_name": repr(other_term),
    }
)
```

`pgettext` allows you to add context to translations. [See here](https://docs.djangoproject.com/en/4.2/ref/utils/#django.utils.translation.pgettext).

# No. 3

```
#: recordtransfer/models.py:863
#, python-brace-format
msgid "{0}, totalling {1}"
msgstr ""
```

Needs to be refactored. See `get_human_readable_file_count`

# No. 4

```
#: recordtransfer/templates/includes/frequently_asked_questions.html:24
msgid ""
"\n"
"                        — You can organize related records by theme, time "
"period, or project, even if\n"
"                        they're submitted at different times.\n"
"                        "
msgstr ""
```

The leading dash `—` needs to be excluded from **all** translatable strings in the FAQ page.

# No. 5

```
#: recordtransfer/views/media.py:182
#, python-brace-format
msgid "Malware was detected in the file \"{_file.name}\""
msgstr ""
```

Should be called like:

```python
gettext('Malware was detected in the file "%(name)s"') % {"name": _file.name}
```

PR #925, opened 2025-08-05
879 fix questionable gettext calls

Eashana0104:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/879

### DIFF
--- a/app/locale/en/LC_MESSAGES/django.po
+++ b/app/locale/en/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2025-08-05 15:32-0500\n"
+"POT-Creation-Date: 2025-08-05 15:35-0500\n"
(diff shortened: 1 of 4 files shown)


Eashana0104:
Apart from the suggested changes to the gettext() calls, I’ve made a couple of additional improvements:

In point No.2, I added context not only for term_value but also for term_name, as I felt both placeholders would benefit from clearer context for translators.

In point No.3, I updated the string to use proper named placeholders for count and size, ensuring consistency and making the intent clearer for translators.

Additionally, I found several other problematic gettext() calls similar to No.3, and I’ve successfully fixed them to follow best practices for translatable strings in https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/925/commits/51fee3677dad437055bdf86a92d4a4572c0feaad and https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/925/commits/01328fd5089131a75323ef7f39cf6dca6d8b7a53

Let me know if there’s anything you’d like me to revise further!

danloveg (approved):
The `get_human_readable_file_count` needs to be refactored to make it easier to translate, but I think that falls outside the scope of this issue. Everything else looks great.

PR #928, opened 2025-08-06
Ensure emails are sent in user's language

### LINKED ISSUE #880: Ensure emails are sent in user's language
We should test and make sure that emails sent to a user are sent in their preferred language. I don't know if this would require a preferred language option in the User model or if there is some other way to specify this.

yenaing-oo:
Closes #880 

Change notes:
- Added ability to change language preference from the Profile page, which uses the exact same language switching functionality implemented in https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/904
- Language preference is saved to the `User` model now, which allows the language preference to be persistent - this means regardless of the language set for the page _before_ login, the user's preferred language will be used upon login
- Emails will now be sent in the user's preferred language
- Added additional E2E tests for language switching
- Refactored `recordtransfer/commands/send_test_email.py` to allow specifying language for the test email 
- Made first and last name fields for the `UserAccountInfoForm` optional


app/recordtransfer/apps.py
```diff
@@ -9,3 +9,11 @@ class RecordTransferConfig(AppConfig):
...
 
     name = "recordtransfer"
+
+    def ready(self) -> None:
+        """Import signal handlers when Django starts."""
+        from django.contrib.auth.signals import user_logged_in
+
+        from recordtransfer.signals import add_logged_in_flag
+
+        user_logged_in.connect(add_logged_in_flag)
```
danloveg, line 19: The logic here is hard to follow, but I understand what's going on. You connect to this signal, and since you don't have access to the response object, you set a flag here, and access it later in the middleware when you check if that flag is set, and add a cookie to the response.

I wonder if you could implement this logic in `recordtransfer.views.account.Login.form_valid` instead - when a successful login form is submitted, check if the user's language needs to be updated in the cookie. If you can do that, then:

- You can remove `SetLanguageOnLoginMiddleware`
- You can remove `signals.py`
- You can remove this code here.

If you *can't* implement this logic in the view, then this method you wrote does work, but my main complaint is that this code is hidden away in `apps.py` and `signals.py`. If you need to use this method, could you move the `add_logged_in_flag` function from `signals.py` to `auth_signals.py`? You *should* be able to do this as well, and skip running this in `apps.py`:

```
@receiver(user_logged_in)
def add_logged_in_flag(...):
    ...
```

danloveg (changes requested):

app/recordtransfer/apps.py
yenaing-oo: > The logic here is hard to follow, but I understand what's going on. You connect to this signal, and since you don't have access to the response object, you set a flag here, and access it later in the middleware when you check if that flag is set, and add a cookie to the response.
> 
> I wonder if you could implement this logic in `recordtransfer.views.account.Login.form_valid` instead - when a successful login form is submitted, check if the user's language needs to be updated in the cookie. If you can do that, then:
> 
> * You can remove `SetLanguageOnLoginMiddleware`
> * You can remove `signals.py`
> * You can remove this code here.
> 
> If you _can't_ implement this logic in the view, then this method you wrote does work, but my main complaint is that this code is hidden away in `apps.py` and `signals.py`. If you need to use this method, could you move the `add_logged_in_flag` function from `signals.py` to `auth_signals.py`? You _should_ be able to do this as well, and skip running this in `apps.py`:
> 
> ```
> @receiver(user_logged_in)
> def add_logged_in_flag(...):
>     ...
> ```

My bad, I should've included an explanation of how the signal handler was working with the middleware. Yes, that's exactly how it works. The only reason I didn't implement this logic in the `Login` view is because there is one other area where the cookie would have to be updated as well - in `recordtransfer.views.account.ActivateAccount.get`.

I think setting the language cookie in both these places should be fine instead of the middleware approach, since regular login and account activation are the only two places that would need this. I agree the signal + middleware approach is a bit obscure.

danloveg: No problem. I think you could create one function that adds the required cookie to the response, and wrap the response objects in both `ActivateAccount.get` and `LoginView.form_valid`

yenaing-oo: This was exactly what I had in mind.

Addressed by [d78f01d](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/928/commits/d78f01dc9618eef3d5f7cee4aa96eb375ec808c3) and https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/928/commits/f18aab3e738a86c83ce5fd2a9f27ceb036885929

app/recordtransfer/emails.py
```diff
@@ -224,54 +272,54 @@ def send_user_in_progress_submission_expiring(in_progress: InProgressSubmission)
...
-        "Found %d Users(s) to send email to: %s", len(recipients_list), str(recipients_list)
-    )
-    return recipients_list
+    language_groups = {}
+    for user in users:
+        lang_key = user.language or translation.get_language()
+        if lang_key not in language_groups:
+            language_groups[lang_key] = []
+        language_groups[lang_key].append(user.email)
+    return language_groups
```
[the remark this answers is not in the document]
yenaing-oo: > There's a better data structure for this - [defaultdict](https://docs.python.org/3/library/collections.html#collections.defaultdict):

Nice, I didn't know this was a thing

app/recordtransfer/emails.py
```diff
@@ -224,54 +272,54 @@ def send_user_in_progress_submission_expiring(in_progress: InProgressSubmission)
...
-    _send_mail_with_logs(
-        recipients=[to_email],
+    if not to_email:
+        LOGGER.error("Missing email in context for password reset email.")
+        return
+
+    _send_mail(
+        recipient=to_email,
         from_email=_get_do_not_reply_email_address(),
         subject=subject,
```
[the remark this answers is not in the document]
yenaing-oo: This is the first clear-cut example I've encountered where lazy translation makes sense to me! Deferring the translation till we actually need it in the `translation.override()` block is perfect for this use case. 

Addressed by: https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/928/commits/9557f1b087b526ff27d39a642156c74700213a6b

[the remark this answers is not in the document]
yenaing-oo: Also, another interesting thing I discovered is that `makemessages` won't pick up translations if the translation function is not the original function name or `_`. For e.g.,
```python
from django.utils.translation import gettext_lazy as _  # This works fine
from django.utils.translation import gettext_lazy as _lazy  # Not fine
```

danloveg (approved):
Looks great!

Issue #846, opened 2025-07-24
Form Data Loss on Validation Error


Eashana0104:
Currently, when a user submits the form and there’s an error in one or more fields, all previously entered data is cleared—even for fields that were correctly filled out. This creates a frustrating experience, as users are forced to re-enter information unnecessarily. Ideally, the values in valid fields should be retained, and even the incorrect fields should preserve the user’s input so they can easily identify and correct the errors without starting over.

<img width="1470" height="956" alt="Image" src="https://github.com/user-attachments/assets/8d01367c-a48d-4572-821a-2f7b01da052e" />

PR #929, opened 2025-08-06
Fix fields getting cleared after invalid form fields

Eashana0104:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/846


danloveg (approved):
This makes sense, this is much simpler than I thought it'd be!

Issue #887, opened 2025-07-29
Display remaining file count in Step 8 of submission form


Eashana0104:
In Step 8 of the submission form, users are currently shown how much more data they can upload (in MB). However, it would also be useful to inform users how many additional files they can approve at that point, considering the maximum upload limit is 40 files.

Displaying both the remaining data size and the remaining file count would give users a clearer picture of their remaining capacity and help avoid confusion during the upload process.

Somewhat here would be beneficial!

<img width="599" height="150" alt="Image" src="https://github.com/user-attachments/assets/b733d2ce-a93a-4369-937e-4e0ae7ce7c3e" />

PR #930, opened 2025-08-06
887 display remaining file count in step 8 of submission form

Eashana0104:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/887

Implemented Total Count and Remaining Count of files in the same way as Total Size and Remaining Size. Updated the JavaScript accordingly to keep everything in sync and consistent.

Checked edge cases as well!

### DIFF
--- a/app/recordtransfer/static/recordtransfer/js/submission_form/utils.js
+++ b/app/recordtransfer/static/recordtransfer/js/submission_form/utils.js
@@ -167,10 +167,21 @@ export const updateCapacityDisplay = (totalSizeBytes, maxUploadSize, totalCount,
     const totalCountElement = document.getElementById("files-count");
     const remainingCountElement = document.getElementById("remaining-files");
 
-    const updateElement = (element, value, isError, isInteger = false) => {
+    const updateElement = (
+        element,
+        value,
+        isError,
+        isInteger = false,
+        isRemainingField = false
+    ) => {
         if (element) {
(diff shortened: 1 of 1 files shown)


danloveg (changes requested):
It works good! I do have a feature request to make this a bit better.

Can you change the colour or at least make it more apparent that there is no more space to upload files when you reach the max? Not sure if making it red, or adding an exclamation mark would make sense, but even when it says "0 files" it doesn't look like a warning at all.

This should happen with the upload size as well.

Image

danloveg:
> I do think changing the color to red improves the visual clarity and draws attention effectively. However, a dding an exclamation mark felt a bit too strong or urgent for the context—removing it keeps the tone more balanced.
> 
> @danloveg let me know what do you think

The red without the exclamation mark looks the best, we should go with that option

Eashana0104:
> > I do think changing the color to red improves the visual clarity and draws attention effectively. However, a dding an exclamation mark felt a bit too strong or urgent for the context—removing it keeps the tone more balanced.
> > @danloveg let me know what do you think
> 
> The red without the exclamation mark looks the best, we should go with that option

This is resolved in https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/930/commits/1b8b40d4cfe8228ff61bdf2de3e791be8ef1633d

Issue #905, opened 2025-07-31
Home page styles break on short screens


danloveg:
On a short screen, the title text is allowed to go above the header:

![Image](https://github.com/user-attachments/assets/bda7b954-7db2-4f14-8bb9-40e84460e141)

The expected behaviour is that the title never goes higher than the "NCTR Record Transfer" text.

PR #936, opened 2025-08-08
905 home page styles break on short screens

Eashana0104:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/905

Change notes-

- Moved the navbar outside of the #main-container to prevent layout overlap.
- Applied DaisyUI styling enhancements to home.html.
- Removed unused CSS from home.css as it was no longer needed due to DaisyUI.
- Added smooth transitions to improve responsiveness when switching between different screen sizes.

### DIFF
--- a/app/recordtransfer/static/recordtransfer/css/base/base.css
+++ b/app/recordtransfer/static/recordtransfer/css/base/base.css
@@ -33,8 +33,8 @@ body {
 .main-header-content,
 .main-content {
     width: 80%;
-    margin: 0 auto;
-
+    margin-left: auto;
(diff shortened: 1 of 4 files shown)


Eashana0104:
Added some DaisyUI components, which helped fix the bug. While home.html still primarily relies on home.css, we could consider refactoring it to use DaisyUI for consistency. That said, leaving it as-is shouldn’t cause any issues in the future. Let me know what you think @danloveg

danloveg (approved):
I love the animation on the text!

Issue #894, opened 2025-07-30
Hardcoded `https` Links in Emails Not Supported in Development


yenaing-oo:
Links in emails are inaccessible in development as HTTPS is not enabled there. We should use a context variable or a Django setting to control the protocol in email links.

PR #938, opened 2025-08-08
Fix user not being logged in after activation, and make base URL in emails configurable 

### LINKED ISSUE #927: User not getting activated upon Sign Up
Description:
After signing up, the user is automatically logged in. However, if the user logs out and tries to log back in using the same credentials, the login fails.

From debugging, we observed that:

The signup logic correctly reaches the conditional block meant to activate the user.

Despite this, the user does not get activated.

This issue started occurring after the integration of django-axes.

It appears there might be an interaction or conflict with django-axes that is preventing the user activation from completing successfully.

Steps to Reproduce:

- Sign up with a new user.
- User is logged in successfully after registration.
- Log out.
- Try logging in again with the same credentials.
- Login fails — likely because the user is still marked as inactive.

Expected Behavior:
The user should be properly activated during signup and able to log in again using the same credentials.

Notes:

- Confirmed that the relevant if condition for user activation is being executed.
- No errors appear in the logs or UI.
- Suspected interference from django-axes.

yenaing-oo:
Closes #927  and #894

### DIFF
--- a/app/recordtransfer/emails.py
+++ b/app/recordtransfer/emails.py
@@ -6,6 +6,7 @@ import smtplib
 from typing import List, Optional
 
 import django_rq
+from django.conf import settings
 from django.contrib.sites.models import Site
 from django.core.mail import send_mail
 from django.template.loader import render_to_string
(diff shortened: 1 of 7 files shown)


app/recordtransfer/emails.py
```diff
@@ -20,6 +21,22 @@
 
 LOGGER = logging.getLogger("rq.worker")
```
danloveg, line 22: Since you're here in this file, can you change the string to `__name__`?

danloveg (approved):

PR #950, opened 2025-08-11
Sanitize and/or reject malicious file names in accept_file

### LINKED ISSUE #948: Issue: sanitize malicious file names in accept_file
**Description of the Bug**

The `accept_file` function will accept maliciously constructed file names/

**Steps to Reproduce**

N/A

**Expected Behaviour**

`accept_file` should reject files with maliciously constructed names.

**Screenshots**

N/A

**App Configuration**

Both configurations

**Additional Context**

I will create a PR with a failing test to show examples of names we want to sanitize or reject.

danloveg:
Closes #948 

Adds validation for absolute paths, path traversal, control characters, and Windows reserved file names.

I added Windows-specific reserved file names because we can't guarantee that an uploaded file stays in the app's storage. If someone downloads a Bag to their Windows machine, we don't want it to be possible for there to be a reserved Windows filename in that Bag.


yenaing-oo (dismissed):
I wonder if it would be useful to also check the MIME type of the file and reject executables. For e.g., a file could be named `application.pdf` but the contents could actually be `.exe` with the extension changed. If its contents are malicious, I assume ClamAV would catch it regardless.

Aside from this, everything looks good to me.

PR #954, opened 2025-08-11
825 set up translations

Eashana0104:

PR #956, opened 2025-08-11
Update list display columns, add Status filter

### LINKED ISSUE #866: Update Metadata list_display
The Metadata `list_display` could be improved by:

- Removing Acquisition method
- Adding a column for source name
- Adding a column for status
- Adding a column for last updated
- Adding a filter for statuses

The list display for Metadata should look like this:

Image

Here are the functions that can be used to give the source name column and the last updated column:

```python
from caais.db import GroupConcat

# ...

class MetadataAdmin(admin.ModelAdmin):

    # ....

    def source_name(self, obj) -> Optional[str]:
        return obj.source_of_materials.aggregate(
            names_joined=GroupConcat("source_name", separator=", ")
        )["names_joined"]
    
    def last_changed(self, obj) -> Optional[datetime]:
        most_recent = obj.dates_of_creation_or_revision.order_by(
            "-creation_or_revision_date"
        ).first()
        if most_recent:
            return most_recent.creation_or_revision_date
        return None
```

yenaing-oo:
Closes #866

### DIFF
--- a/app/caais/admin.py
+++ b/app/caais/admin.py
@@ -1,5 +1,6 @@
 """CAAIS metadata administrator."""
 
+from datetime import datetime
 from typing import Any, Mapping, Optional
 
 from django.contrib import admin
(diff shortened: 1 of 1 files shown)


danloveg (approved):
Looks great.

I'm going to merge this after merging #954 in case there are any conflicts - I'd rather we handle the conflicts on this branch than that one.

Issue #853, opened 2025-07-24
Back button on form should go to previous step


danloveg:
This is an opinion, but I think that clicking back in the browser on the submission form should go to the previous step, rather than exit the form completely.

I don't know how common it is to hijack back button behaviour, so I don't know if this is good practice or not. The form might feel more natural if a user can click back to go to the previous step.

Eashana0104:
@danloveg I did some research on whether it is ideal to hijack back button behavior. Usually, its not perfect but for our case ie, Multi-step wizards, It is usually adopted and is a valid thing to do!

danloveg:
@yenaing-oo we're stumped on this one. The back button and forward buttons are interesting in that they do not work exactly like the previous and next buttons - they take you to the previous page. So if you load an in-progress submission, and press back, you don't go back in the form, you go back to the profile page.

A simpler solution might be to open the unsaved changes modal when someone presses back to exit the form and don't try to override the window's history.

The main issue seems to be that the form's URL is `/submission/` for the entirety of the form, so typical back button behaviour doesn't work to bring you to a previous step (since it takes you back to the previous URL).

Ash made a partially working PR #960 but I took a look and couldn't fix it in a way that made the form feel natural. I looked into using a [NamedUrlWizardView](https://django-formtools.readthedocs.io/en/latest/wizard.html#usage-of-namedurlwizardview) so that each step would have a different URL, but the URL doesn't change throughout the form because HTMX only reloads the form when you press Previous/Next.

I'm tempted to close this issue and open one for just opening the unsaved changes modal when you press back, because the behaviour we're going for here is really complicated to implement. If you have any ideas how to make this work, let us know.

PR #960, opened 2025-08-11
853 back button on form should go to previous step

Eashana0104:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/853


danloveg:
We're going to simplify the requirements for this issue. Instead of using the browser buttons for navigating through the form, we're going to instead show the unsaved changes modal when pressing back instead of the default browser confirmation. This should make it clear to users that pressing back will clear the form.

PR #967, opened 2025-08-13
Migrate password change to own page, send emails for password reset and change

### LINKED ISSUE #962: Enhancement: Perform password change on separate form
Currently, if a user would like to change their password while they are logged in, they would have to do so on the Account tab of their Profile page. This form deals with too many different fields - the password fields should be part of their own form with it's own separate validation. This will make it easier to test and maintain. Also, the naming of this action should be "Change password".

I'd like to see password change being done on its own page or using a modal, possibly using Django's built-in [PasswordChangeForm](https://github.com/django/django/blob/4926591343756d9fdd5d57680e021364ff71288f/django/contrib/auth/forms.py#L531). `recordtransfer.forms.user_forms.UserAccountInfoForm` will have to be refactored to remove password related fields/validation.

yenaing-oo:
Closes #962 

Change notes:
- Migrated the password change functionality from the Profile page to a separate page that uses Django's built-in [PaswordChangeView](https://github.com/django/django/blob/stable/5.2.x/django/contrib/auth/views.py#L354)
- Automated email notifications for successful password change and password reset

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 *.sh text=auto eol=lf
 *.conf text=auto eol=lf
+*.conf.template text=auto eol=lf
--- a/app/caais/migrations/0006_extent_types.py
+++ b/app/caais/migrations/0006_extent_types.py
@@ -4,22 +4,26 @@ from caais.models import ExtentType
 
 
 def populate_extents(apps, schema_editor):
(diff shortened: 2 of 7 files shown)


yenaing-oo:
TODO: Send email upon successfully changing/setting password through password change or password reset. Currently, an email is only being sent for a password reset request

PR #970, opened 2025-08-14
Skip running `npm run build` when only discovering tests

### LINKED ISSUE #969: Issue: Webpack runs during pytest discovery
**Description of the Bug**

Collecting Python tests in VS Code runs webpack build which slows down test discovery.

**Steps to Reproduce**

1. Open this repo in VSCode
2. In the test tab, click the **Refresh Tests** button to re-discover tests
3. Open the output tab in VSCode and switch to the Python tab
4. You will see the webpack development build being run after pytest finds files

**Expected Behaviour**

Webpack should not need to run while discovering tests, since those resources aren't needed until E2E tests have to run.

**Screenshots**

The command used to discover tests:

Image

The Webpack build being run later on:

Image

**App Configuration**

N/A - tests only.

**Additional Context**

It looks like pytest is being called with `--collect-only`. I believe a simple solution would be to *not* build the static assets with Webpack when that argument is used.

danloveg:
Closes #969 

Checks if tests are only being collected (i.e., not run) and skips building static assets with Webpack if that's the case.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -14,9 +14,7 @@ def pytest_runtestloop(session: pytest.Session) -> None:
     run.
     """
     selected_items = session.items
-    if not session.config.option.collectonly and any(
-        "/e2e/" in item.nodeid for item in selected_items
-    ):
+    if any("/e2e/" in item.nodeid for item in selected_items):
         try:
             npm_cmd = shutil.which("npm") or "npm"
             subprocess.run(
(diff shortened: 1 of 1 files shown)


yenaing-oo (approved):
Test discovery is so much faster now (10x)! No wonder it was slow before - I had always assumed it was VSCode + Python that was slow whenever I hit "Refresh Tests". Great find!

PR #1000, opened 2025-08-20
Switch to `pnpm`

### LINKED ISSUE #981: Enhancement: Switch to pnpm

In an effort to make the container build as fast as possible, we might consider using [pnpm](https://pnpm.io/) which is supposedly 2x faster than npm. We've had great performance results switching from Poetry to uv for Python, this could be another good option for increased build speed.

- Install dependencies in container with `pnpm`
- Use a cache mount for `pnpm` like we've done with `uv`
- Remove `package-lock.json` in favour of pnpm lock file
- Update dev docs for installing pnpm. I think you can just do this: `npm install -g pnpm@latest-10`

**Describe alternatives you've considered**

N/A

**Additional context**

N/A

yenaing-oo:
Closes #981

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -16,14 +16,14 @@ RUN apt-get update && \
     libmagic1 && \
     rm -rf /var/lib/apt/lists/*
 
+# Install pnpm globally
+RUN npm install -g pnpm@latest-10
+
 # Copy uv-related files, and install Python dependencies
 # Uses a persistent cache mount for uv (see https://docs.astral.sh/uv/guides/integration/docker/#caching)
 COPY pyproject.toml uv.lock ${PROJ_DIR}
 RUN --mount=type=cache,target=/root/.cache/uv \
     uv sync
-
-# Install pnpm globally
-RUN npm install -g pnpm@latest-10
 
 # Install Node.js dependencies with pnpm
 # Uses a persistent cache mount for pnpm (see https://pnpm.io/docker#minimizing-docker-image-size-and-build-time)


yenaing-oo:
@danloveg 
Would you be able to update the Windows-specific instructions in `DEVELOPERS.md` to install `pnpm`? I don't have access to WSL so I can't verify whether certain commands work.

danloveg:
> @danloveg Would you be able to update the Windows-specific instructions in `DEVELOPERS.md` to install `pnpm`? I don't have access to WSL so I can't verify whether certain commands work.

Just tested in, the same command works, you can just copy the instructions you've made for Linux/MacOS to the WSL section.

Screenshot 2025-08-22 144454

Containerfile
```diff
@@ -20,9 +22,14 @@ COPY pyproject.toml uv.lock ${PROJ_DIR}
 RUN --mount=type=cache,target=/root/.cache/uv \
     uv sync
 
-# Install Node.js dependencies
-COPY package*.json ${PROJ_DIR}
-RUN npm install --no-color
+# Install pnpm globally
+RUN npm install -g pnpm@latest-10
```
danloveg, line 26: This should be moved up to the same RUN command where gettext and libmagic1 are installed. The reason being, this is not a dependency that needs to be re-installed really ever, but the ordering of this Containerfile will make it so that installing `pnpm` is dependent on any time `pyproject.toml` or `uv.lock` change (which might be semi-often).

For the best caching performance I'll recommend you move this upwards in the Containerfile.

Containerfile
```diff
@@ -20,9 +22,14 @@ COPY pyproject.toml uv.lock ${PROJ_DIR}
...
+
+# Install Node.js dependencies with pnpm
+# Uses a persistent cache mount for pnpm (see https://pnpm.io/docker#minimizing-docker-image-size-and-build-time)
+COPY package.json pnpm-lock.yaml ${PROJ_DIR}
+RUN --mount=type=cache,id=pnpm,target=${PNPM_HOME}/store \
+    pnpm install --frozen-lockfile
```
danloveg, line 32: Can you make this:

```suggestion
    pnpm install --frozen-lockfile --prod
```

This will require moving `webpack` out of the dev dependencies and into the main dependencies which is a change I've been wanting to make for a little while.

I don't *think* we need ESLint in the container. I only use ESLint locally, but if you think there's a use case for linting in the container, maybe we can install all the dev dependencies, too.

But still, I think Webpack needs to be a prod dependency.

danloveg (changes requested):
Looking good! I just had a few comments about the Containerfile but other than that everything looks good to me.

Containerfile
danloveg: I think that's a matter of preference. I would personally put it in the same RUN command but if you think the file would be easier to read if it was in a separate command then go for it

yenaing-oo: I see, I've made that change here - https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/1000/commits/218740614755765eed409b334ccc4d363b709157

Issue #996, opened 2025-08-20
Issue: Tooltips on profile tables are hidden below header


danloveg:
**Description of the Bug**

The tooltips on the profile page tables get hidden under the table's header.

**Steps to Reproduce**

Provide steps to reproduce the bug as described above (i.e., how did you find this bug?):

1. Go to `/user/profile/`
2. Hover over any of the items in the Actions column on the first row of the table.
3. The full tooltip won't be visible.

**Expected Behaviour**

The tooltip should appear over top of the table header.

**Screenshots**

![Image](https://github.com/user-attachments/assets/25c20a28-e573-4d6e-b0fc-ab8a1b20e3b2)

**App Configuration**

Does this issue apply to the development configuration (`compose.dev.yml`), the production configuration (`compose.prod.yml`), or both?

Both

**Additional Context**

N/A

danloveg:
> [@danloveg](https://github.com/danloveg) I explored a few approaches to resolve this. The issue occurs because the tooltips have a lower z-index than the table header. As a result, when tooltips are shown above an element, they get hidden behind the sticky header.
> 
> If I simply increase the z-index of the tooltips, it creates another issue: when the table height is smaller and the user scrolls, those tooltips remain visible even after the related row has scrolled out of view.
> 
> I’m not entirely sure what the best solution is here. One option I can think of is to remove the sticky behavior from the table header, though that may not be ideal. It would solve our problem currently but It would not show user table headers while scrolling. This scrolling issue would only be the case if user purposely tries to increase the PAGINATE_BY setting. Let me know what your thoughts are on this!

That all makes sense to me. In that case, I have a super simple alternate solution that preserves all the functionality of these tables. You could just add [tooltip-left](https://daisyui.com/components/tooltip/?lang=en#left) to the tooltips.

![Image](https://github.com/user-attachments/assets/58d8200b-cf84-4374-9a1e-37bb757a7d48)

PR #1001, opened 2025-08-20
Fix Tooltips hidding behind headers

Eashana0104:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/996


danloveg (approved):
That works, nice and simple!

PR #1004, opened 2025-08-20
Fix broken uploaded file removal

### LINKED ISSUE #1003: Issue: Removal of uploaded files is broken
**Description of the Bug**

Apparently, Django treats URLs ending with and without a trailing slash as distinct. By default, [CommonMiddleware](https://docs.djangoproject.com/en/5.2/ref/middleware/#django.middleware.common.CommonMiddleware) adds a slash to the end of a URL if Django receives a request to a URL it doesn't recognize, to see if it will match any of the registered URLs and if so, redirect to that URL. This is because the [APPEND_SLASH](https://docs.djangoproject.com/en/5.2/ref/settings/#std-setting-APPEND_SLASH) setting is True by default. 

This fail-safe works fine for GET requests, but does not work for DELETE and POST requests as the follow-up request after a redirect is a GET that does not carry over the data from the DELETE/POST initial request.

**Steps to Reproduce**

1. Upload a file on the Upload Files step of the Submission Form
2. Try removing the file
3. The file does not actually get removed. Save the form and resume it to confirm the file is still there.
4. See logs to view runtime error.

**Expected Behaviour**

This error should be found in the logs upon removal of an uploaded file:

```log
2025-08-20 16:46:23 ERROR 2025-08-20 16:46:23,695 log: Internal Server Error: /upload-session/40ViEa0vlHMtm4oYGphv69kNSjXEOVlC/files/13c-c000117-d0028-001.pdf
2025-08-20 16:46:23 Traceback (most recent call last):
2025-08-20 16:46:23   File "/opt/secure-record-transfer/.venv/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
2025-08-20 16:46:23     response = get_response(request)
2025-08-20 16:46:23                ^^^^^^^^^^^^^^^^^^^^^
2025-08-20 16:46:23   File "/opt/secure-record-transfer/.venv/lib/python3.11/site-packages/django/utils/deprecation.py", line 122, in __call__
2025-08-20 16:46:23     response = self.process_response(request, response)
2025-08-20 16:46:23                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-08-20 16:46:23   File "/opt/secure-record-transfer/.venv/lib/python3.11/site-packages/django/middleware/common.py", line 108, in process_response
2025-08-20 16:46:23     return self.response_redirect_class(self.get_full_path_with_slash(request))
2025-08-20 16:46:23                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-08-20 16:46:23   File "/opt/secure-record-transfer/.venv/lib/python3.11/site-packages/django/middleware/common.py", line 87, in get_full_path_with_slash
2025-08-20 16:46:23     raise RuntimeError(
2025-08-20 16:46:23 RuntimeError: You called this URL via DELETE, but the URL doesn't end in a slash and you have APPEND_SLASH set. Django can't redirect to the slash URL while maintaining DELETE data. Change your form to point to localhost:8000/upload-session/40ViEa0vlHMtm4oYGphv69kNSjXEOVlC/files/13c-c000117-d0028-001.pdf/ (note the trailing slash), or set APPEND_SLASH=False in your Django settings.
2025-08-20 16:46:23 ERROR 2025-08-20 16:46:23,753 basehttp: "DELETE /upload-session/40ViEa0vlHMtm4oYGphv69kNSjXEOVlC/files/13c-c000117-d0028-001.pdf HTTP/1.1" 500 13747
```

**Screenshots**

N/A

**App Configuration**

Both

**Additional Context**

Not sure how we did not catch this before. This regression does not appear to be due to the recent Django upgrade, since the mentioned behaviour of Django treating URLs with and without a trailing slash has existed for a long time.

yenaing-oo:
Closes #1003

### DIFF
--- a/app/recordtransfer/tests/e2e/test_submission_form_e2e.py
+++ b/app/recordtransfer/tests/e2e/test_submission_form_e2e.py
@@ -1139,7 +1139,7 @@ class SubmissionFormWizardTest(SeleniumLiveServerTestCase):
         )
 
     def test_removing_uploaded_file(self) -> None:
-        """Test that the user can remove an uploaded file from the Review step."""
+        """Test that removign an uploaded file from the Upload Files step functions correctly."""
         driver = self.driver
 
         # Complete the form till the Review step


yenaing-oo:
> @danloveg I'm currently looking into why my E2E test passes even without the backslash fix.

_**Saving the form and resuming**_ instead of _**going to the previous step and back to the Upload Files step**_ within my E2E test passes as well without the backslash fix. I'm guessing Django's test server is not as strict with the lack of backslash in the DELETE URL.

PR #1016, opened 2025-08-22
Set up django-admin-interface for admin theming

### LINKED ISSUE #1005: Issue: Heading background colour incorrect in Admin site
**Description of the Bug**

The section headings in the admin site, which used to be red, are now a light grey colour. There is not enough contrast between the headings and the text in the light theme now (see Screenshots below).

**Steps to Reproduce**

1. Go to `/admin/`
2. Make sure the light theme is selected.
3. Notice that the heading styles are no longer red.

**Expected Behaviour**

The headings should be red in the light theme. The headings are also supposed to be dark red in the dark theme of the site. Though for the dark theme, there is enough contrast on the headings so that bug isn't quite as bad.

**Screenshots**

Light theme:

Image

Dark theme:

Image

**App Configuration**

Both

**Additional Context**

Since we upgraded to Django 5.2, the admin templates and their CSS classes seem to have subtly changed.

I fixed the light theme in `app/recordtransfer/templates/admin/base_site.html` with these styles, but I'm wondering whether there's a more elegant way of fixing this issue.

```css
.module caption {
    background-color: var(--primary);
}

fieldset .inline-heading {
    background-color: var(--primary);
}

.inline-group h2 {
    background-color: var(--primary);
}
```

yenaing-oo:
Closes #1005

Sets up [django-admin-interface](https://github.com/fabiocaccamo/django-admin-interface#django-admin-interface) to easily customize a theme for the admin interface. Our own custom theme is stored as a fixture and loaded onto the database as part of a migration.


danloveg:
> @danloveg How do you feel about using [django-admin-interface](https://github.com/fabiocaccamo/django-admin-interface#django-admin-interface)?
> 
> With that, you can easily customize the desired theme without any CSS/HTML through the "Theme" admin. Once we've created a theme we like, we can export it as a JSON fixture, and load that fixture on startup.
> 
> One thing we'll have to fix is switching between a dark and a light theme. The default theme switcher button on the admin navbar does not seem compatible with [django-admin-interface](https://github.com/fabiocaccamo/django-admin-interface#django-admin-interface). The themes are being stored in the database, and the selected theme is queried with the required CSS for it dynamically being generated. We might be able to overide the theme switcher button to send a POST to use a different theme, or it might just be easier to disable that button entirely.
> 
> Feel free to play around with the Themes, which I've already set up on this branch. Curious to know what you think.

This looks like a useful app, I like this. I think disabling the button would make the most sense to me.

I see you're still actively working on the theme, but the colours are incorrect right now. I don't think this needs a re-design, we just need to make it look like how it used to.

yenaing-oo:
I've updated the styling so that it matches what it looked like before, except with two small differences:
- I can't get rid of this block of color below the navbar:
image

- The "Theme" admin only allows me to set one color for all the save buttons:
image
as opposed to this:
image

danloveg:
> > I have no concerns with the code changes you've made, but can you document this in the docs? I think this would fit within **admin_guide**.
> > Feel free to link out to any other documentation for the details, but I would like at least some reference in our docs for admins to know that it is possible to change the admin theme.
> 
> That's a good idea.
> 
> Since [django-admin-interface](https://github.com/fabiocaccamo/django-admin-interface#readme) doesn't provide any documentation on what each individual setting in the Theme admin controls, I'll be documenting them (to the best of my ability - it might not be comprehensive) using screenshots.

That sounds great, thanks

yenaing-oo:
The docs for managing themes are now ready for review. 

I want to note that I'm using the hardcoded NCTR logo and favicon (as we're currently already doing) instead of relying on `django-admin-interface` for those. The reason for this is because:

- Uploaded images are not included when exporting a `Theme` as a fixture. Only the file paths are included. This means when the admin site is loaded, unless those images are available under `MEDIA_URL` + `admin-interface/logo/` + filename, they won't be found (see [reference](https://github.com/fabiocaccamo/django-admin-interface/blob/210037641aeac2795560d46366cfe3b044d2e754/admin_interface/models.py#L65))
- Our app is specific to NCTR to begin with, so having a hardcoded logo and favicon should be fine. If another party wants to use this app with their branding, they can modify [here](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/6bcf29c13c060bab90cd8531dab505d9be65ac59/app/recordtransfer/templates/admin/base_site.html#L6) and [here](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/6bcf29c13c060bab90cd8531dab505d9be65ac59/app/recordtransfer/templates/admin/base_site.html#L88) in the code.
- We want to process all static assets with our webpack setup

docs/admin_guide/managing_admin_themes.rst
```diff
@@ -0,0 +1,370 @@
...
+4. Click **Save** to create the theme
+
+Modifying an Existing Theme
+---------------------------
+
+To modify an existing theme:
+
+1. Navigate to **Admin Interface** → **Themes** in the admin
+2. Click on the theme you want to modify from the list
+3. Make your desired changes to any of the theme settings (see :ref:`theme-settings` above for detailed descriptions)
```
danloveg, line 58: I think this should say below


```suggestion
3. Make your desired changes to any of the theme settings (see :ref:`theme-settings` below for detailed descriptions)
```

danloveg (changes requested):
Nice! Your reasoning for the favicon and the logo make sense.

I have one very tiny suggestion and then this looks good to go.

Issue #1013, opened 2025-08-21
Issue: Export submissions as CSV dont work on Submission Group table


Eashana0104:
**Description of the Bug**
Currently, the 'Export submissions as CSV' link in the Submission Group table leads to a 404 error page when the group has no submissions. Instead, it should display a user-friendly error message such as: 'This submission group has no submissions associated with it.'

**Steps to Reproduce**
Provide steps to reproduce the bug as described above (i.e., how did you find this bug?):

1. Go to the Profile page
2. Click on Submission group header
3. Create a group with no submissions associated to it
4. Click Download icon for Export submissions as CSV
5. See the 404 error

**Expected Behaviour**

Provide a clear and concise description of what you expected to happen (i.e., why is this a bug?).

This should have given an error like "This submission group has no submissions associated with it" instead of giving a 404 error

**Screenshots**

<img width="1167" height="604" alt="Image" src="https://github.com/user-attachments/assets/ee807f18-f87c-4b4e-a071-42ba9180855b" />

**App Configuration**

Does this issue apply to the development configuration (`compose.dev.yml`), the production configuration (`compose.prod.yml`), or both?

both

**Additional Context**

yenaing-oo:
Would it be better if that button was hidden entirely if the submission group has not submissions associated with it?

PR #1017, opened 2025-08-22
Add error when user tries to download CSV for empty group

Eashana0104:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/1013

### DIFF
--- a/app/recordtransfer/views/post_submission.py
+++ b/app/recordtransfer/views/post_submission.py
@@ -125,7 +125,7 @@ def submission_group_bulk_csv_export(request: HttpRequest, uuid: str) -> HttpRes
     if not related_submissions.exists():
         messages.error(
             request,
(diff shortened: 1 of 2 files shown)


app/recordtransfer/views/post_submission.py
```diff
@@ -122,7 +123,12 @@ def submission_group_bulk_csv_export(request: HttpRequest, uuid: str) -> HttpRes
         )
 
     if not related_submissions.exists():
-        raise Http404("No submissions found in this group")
+        messages.error(
+            request,
+            "This submission group has no submissions associated with it. Please add some submissions before attempting to export.",
```
danloveg, line 128: This message is a bit long, and needs translation. How about:

```suggestion
            _("This submission group has no submissions associated with it to export."),
```

danloveg (changes requested):
Looks good, just have one change request here.

PR #1019, opened 2025-08-25
Fix reserved identifier names allowed

### LINKED ISSUE #1018: Issue: Reserved Identifier types allowed in submission form
**Description of the Bug**

It is possible to submit an Identifier in the form with the reserved names "Accession Identifier" or "Accession Number". These are reserved because they are what is used to assign the Accession identifier field in the Metadata object.

**Steps to Reproduce**

1. Fill out the form, making sure to create an Identifier with the Identifier type "Accession Number" or "Accession Identifier"
2. Submit the form
3. Log in to the admin site
4. Go to the submitted Metadata object
5. You will see the Accession Identifier is already assigned!

**Expected Behaviour**

Donors should not be allowed to assign the accession identifier for any submission.

**Screenshots**

Filling out the form:

Image

Looking in the admin later on:

Image

**App Configuration**

Both.

**Additional Context**

N/A

danloveg:
Closes #1018 

Disallows Accession Identifier and Accession Number to be submitted in the form.

Screenshot 2025-08-25 135847

### DIFF
--- a/app/caais/admin.py
+++ b/app/caais/admin.py
@@ -9,6 +9,7 @@ from django.http import HttpRequest
 from django.template.response import TemplateResponse
 
 from caais import settings as caais_settings
+from caais.constants import ACCESSION_IDENTIFIER_TYPE
 from caais.db import GroupConcat
 from caais.forms import (
     InlineAppraisalForm,
@@ -70,7 +71,7 @@ class IdentifierInlineAdmin(admin.StackedInline):
     extra = 0
 
     def get_queryset(self, request):
-        ids = super().get_queryset(request).exclude(identifier_type="Accession Identifier")
+        ids = super().get_queryset(request).exclude(identifier_type=ACCESSION_IDENTIFIER_TYPE)
         return ids
 
 
--- a/app/caais/constants.py
+++ b/app/caais/constants.py
@@ -1,0 +1,1 @@
+ACCESSION_IDENTIFIER_TYPE = "Accession Identifier"
--- a/app/caais/constants.py
+++ b/app/caais/constants.py
@@ -1,1 +1,2 @@
+#: Reserved :py:class:`~caais.models.Identifier` type for accession identifiers.
 ACCESSION_IDENTIFIER_TYPE = "Accession Identifier"
--- a/app/caais/forms.py
+++ b/app/caais/forms.py
@@ -5,8 +5,10 @@ from typing import ClassVar
 from django import forms
 from django.core.exceptions import ValidationError
 from django.utils.translation import gettext
+from django.utils.translation import gettext_lazy as _
 from django_countries.fields import CountryField
 
+from caais.constants import ACCESSION_IDENTIFIER_TYPE
 from caais.models import (
     AbstractTerm,
     Appraisal,
@@ -229,7 +231,7 @@ class MetadataForm(CaaisModelForm):
     def save(self, commit: bool = True) -> Metadata:
         """Save the accession identifier input as an Identifier on the metadata object.
 
-        The Identifier is given the reserved name "Accession Identifier".
+        The Identifier is given the reserved type identified by ACCESSION_IDENTIFIER_TYPE.
         """
         metadata: Metadata = super().save(commit)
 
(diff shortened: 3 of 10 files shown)


yenaing-oo (commented):
Interesting bug - it took me a while to figure out how a metadata object is determined to already have an accession identifier assigned or not. It was hard to understand because property naming and reverse names were used. I also got confused between Accession Identifier and Number until I realized they were aliases and only "Accession Identifier" was actually being used.

I see the lookup is happening [here](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/f40028e3f3bfae253406e50fdaad7fb611a7c4a1/app/caais/managers.py#L59-L62). 

Would it be a good idea to use a constant dataclasses such as `ReservedNames.ACCESSION_IDENTIFIER` or something similar? That way, it doesn't have to be hardcoded throughout the app. I see raw strings being used in multiple places.

Also, is it safe to remove "Accession Number" from the lookup? Or is that term considered a valid alternative in archival terminology, hence why we would want to keep it.

danloveg:
> Interesting bug - it took me a while to figure out how a metadata object is determined to already have an accession identifier assigned or not. It was hard to understand because property naming and reverse names were used. I also got confused between Accession Identifier and Number until I realized they were aliases and only "Accession Identifier" was actually being used.
> 
> I see the lookup is happening [here](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/f40028e3f3bfae253406e50fdaad7fb611a7c4a1/app/caais/managers.py#L59-L62).
> 
> Would it be a good idea to use a constant dataclasses such as `ReservedNames.ACCESSION_IDENTIFIER` or something similar? That way, it doesn't have to be hardcoded throughout the app. I see raw strings being used in multiple places.
> 
> Also, is it safe to remove "Accession Number" from the lookup? Or is that term considered a valid alternative in archival terminology, hence why we would want to keep it.

Yeah, I don't love how this is organized in CAAIS. I'm not an archivist, but I would think that the accession identifier, which is supposed to uniquely identify each Metadata object, should be its own thing, but in CAAIS v1.0, it's considered an Identifier.

I would still want to reserve the name "Accession Number" even if that's not what's used to do the lookup. It's too similar to Accession Identifier and is what the accession ID is called in AtoM.

I can definitely move that string into a constant. I also just realized I need to make sure you're not allowed to create an accession identifier in the Metadata admin as well.

Thanks for having a look, made me realize this needs a bit more work.

danloveg:
> > Interesting bug - it took me a while to figure out how a metadata object is determined to already have an accession identifier assigned or not. It was hard to understand because property naming and reverse names were used. I also got confused between Accession Identifier and Number until I realized they were aliases and only "Accession Identifier" was actually being used.
> > I see the lookup is happening [here](https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/f40028e3f3bfae253406e50fdaad7fb611a7c4a1/app/caais/managers.py#L59-L62).
> > Would it be a good idea to use a constant dataclasses such as `ReservedNames.ACCESSION_IDENTIFIER` or something similar? That way, it doesn't have to be hardcoded throughout the app. I see raw strings being used in multiple places.
> > Also, is it safe to remove "Accession Number" from the lookup? Or is that term considered a valid alternative in archival terminology, hence why we would want to keep it.
> 
> Yeah, I don't love how this is organized in CAAIS. I'm not an archivist, but I would think that the accession identifier, which is supposed to uniquely identify each Metadata object, should be its own thing, but in CAAIS v1.0, it's considered an Identifier.
> 
> I would still want to reserve the name "Accession Number" even if that's not what's used to do the lookup. It's too similar to Accession Identifier and is what the accession ID is called in AtoM.
> 
> I can definitely move that string into a constant. I also just realized I need to make sure you're not allowed to create an accession identifier in the Metadata admin as well.
> 
> Thanks for having a look, made me realize this needs a bit more work.

OK, once I dug into the code I agreed that we should use just *one* constant reserved identifier type. So I'm using the type name "Accession Identifier" only now, not "Accession Number" anymore.

I've also verified that you aren't able to add these accession numbers through the Identifier inline.

admin_error

PR #1025, opened 2025-08-26
Refresh submission table when group is deleted

### LINKED ISSUE #1020: Issue: Deleting a group does not clear it from the submission table
**Description of the Bug**

If you delete a group from the submission group table, it appears like it's not actually unlinked from the submission in the submission table. If you try to click the group name that was deleted on the submission, then you get a 404 error.

**Steps to Reproduce**

1. Create a group named GROUP.
2. Assign it to one or more submissions.
3. Delete the group from the submission group table in the profile.
4. Click the Submission tab.
5. Find the submission linked to that group, and click the group link.
6. You will get a 404 error.

**Expected Behaviour**

The group should be cleared from the submission in the table so that it does not appear like it still exists.

**Screenshots**

Deleting the group:

Image

Finding the submission that was linked to that group:

Image

After clicking the group link, you get a 404 error:

Image

**App Configuration**

Both.

**Additional Context**

The message in the delete modal does not have proper pluralization. (1 submission**s**)

danloveg:
Closes #1020

Refreshes the submission table when a group is deleted, in case there were submissions attached to that group.

---

EDIT:

Also, fixed the `lint-js.yml` workflow. Sometimes the workflow will fail like this:

pnpm_not_found

The `cache:` argument in `setup-node` says that the package manager needs to be pre-installed. I take this to mean that the package manager needs to be installed in the worker before the node setup runs.

From [actions/setup-node](https://github.com/actions/setup-node?tab=readme-ov-file#usage) regarding the cache option:

> Supported values: npm, yarn, pnpm.
> Package manager should be pre-installed

And from the [pnpm/action-setup](https://github.com/pnpm/action-setup?tab=readme-ov-file#use-cache-to-reduce-installation-time) documentation, they install pnpm before setting node up.

```yaml
      - uses: pnpm/action-setup@v4
        name: Install pnpm
        with:
          version: 10
          run_install: false

      - name: Install Node.js
        uses: actions/setup-node@v4
        with:
          node-version: 20
          cache: 'pnpm'
```

### DIFF
--- a/.github/workflows/lint-js.yml
+++ b/.github/workflows/lint-js.yml
@@ -24,16 +24,17 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4
+        with:
+          version: latest
+          run_install: false
+
       - name: Setup Node.js
         uses: actions/setup-node@v4
         with:
           node-version: "22"
           cache: "pnpm"
-
-      - name: Install pnpm
-        uses: pnpm/action-setup@v4
-        with:
-          version: latest
 
       - name: Install dependencies
         run: pnpm install --frozen-lockfile


yenaing-oo (approved):
Works perfectly - the change to `lint-js.yml` makes sense as well.

PR #1039, opened 2025-08-27
Fix new submission group button on form does not work

### LINKED ISSUE #1037: Issue: New submission group button on form does not work
**Description of the Bug**

When you click the New Submission Group button on the submission form, instead of opening the modal, it goes to the next step.

**Steps to Reproduce**

1. Log in, and start a submission
2. On step 7 of the form, click **+ New submission group**
3. You will be brought to the next page of the form, instead of being shown the modal

**Expected Behaviour**

The modal should be brought up so you can create a new submission group.

**Screenshots**

Image

**App Configuration**

Tested, happens in both.

**Additional Context**

This bug also seems to affect the browser history. If you try to click **Save** to save the form after this happens, you will be asked to confirm whether you want to leave. You *should* be brought right back to the profile instead of being asked to save.

yenaing-oo:
Closes #1037 

Change notes:
- Also adds an info message to the Group Submission step

### DIFF
--- a/app/recordtransfer/templates/recordtransfer/submission_form_groupsubmission.html
+++ b/app/recordtransfer/templates/recordtransfer/submission_form_groupsubmission.html
@@ -13,6 +13,7 @@
                         hx-target="#modal-content-container"
                         hx-swap="innerHTML"
                         hx-select="#modal-content"
+                        type="button"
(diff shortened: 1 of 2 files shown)


danloveg (approved):
Works great. Nice work on handling the logout button, too.

PR #1041, opened 2025-08-28
Remove upload session endpoint

### LINKED ISSUE #1040: Enhancement: Remove `/upload-session/` endpoint

Once the user has made it to the upload files part of the form, there is nothing to stop them from repeatedly POST-ing to the `/upload-session/` endpoint.

Instead of requiring a POST to make an upload session, we should create an upload session behind the scenes when one is needed, and only create one.

Instead of POST-ing to the `/upload-session/` endpoint, we should create a session for the user when they get to the upload files part of the form. This upload session token will be part of the wizard's session data since it's an input field.

This would remove the need to have that endpoint.

The proper place to set the session token would probably be in `SubmissionFormWizard.get_form_initial`. We can create an UploadSession in that function and assign the hidden `session_token` input at that point. There is no need to keep the UploadSession after generating the token.

If the user tampers with the upload session token input, do nothing. The session will eventually expire.

As part of implementing this change, there should be no need for either of these Javascript functions:

- `fetchNewSessionToken`
- `setSessionToken`

The `getSessionToken` function is still valid, though. 

**Describe alternatives you've considered**

Right now, we don't allow a session to be created until the upload files step is reached, but this is not quite enough protection with regards to creating multiple sessions.

**Additional context**

N/A

danloveg:
Closes #1040 

Removes the upload session creation endpoint, and handles creating the upload session internally.

Uses the `extra_data` field on the session storage to store the session token, and passes that token via the JS context to the template. Instead of the uppy form loading the session token from the input field, it's now loaded via context.

When the submission form is saved, the extra data is also saved. And of course, when the submission form is loaded, this extra data is also loaded.

I've additionally added some more security measures to the UploadFilesForm, namely:

1. The UploadSession objects weren't being filtered by user in the `clean()` function, and now they are.
2. Catches errors where the user tampers with the hidden session_token input.

### DIFF
--- a/app/recordtransfer/forms/submission_forms.py
+++ b/app/recordtransfer/forms/submission_forms.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import logging
 import re
 from datetime import datetime
 from typing import Any, Optional, OrderedDict, TypedDict, Union
@@ -21,6 +22,8 @@ from recordtransfer.constants import (
 from recordtransfer.enums import SubmissionStep
 from recordtransfer.forms.mixins import ContactInfoFormMixin, HiddenCaptchaMixin
 from recordtransfer.models import SubmissionGroup, UploadSession, User
+
+LOGGER = logging.getLogger(__name__)
 
 
 class ReviewFormItem(TypedDict):
@@ -743,6 +746,7 @@ class UploadFilesForm(SubmissionForm):
 
     def __init__(self, *args, **kwargs):
         self.user = kwargs.pop("user", None)
+        self.correct_session_token = kwargs.pop("correct_session_token", None)
         super().__init__(*args, **kwargs)
 
     general_note = forms.CharField(
@@ -777,7 +781,16 @@ class UploadFilesForm(SubmissionForm):
         upload_session = UploadSession.objects.filter(token=token, user=self.user).first()
 
         if not upload_session:
-            self.add_error("session_token", _("Invalid upload. Please try again."))
+            self.add_error("session_token", _("Invalid upload session state. Please try again."))
+            return cleaned_data
+
+        if upload_session.token != self.correct_session_token:
+            LOGGER.warning(
+                "**SUSPICIOUS OPERATION**: hidden session_token input was changed by %(username)s "
+                "in the upload files form",
+                self.user.username,
+            )
+            self.add_error("session_token", _("Invalid upload session state. Please try again."))
             return cleaned_data
 
         if upload_session.file_count == 0:
--- a/app/recordtransfer/forms/submission_forms.py
+++ b/app/recordtransfer/forms/submission_forms.py
@@ -788,7 +788,7 @@ class UploadFilesForm(SubmissionForm):
             LOGGER.warning(
                 "**SUSPICIOUS OPERATION**: hidden session_token input was changed by %(username)s "
                 "in the upload files form",
-                self.user.username,
+                self.user.username if self.user else "unknown user",
             )
             self.add_error("session_token", _("Invalid upload session state. Please try again."))
             return cleaned_data
--- a/app/recordtransfer/forms/submission_forms.py
+++ b/app/recordtransfer/forms/submission_forms.py
@@ -741,6 +741,10 @@ class UploadFilesForm(SubmissionForm):
 
         submission_step = SubmissionStep.UPLOAD_FILES
 
+    def __init__(self, *args, **kwargs):
+        self.user = kwargs.pop("user", None)
+        super().__init__(*args, **kwargs)
+
     general_note = forms.CharField(
         required=False,
         min_length=4,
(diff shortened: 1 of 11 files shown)


app/recordtransfer/forms/submission_forms.py
```diff
@@ -762,26 +770,32 @@ class Meta:
...
 
-        try:
-            upload_session = UploadSession.objects.get(token=session_token)
-        except UploadSession.DoesNotExist:
-            self.add_error("session_token", "Invalid upload. Please try again.")
+        if upload_session.token != self.correct_session_token:
+            LOGGER.warning(
+                "**SUSPICIOUS OPERATION**: hidden session_token input was changed by %(username)s "
+                "in the upload files form",
+                self.user.username if self.user else "unknown user",
```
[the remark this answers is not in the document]
danloveg: Fixed in https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/1041/commits/a36611e6d63e2970a579d2d59a61c8c5d0b7aad0

app/recordtransfer/forms/submission_forms.py
```diff
@@ -762,26 +770,32 @@ class Meta:
...
 
     def clean(self) -> dict:
         """Check that the session token is valid and that at least one file has been uploaded."""
         cleaned_data = super().clean()
-        session_token = cleaned_data.get("session_token")
+        token = cleaned_data.get("session_token")
 
-        if not session_token:
-            self.add_error("session_token", "Invalid upload. Please try again.")
+        upload_session = UploadSession.objects.filter(token=token, user=self.user).first()
```
yenaing-oo, line 781: I can see why this check is necessary with the current setup, since the upload session is being used to set the `upload_session` field for an `InProgressSubmission` here:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/a36611e6d63e2970a579d2d59a61c8c5d0b7aad0/app/recordtransfer/views/pre_submission.py#L304-L317

Would it be possible to skip the need for this check in the form altogether by loading the `upload_session` for an `InProgressSubmission` from `self.storage` like this?

In `SubmissionFormWizard.save_form_data`:

```python
def save_form_data(self, request: HttpRequest) -> None:
       ...
        session = UploadSession.objects.filter(token=self.storage.extra_data.get("session_token"), user=self.request.user).first()
        if session:
            self.in_progress_submission.upload_session = session
        ...
```

That way, we won't even need to include `upload_session` as a field under `UploadFilesForm`.

yenaing-oo: > It's not completely necessary functionality-wise to have the session_token in the form, but, I still want to validate the upload session on the backend before considering the form "clean." The easiest way to do that is just to check it in the clean() method.

Yes, I feel the same way. The `clean()` method in `UploadFilesForm` seems to be the most convenient and fitting place to perform the required check (at least one file is uploaded).

app/recordtransfer/tests/unit/test_forms.py
```diff
@@ -982,7 +982,7 @@ def test_form_no_files_uploaded(self) -> None:
             "session_token": self.upload_session.token,
             "general_note": "Some general note",
         }
-        form = UploadFilesForm(data=form_data)
```
yenaing-oo, line 985: Could you add this test to `UploadFilesFormTest`?:

```python
    def test_form_tampered_session_token(self) -> None:
        """Case where the session token is tampered with. The session token is valid, but belongs
        to a different user.
        """
        other_user = User.objects.create_user(username="otheruser", password="12345")
        other_session = UploadSession.new_session(user=other_user)
        form_data = {
            "session_token": other_session.token,
            "general_note": "Some general note",
        }
        form = UploadFilesForm(data=form_data, correct_session_token=self.upload_session.token, user=self.user)
        self.assertFalse(form.is_valid())
        self.assertIn("session_token", form.errors)
```

Also, I think the other tests are passing even without the kwarg `user` being passed because the lookup for the UploadSession is successful since it's looking for one without a user associated with it (`user=None`). Could you modify the other tests to pass a user that is linked to `self.upload_session`?

yenaing-oo:
The seed data is now invalid since it's missing `session_token` in `self.storage.extra_data`. Would you like me to update the seed data in this branch?

yenaing-oo:
> > The seed data is now invalid since it's missing `session_token` in `self.storage.extra_data`. Resuming either of the in-progress submissions after resetting and seeding would give this: image
> > Would you like me to update the seed data in this branch?
> 
> For sure, go for it.

I've done that here:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/1041/commits/8867537e65c266ef176617df5bd7bf4bfb021460

danloveg:
@yenaing-oo I don't know if you're allowed to approve this since you made the last commit, but let me know if anything needs to be changed. If not, I can approve this

yenaing-oo:
Yes, I think there is one more thing that can be changed:
https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/0bb187d4e567e66822e0d944e7d8e20d059b43d3/app/recordtransfer/views/pre_submission.py#L788-L790

I think we should filter by user as well here (just to be safe), and use the session token from `self.storage.extra_data` instead of from the form.

danloveg:
> Yes, I think there is one more thing that can be changed:
> 
> https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/blob/0bb187d4e567e66822e0d944e7d8e20d059b43d3/app/recordtransfer/views/pre_submission.py#L788-L790
> 
> I think we should filter by user as well here (just to be safe), and use the session token from `self.storage.extra_data` instead of from the form.
> 
> Aside from that, I think this should be ready to go.

That makes sense to me, I made that change in https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/pull/1041/commits/0025061e3df9f78ed5eb7f0b3a0c3da9c6940fe4

yenaing-oo (approved):
Looks good to go!

Issue #1096, opened 2025-12-12
Issue: Login button has no hover styling


danloveg:
### Description of the Bug

When hovering over the Log in button, there are no hover styles to indicate that you are about to click the button or that the button can be interacted with.

### Steps to Reproduce

1. Log out if you're logged in
2. Enter your username and password in the log in form
3. Hover over the Log in button
4. Click Log in

Notice in step 3 that the colours and styling of the button did not change, but you're still able to log in just fine.

### Expected Behaviour

The button's styling should change to indicate that you are about to make an action.

### Suggested Changes

Change the Log in button to have the same hover styling as the Save button on the profile form.

### App Version

v2025.12.4

### App Configuration

- [x] Development (`compose.dev.yml`)
- [x] Production (`compose.prod.yml`)

PR #1103, opened 2025-12-19
Add hover styles for Login buttons similar to Sign Up button

Eashana0104:
Closes https://github.com/NationalCentreTruthReconciliation/Secure-Record-Transfer/issues/1096.

Change notes-
I have replicated the styling of Sign Up button on sign up form on the login button to keep uniformity between similar-looking buttons. In your suggestion I saw you suggested to make it look like outline save button. Let me know what you think. If you think that would look better, I can make the appropriate changes to both login and signup buttons

### DIFF
--- a/app/recordtransfer/static/recordtransfer/css/base/accounts.css
+++ b/app/recordtransfer/static/recordtransfer/css/base/accounts.css
@@ -1,0 +1,28 @@
+/* Error animation */
+@keyframes errorAppear {
+    0% {
+        opacity: 0;
+        transform: translateY(-10px);
+    }
+
+    50% {
+        transform: translateY(0) translateX(3px);
+    }
+
+    65% {
+        transform: translateY(0) translateX(-2px);
+    }
+
+    80% {
+        transform: translateY(0) translateX(1px);
+    }
+
+    100% {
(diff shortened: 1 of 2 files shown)


danloveg (changes requested):
This looks good, but for some reason the error message isn't being animated anymore. Do you remember adding the error message animation? There's no feedback when you enter your password wrong multiple times that tells you your request has been received. It used to be that when you submitted an invalid login attempt, that the error message would animate. Can you take a look at that?

Image

Eashana0104:
> This looks good, but for some reason the error message isn't being animated anymore. Do you remember adding the error message animation? There's no feedback when you enter your password wrong multiple times that tells you your request has been received. It used to be that when you submitted an invalid login attempt, that the error message would animate. Can you take a look at that?
> 
> Image

Yeah I remember adding that, makes me think if this code interfered with the animations, I doubt it just because how little the code change is. I ll look into it!

danloveg:
> Yeah I remember adding that, makes me think if this code interfered with the animations, I doubt it just because how little the code change is. I ll look into it!

Yeah my guess is that the animation got accidentally removed at some point.

PR #1105, opened 2025-12-19
Migrate build and runtime to bun

### LINKED ISSUE #1093: Migrate to bun
I'm considering [bun](https://bun.com/docs) as an all-in-one package manager and static asset bundler. I would like to speed up the static asset build step and moving away from Webpack could help that happen.

Marking this as an opinion for now.

danloveg:
Closes #1093

Swaps NodeJS and NPM for Bun.

- Use Bun instead of Webpack for building JS
- Use Tailwind CLI directly instead of Webpack for building CSS
  - There *is* a plugin for Bun + Tailwind but it's buggy and wasn't working properly
- Move buildable assets to `app/frontend/`
  - Assets not in this directory can be loaded as regular Django assets can be.

Moving to Bun simplifies the Javascript toolchain by combining the runtime, package installer, and asset bundler into one tool. Whereas before we had Node, NPM, and Webpack. Moving to Bun has massively improved the build time. The development and production build each take about 1s with Bun. The same builds take about 5s and 20s for the development and production builds with Webpack, respectively. So a 5x improvement for development, and a 20x improvement for production.


danloveg:
Closing - this change is too large and the benefits have ended up being minimal for the amount of new complexity this change adds. There are two main issues with this approach:

- Bun build is not as full-featured as Webpack. It is fast, but it doesn't work with Tailwind well.
- Tailwind requires NodeJS, and installing the standalone CLI executable is a risky and janky solution.

